PR #5, opened 2024-12-25
refactor: no DPDK includes in header files

3Hren:
This commit makes our codebase's API (which are header files) isolated from DPDK by removing DPDK includes in header files. Ideally we also must hide rte_* symbols in translation units. This potentially gives us independence from DPDK version, but this change is not going to be implemented right now.

Segregating header files from DPDK includes is also a prerequesite for DPDK-less build for lightweighted checks in github.

### DIFF
--- a/dataplane/packet/packet.c
+++ b/dataplane/packet/packet.c
@@ -3,6 +3,7 @@
 #include <stdint.h>
 
 #include <rte_ether.h>
+#include <rte_ip.h>
 
 /*
  * TODO: analyze if the valid packet parsing may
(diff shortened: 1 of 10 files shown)


dataplane/packet/packet.c
```diff
@@ -217,3 +218,8 @@ parse_packet(struct packet *packet) {
 
 	return 0;
 }
+
+struct packet *
```
3Hren, line 222: This change may have a significant impact on performance, but I hope the LTO must be the thing that makes this change irrelevant.

GeorgyKirichenko: I think the performance degradation will not be significant, so as we do not have any measurements now let merge it. In any case we could manage this in the future.

GeorgyKirichenko (approved):

PR #48, opened 2025-03-06
feat: implement MapTrie for RIB

sakateka:


controlplane/modules/route/internal/rib/map_trie.go
```diff
@@ -0,0 +1,70 @@
...
+
+type MapTrie [129]map[MapTrieKey]*RoutesList
+
+func NewMapTrie(preAllocSize int) MapTrie {
+	mt := MapTrie{}
+	for idx := range mt {
+		mt[idx] = make(map[MapTrieKey]*RoutesList, preAllocSize)
+	}
+	return mt
+}
```
3Hren, line 19: Suggest better naming - `capacity`.

```suggestion
func NewMapTrie(cap int) MapTrie {
	mt := MapTrie{}
	for idx := range mt {
		mt[idx] = make(map[MapTrieKey]*RoutesList, cap)
	}
	return mt
}
```

controlplane/modules/route/internal/rib/map_trie.go
```diff
@@ -0,0 +1,70 @@
...
+func (m *MapTrie) Lookup(addr netip.Addr) (*RoutesList, bool) {
+	maxBits := 129
+	base := 128
+	if addr.Is4() {
+		base = 32
+		maxBits = 33
+	}
+	for n := range maxBits {
+		bits := base - n
+		p, err := addr.Prefix(bits)
```
3Hren, line 30: If it is impossible, maybe just ignore error here?

controlplane/modules/route/internal/rib/map_trie.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		mtk := MapTrieKey{Prefix: p}
+		if _, ok := m[bits][mtk]; ok {
+			return nil, false
+		}
+	}
+	return nil, false
+}
+
+func (m *MapTrie) Insert(key MapTrieKey, route Route) {
```
3Hren, line 42: Suggest instead of just `Insert` function like `InsertOrUpdate(key, onVacant func() Route, onOccupied func(*Route))`

3Hren: Or [entry-like](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html) API

3Hren (changes requested):

3Hren (approved):

PR #61, opened 2025-03-11
feat: HTTP <-> gRPC transparent proxy

3Hren:
The idea is to terminate HTTP requests on Gateway side, then just pass them directly to gRPC streams without modification.

This approach allows to forward requests/responses without knowing its protobuf definition, and that's the reason why we don't use go-grpc-gateway to solve the issue.

### DIFF
--- a/controlplane/internal/gateway/gateway.go
+++ b/controlplane/internal/gateway/gateway.go
@@ -166,9 +166,7 @@ func (m *Gateway) Run(ctx context.Context) error {
 		})
 	}
 
-	// Run built-in modules.
 	for _, builtInModule := range m.builtInModules {
-		builtInModule := builtInModule // capture for closure.
 		wg.Go(func() error {
 			m.log.Infow("starting built-in module", zap.String("module", fmt.Sprintf("%T", builtInModule)))
 			return builtInModule.Run(ctx)
--- a/controlplane/internal/gateway/gateway.go
+++ b/controlplane/internal/gateway/gateway.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/http"
+	"time"
 
 	"github.com/siderolabs/grpc-proxy/proxy"
 	"go.uber.org/zap"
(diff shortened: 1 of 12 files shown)


controlplane/internal/gateway/gateway.go
```diff
@@ -173,3 +187,31 @@ func (m *Gateway) Run(ctx context.Context) error {
...
+
+	// Set up graceful shutdown.
+	go func() {
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		m.log.Infow("shutting down HTTP server", zap.String("addr", m.cfg.Server.HTTPEndpoint))
+		if err := server.Shutdown(shutdownCtx); err != nil {
+			m.log.Warnw("failed to  shut down HTTP server", zap.Error(err))
```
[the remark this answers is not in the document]
3Hren: Fixed.

controlplane/internal/gateway/proxy.go
```diff
@@ -0,0 +1,202 @@
...
+		return
+	}
+
+	// Read request body (binary protobuf payload).
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to read request body: %v", err), http.StatusBadRequest)
+		return
+	}
+	defer r.Body.Close()
```
sakateka, line 76: Is it necessary to defer the `Close` method?

controlplane/internal/gateway/proxy.go
```diff
@@ -0,0 +1,202 @@
...
+			zap.String("service", service),
+			zap.String("method", method),
+			zap.Error(err),
+		)
+		http.Error(w, fmt.Sprintf("Failed to connect to backend: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	// Create byte buffer for response.
+	var respData []byte
```
sakateka, line 101: Why was the declaration placed so far from its actual usage?

controlplane/internal/gateway/proxy.go
3Hren: It appears isn't necessary to `Close` at all. I removed this line.

controlplane/internal/gateway/proxy.go
3Hren: Fixed.

sakateka (approved):

PR #82, opened 2025-03-18
Agent introspection

GeorgyKirichenko:
The commit adds routines to fetch and decode information about agents connected to `YANET` shared memory. At the point one may list all connected agents with history and fetch information about `pid` and memory consumption of each connection instance in form like
```
Agent bootstrap
  Pid 12579 Limit 4194304 Allocated 336 Freed 0
  Pid 12572 Limit 4194304 Allocated 336 Freed 0
  Pid 12565 Limit 4194304 Allocated 336 Freed 0
  Pid 12559 Limit 4194304 Allocated 336 Freed 0
  Pid 5479 Limit 4194304 Allocated 336 Freed 0
```

### DIFF
--- a/lib/controlplane/agent/agent.c
+++ b/lib/controlplane/agent/agent.c
@@ -7,7 +7,10 @@
 #include <fcntl.h>
 #include <unistd.h>
 
+#include <errno.h>
+
 #include "common/memory.h"
+#include "common/strutils.h"
 
 #include "dataplane/config/zone.h"
 
@@ -90,7 +93,8 @@ agent_attach(
 	struct agent *new_agent = (struct agent *)memory_balloc(
 		&cp_config->memory_context, sizeof(struct agent)
 	);
-	strncpy(new_agent->name, agent_name, 80);
+	strtcpy(new_agent->name, agent_name, 80);
+	new_agent->memory_limit = memory_limit;
 	block_allocator_init(&new_agent->block_allocator);
 	memory_context_init(
 		&new_agent->memory_context,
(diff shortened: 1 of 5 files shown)


lib/controlplane/agent/agent.c
```diff
@@ -426,22 +430,99 @@ yanet_get_cp_pipeline_module_info(
...
+		struct cp_agent_info *agent_info = (struct cp_agent_info *)
+			malloc(sizeof(struct cp_agent_info) +
+			       sizeof(struct cp_agent_instance_info) *
+				       instance_count);
+		if (agent_info == NULL) {
+			cp_agent_list_info_free(agent_list_info);
+			agent_list_info = NULL;
+		}
+
+		strncpy(agent_info->name, agent->name, 80);
```
3Hren, line 509: Potential arbitrary memory read in case of `len(agent->name) == 80`.

3Hren (changes requested):

lib/controlplane/agent/agent.c
```diff
@@ -426,22 +430,99 @@ yanet_get_cp_pipeline_module_info(
...
 }
 
-struct cp_agent_list_info {};
+int
+yanet_get_cp_agent_instance_info(
+	struct cp_agent_info *agent_info,
+	uint64_t index,
+	struct cp_agent_instance_info **instance_info
+) {
+	if (index >= agent_info->instance_count) {
```
3Hren, line 439:
```suggestion
	if (index >= agent_info->instance_count) {
	    errno = ERANGE;
```

lib/controlplane/agent/agent.c
GeorgyKirichenko: There are new function `strtcpy` implemented as `man string_copying` suggested.

3Hren (approved):

PR #98, opened 2025-03-27
Code move and cleanup

GeorgyKirichenko:
Split `lib/dataplane/config/zone.h` into partitions


controlplane/internal/ffi/agent.go
```diff
@@ -2,6 +2,8 @@ package ffi
 
 //#cgo CFLAGS: -I../../../ -I../../../lib
 //#cgo LDFLAGS: -L../../../build/lib/controlplane/agent -lagent
+//#cgo LDFLAGS: -L../../build/lib/controlplane/config -lconfig_cp
```
sakateka, line 5: You need to go up one level.

controlplane/internal/ffi/shm.go
```diff
@@ -2,6 +2,8 @@ package ffi
 
 //#cgo CFLAGS: -I../../../
 //#cgo LDFLAGS: -L../../../build/lib/controlplane/agent -lagent
+//#cgo LDFLAGS: -L../../build/lib/controlplane/config -lconfig_cp
+//#cgo LDFLAGS: -L../../build/lib/dataplane/config -lconfig_dp
```
sakateka, line 6: You need to go up one level.

Issue #193, opened 2025-06-04
Get RID of Numa Node abstraction from dataplane


GeorgyKirichenko:
**Motivation**
`YANET` may have as many configuration (`dp`+`cp`) as `NUMA` nodes exist what looks as an artificial limitation. Also `NUMA` abstraction also leaks to `controlplane` configuration.

**Possible solution**
Change `dataplane` configuration to a list of instances (`cp`+`dp`) with  `NUMA` placement options assigned to. `Controlplane` will use instance identifier instead of `NUMA` one. Also we may have more than one configuration instances on a single `NUMA` node.

egnees:
Draft of the possible solution is implemented in [PR#198](https://github.com/yanet-platform/yanet2/pull/198).

PR #198, opened 2025-06-06
Get rid of NUMA node abstraction from dataplane

egnees:
Refers to [ISSUE#193](https://github.com/yanet-platform/yanet2/issues/193).

YANET2 shared memory is divided into regions which we call dataplane instances. Until now, one dataplane instance corresponded to the single NUMA node, which is an artificial limitation. 

This PR allows several instances to be attached to the same NUMA node.

For this, dataplane config is changed.

```yaml
dataplane:
  storage: /dev/hugepages/yanet
  dpdk_memory: 1024
  instances:
    - dp_memory: 1073741824
      cp_memory: 1073741824
      numa_id: 0
  devices:
    - port_name: 07:00.0
      mtu: 7000
      max_lro_packet_size: 7200
      rss_hash: 0
      workers:
        - core_id: 4
          instance_id: 0
          rx_queue_len: 1024
          tx_queue_len: 1024
    - port_name: virtio_user_kni0
      mac_addr: 52:54:00:6b:ff:a5
      mtu: 7000
      max_lro_packet_size: 7200
      rss_hash: 0
      workers:
        - core_id: 5
          instance_id: 0
          rx_queue_len: 1024
          tx_queue_len: 1024
  connections:
    - src_device_id: 0
      dst_device_id: 1
    - src_device_id: 1
      dst_device_id: 0
```

The`instances` list appeared, fields `numa_count`, `dp_memory` and `cp_memory` were removed. 

Each instance is configured with the `dp_memory`, `cp_memory` and index of the numa node it is attached to.

Workers must be attached to instances. For this, the `instance_id` field is specified for each worker.

## Changes in the implementation

The NUMA map appeared to be useless and was removed from the project.

The NUMA index is still stored in the `dp_config` and can be used to attach workers to the local NUMA node.

### Agent API

Agent must be connected to the dataplane instance. For this, agent must specify instance id (and not NUMA node id as in the past). The corresponding changes were introduced in the [agent API](https://github.com/yanet-platform/yanet2/blob/get-rid-of-numa/api/agent.h). 

### Controlplane

For now, controlplane inspects dataplane per dataplane instance and not per NUMA node as in the past. The corresponding changes were introduced in the [inspect protofile](https://github.com/yanet-platform/yanet2/blob/get-rid-of-numa/controlplane/ynpb/inspect.proto) and in the [inspect service](https://github.com/yanet-platform/yanet2/blob/get-rid-of-numa/controlplane/internal/gateway/inspect_service.go).

Also, agent API changes lead to [ffi](https://github.com/yanet-platform/yanet2/tree/get-rid-of-numa/controlplane/ffi) refactoring. 

### Modules

For now, each GO module connects to a dataplane instance.

## Testing

For now, only one [unit test](https://github.com/yanet-platform/yanet2/tree/get-rid-of-numa/dataplane/unittest/config) was implemented. This test checks new config can be parsed successfully. Probably, more tests are required.


sakateka:
Try this command for clang-format from the root of the project
```sh
fd -e h -e c -E build -E subprojects -x clang-format-19 -i
```
Or try this if `fd` is not available:
```
find . -path './build' -prune -o -path './subprojects' -prune -o \( -name "*.h" -o -name "*.c" \) -print -exec clang-format-19 -i {} +
```

PR #219, opened 2025-06-20
forward: added functionality for delete instance config

egnees:
Added function for delete config of forward module instance.

The config can be deleted only if it is not referenced by any pipeline.

The delete command is parameterised with dataplane instance of forward agent, for which delete is requested, and with the config name. The delete command returns bitmap of instances for which operation was successful. If delete is not possible for some instance, the error wont be thrown.

### DIFF
--- a/lib/controlplane/agent/agent.c
+++ b/lib/controlplane/agent/agent.c
@@ -994,6 +994,10 @@ agent_free_unused_modules(struct agent *agent) {
 
 void
 agent_free_unused_agents(struct agent *agent) {
+	if (agent == NULL) {
+		return;
+	}
+	agent = ADDR_OF(&agent->prev);
 	while (agent != NULL) {
 		struct agent *prev_agent = ADDR_OF(&agent->prev);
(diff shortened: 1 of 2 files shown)


lib/controlplane/agent/agent.c
```diff
@@ -966,3 +973,27 @@ void
...
+	while (ADDR_OF(&agent->prev) != NULL) {
+		struct agent *prev_agent = ADDR_OF(&agent->prev);
+
+		if (prev_agent->loaded_module_count == 0) {
+			SET_OFFSET_OF(&agent->prev, ADDR_OF(&prev_agent->prev));
+			agent_cleanup(prev_agent);
+			continue;
+		}
+
+		agent = ADDR_OF(&agent->prev);
```
sakateka, line 997: It can lead to a null pointer dereference during the next check of the while condition. The loop conditions need to be reworked.

sakateka (changes requested):

sakateka (approved):

PR #226, opened 2025-07-17
Generic packet filter

egnees:
This PR brings generic packet filter. Refer to [this](https://github.com/yanet-platform/yanet2/blob/classification/filter/README.md) tutorial for more details.

Filter is composed of classifiers for every interesting packet attribute (for example, IPv4 source address and source port).

Every classifier can be implemented by user (or default classifiers can be used). For this, user specifies functions for classifier initialization, classifier lookup and free classifier storage. Then filter is composed from combination of classifiers.

## Macros

To inline classifier lookups, `FILTER_QUERY` macros can be used. Macros allows the performance of generic packet filter be comparable to a hardcoded one. To use it, `FILTER_DECLARE` macros must be called before, which declares list of filter attributes, or filter signature. As this signature is constantly defined in compile time, compiler can inline queries to filter, which significantly increases its performance.

## Current macros limitations

-  ~~Two filters can not be created in the same scope~~ Fixed
- ~~One can query, init and free filter only from the same scope where it was declared (or from the child scope)~~ Fixed

## TODO
- [x] Support irregular network masks (192.160.1.0/255.240.255.0) [can be done later independently]
- [x] Add default classifier for IPv6 [can be done later independently]
- [x] Process with current macros limitations
- [x] Readme
- [x] Test non-terminative actions [can be done later independently]
- [x] Perf


filter/filter.c
```diff
@@ -0,0 +1,194 @@
...
+		struct filter_vertex *v = &filter->v[vertex];
+
+		// store calculated classifier in the parent vertex
+		filter->v[vertex / 2].slots[vertex & 1] =
+			value_table_get(&v->table, v->slots[0], v->slots[1]);
+	}
+
+	// get result from root
+
+	// root is 1 when n>1 and 0 else.
```
sakateka, line 157: These comments should be included in the struct itself as well.

filter/tests/basic_proto.c
```diff
@@ -0,0 +1,93 @@
...
+	query_udp_packet(&filter, 2);
+
+	filter_free(&filter);
+}
+
+int
+main() {
+	void *memory = malloc(1 << 24);
+
+	test_vlan_1(memory);
```
sakateka, line 86: Is there a copy-pasted name?

filter/util.h
```diff
@@ -0,0 +1,104 @@
...
+	struct filter *filter, struct packet *packet, uint32_t expected_action
+);
+
+void
+query_filter_and_expect_no_actions(
+	struct filter *filter, struct packet *packet
+);
+
+struct filter_rule_builder {
+	struct net6 net6_dst[10];
```
sakateka, line 33: magic constant 10

filter/util.c
```diff
@@ -0,0 +1,227 @@
...
+#include <assert.h>
+
+#include <rte_ether.h>
+#include <rte_ip.h>
+#include <rte_mbuf.h>
+#include <rte_tcp.h>
+#include <rte_udp.h>
+
+static struct rte_mbuf *
+make_mbuf(
```
sakateka, line 14: Can we merge this with https://github.com/yanet-platform/yanet2/blob/main/lib/dataplane/module/testing.c#L6?

And these helpers should be included only in the testing build target.

sakateka (changes requested):

common/remap.h
```diff
@@ -179,8 +177,22 @@ remap_table_new_key(struct remap_table *table, uint32_t *key) {
 			return -1;
 		}
 
+		for (uint64_t chunk_idx = 0; chunk_idx < old_chunk_count;
```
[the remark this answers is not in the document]
egnees: Added `EQUATE_OFFSET(PTR1, PTR2)` macros which makes `PTR1 = PTR2` in terms of relative pointers. After that,  `ADDR_OF(PTR1) == ADDR_OF(PTR2)`. Also added docs for macros `EQUATE_OFFSET`, `SET_OFFSET` and  `ADDR_OF`.

filter/attribute/proto.h
```diff
@@ -0,0 +1,140 @@
...
+		const struct filter_proto *proto = &r->transport.proto;
+		if (proto->proto != IPPROTO_TCP) { // not TCP
+			continue;
+		}
+		if (proto->disable_bits & proto->enable_bits) {
+			// impossible
+			continue;
+		}
+		value_table_new_gen(&c->tcp_flags);
+		int16_t mask = proto->disable_bits ^ ((1 << 9) - 1) ^
```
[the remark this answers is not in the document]
egnees: Fixed

filter/attribute/proto.h
```diff
@@ -0,0 +1,140 @@
...
+			value_table_touch(
+				&c->tcp_flags, 0, m | proto->enable_bits
+			);
+		}
+		value_table_touch(&c->tcp_flags, 0, proto->enable_bits);
+	}
+
+	value_table_compact(&c->tcp_flags);
+	c->max_tcp_class = 0;
+	for (uint16_t i = 0; i < (1 << 9); ++i) {
```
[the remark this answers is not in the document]
egnees: Fixed

filter/tests/basic_proto.c
egnees: Fixed

filter/bench/func.c
```diff
@@ -0,0 +1,262 @@
...
+#define PACKETS 1000000
+
+////////////////////////////////////////////////////////////////////////////////
+
+struct packet *
+gen_packets(size_t count) {
+	struct packet *packets = malloc(sizeof(struct packet) * count);
+	int g = 3;
+	for (size_t i = 0; i < count; ++i) {
+		g += 13 * 17;
```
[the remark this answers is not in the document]
egnees: What do you mean here? Add separate public function for generate packets in test utils?

filter/filter.c
egnees: Added more docs in filter.h

filter/util.c
egnees: Wrapped helpers in the static library and linked with tests and benchmarks only. 

These two functions are different and with different signature. I think I will be good to make separated builder for the `rte_mbuf`. I hope to think about it and make it in the separated PR.

PR #240, opened 2025-08-27
Benchmark filter versus DPDK ACL

egnees:
This PR fixes minor bugs in filter and brings benchmark versus DPDK ACL module.

Also, as experiment the trie data structure was implemented, which allows to make packet classification on IPv4 address. Trie appears to be less effective on the lookup query, than LPM. So, the LPM is still used and trie is being study for now.


sakateka (commented):
It appears the trie cannot be used across processes via shared memory, although this might be acceptable. Additionally, `collect_rule_classifiers_rec` uses a stack without bounds checking, which is potentially problematic.

PR #276, opened 2025-11-27
Yanet Mock + Allocator ASan

egnees:
This PR brings Yanet Mock, which allows to run controlplane together with the mocked dataplane in the single thread, without DPDK. The usage example can be found in `mock/tests` and `mock/go`.

Also, the minimal viable ASan annotations added in the block allocator. In future, they can be extended with the [red zones](https://docs.oracle.com/javame/config/cdc/cdc-opt-impl/ojmeec/1.0/reference/html/z4000c881287235.html) to handle more memory issues.

### DIFF
--- a/common/asan.h
+++ b/common/asan.h
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <stddef.h>
+
+#if defined(__has_feature)
+#if __has_feature(address_sanitizer)
+#define HAVE_ASAN 1
+#endif
+#endif
+
+#if !defined(HAVE_ASAN)
+#if defined(__SANITIZE_ADDRESS__)
+#define HAVE_ASAN 1
+#endif
+#endif
+
+#ifdef HAVE_ASAN
+
+#include <sanitizer/asan_interface.h>
+
+static inline void
+asan_poison_memory_region(void *addr, size_t size) {
+	__asan_poison_memory_region(addr, size);
+}
+static inline void
+asan_unpoison_memory_region(void *addr, size_t size) {
+	__asan_unpoison_memory_region(addr, size);
+}
+
+#else
+
+static inline void
+asan_poison_memory_region(void *addr, size_t size) {
+	(void)addr;
+	(void)size;
+}
+static inline void
+asan_unpoison_memory_region(void *addr, size_t size) {
+	(void)addr;
+	(void)size;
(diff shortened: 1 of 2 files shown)


lib/utils/packet.c
```diff
@@ -0,0 +1,398 @@
...
+
+#include <rte_build_config.h>
+#include <string.h>
+
+#include "lib/dataplane/packet/packet.h"
+
+////////////////////////////////////////////////////////////////////////////////
+
+static struct rte_mbuf *
+make_mbuf4(
```
sakateka: Can we merge these functions to avoid seeding havoc?
https://github.com/yanet-platform/yanet2/blob/482815b9d543b3ea8c6b8e092fafdccbffec88b2/lib/dataplane/module/testing.c#L6

lib/utils/packet.c
```diff
@@ -0,0 +1,398 @@
...
+static void
+init_mbuf(struct rte_mbuf *m, struct packet_data *data, uint16_t buf_len) {
+	m->priv_size = 0;
+	m->buf_len = buf_len;
+	uint32_t mbuf_size = sizeof(struct rte_mbuf) + m->priv_size;
+
+	/* start of buffer is after mbuf structure and priv data */
+	m->buf_addr = (char *)m + mbuf_size;
+
+	// what for?
```
sakateka: I think it was just a precaution, but from the current point of view, it's not necessary

lib/utils/go/packet.go
```diff
@@ -0,0 +1,148 @@
...
+	return (*PacketList)(&(*C.struct_packet_front)(pf).drop)
+}
+
+func Free(pf *PacketFront) {
+	pf.OutputList().Free()
+	pf.InputList().Free()
+	pf.DropList().Free()
+}
+
+func main() {
```
sakateka: 🤔

common/go/xpacket/packet.go
```diff
@@ -0,0 +1,52 @@
+package xpacket
```
sakateka: Why not just move this file from tests/go/common/helpers.go here and update the references in other places? It seems like the right time to do this rather than duplicating it.

sakateka (changes requested):

PR #516, opened 2026-03-09
feat: metrics + usage in balancer

egnees:


common/commonpb/metrics_test.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func TestMetricValueToProto_Histogram(t *testing.T) {
+	histogram := metrics.NewHistogram([]float64{10, 50, 100})
+	histogram.Observe(5)   // bucket 0 (<=10)
+	histogram.Observe(25)  // bucket 1 (<=50)
+	histogram.Observe(75)  // bucket 2 (<=100)
+	histogram.Observe(200) // bucket 3 (+Inf)
+
+	result := MetricValueToProto(histogram)
```
3Hren: A conversion test should verify the mapping between Go types and proto fields, not re-test the metric logic itself. Here the test calls `histogram.Observe(...)` and then checks that values ended up in the right buckets — but that is the responsibility of the `histogram` package tests. The conversion test should start from a metric already in a known state (e.g. buckets pre-populated) and assert only that `MetricValueToProto` maps each field to the correct proto field.

common/go/metrics/histogram_test.go
```diff
@@ -0,0 +1,142 @@
...
+		{-7, 1},
+		{-5, 1},
+		{-1, 2},
+		{0, 2},
+		{3, 3},
+		{5, 3},
+		{100, 4},
+	}
+
+	for _, tc := range tests {
```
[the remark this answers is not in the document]
egnees: Fixed.

common/go/metrics/map.go
```diff
@@ -0,0 +1,92 @@
...
+// GetOrCreate returns the metric for the given label set, creating it via
+// create if it does not yet exist. Order of labels is important
+func (m *MetricMap[T]) GetOrCreate(id MetricID, create func() T) T {
+	h := hashID(id)
+
+	m.mu.RLock()
+	if bucket, ok := m.entries[h]; ok {
+		for idx := range bucket {
+			if bucket[idx].id.EqualOrdered(id) {
+				m.mu.RUnlock()
```
saushew: Removing the read lock before returning may cause panic on the return execution.

`defer` will solve this problem, but I'd prefer to decompose this function to private `get(...)` and `create(...)`, for more straight forward mutex handling

PR #616, opened 2026-04-01
Replace FILTER_INIT and FILTER_FREE macro using inline functions

GeorgyKirichenko:
Use inline function instead of macro with increases code readability and improves debugging. The commit is a part of filter library refactoring.


3Hren (changes requested):
No DPDK updates, sir

3Hren (approved):

3Hren (approved):

PR #643, opened 2026-04-14
refactor(balancer): dataplane/controlplane/cli

egnees:


modules/balancer/dataplane/dataplane.c
```diff
@@ -1,114 +1,148 @@
...
-			packets_count = 0;
-		}
+		context.common_stats->incoming_bytes += packet->mbuf->pkt_len;
+
+		int is_ipv6 = packet->network_header.type ==
+			      rte_cpu_to_be_16(RTE_ETHER_TYPE_IPV6);
+		int is_icmp = packet->transport_header.type == IPPROTO_ICMP ||
+			      packet->transport_header.type == IPPROTO_ICMPV6;
+		int idx = is_icmp * 2 + is_ipv6;
+		batcher_add(&batchers[idx], &context, packet);
```
GeorgyKirichenko, line 134: What about non-IP packets?

modules/balancer/dataplane/dataplane.c
```diff
@@ -1,114 +1,148 @@
...
+	 *
+	 * Batched processing allows the filter engine to evaluate multiple
+	 * packets at once, which is significantly faster than one-at-a-time
+	 * lookups due to memory prefetching and reduced per-packet overhead.
+	 *
+	 * Batches are flushed when they reach MAX_BATCH_SIZE or when all
+	 * input packets have been classified.
+	 */
+	enum { l4_ipv4, l4_ipv6, icmp_ipv4, icmp_ipv6, batcher_count };
+	struct packet_batcher batchers[batcher_count] = {
```
GeorgyKirichenko, line 116: Could we allocate entries on stack like `struct packet *ip4[packet_count(&packet_front->input)]` instead of fixed-size arrays?

modules/balancer/dataplane/dataplane.c
egnees: Should we drop them or pass further?

modules/balancer/dataplane/dataplane.c
egnees: Dataplane counters shows that under high-load batch size can extend 32 packets. Also, in theory module before balancer can produce any count of packets which can cause stack overflow. I think manual batching is more robust in that case. 

Also, small packet batches allows to use specialized algorithms further, which would be inefficient on huge batches.

modules/balancer/dataplane/dataplane.h
```diff
@@ -1,6 +1,70 @@
...
\ No newline at end of file
+new_module_balancer();
+
+struct balancer_packet_handler {
+	struct cp_module cp_module;
+
+	uint64_t common_counter_id;
+	uint64_t icmp_v4_counter_id;
+	uint64_t icmp_v6_counter_id;
+	uint64_t l4_counter_id;
```
GeorgyKirichenko, line 26: icmp v4 and v6 are l4 level protocol too so it is not clean what the naming means here

modules/balancer/dataplane/dataplane.h
```diff
@@ -1,6 +1,70 @@
...
+	struct net4_addr source_v4;
+	struct net6_addr source_v6;
+
+	/*
+	 * RCU guard for the inner atomic changes on the packet handler.
+	 * It includes changes on reals ring of virtual services.
+	 */
+	rcu_t rcu;
+
+	/* ---Controlplane data --- */
```
GeorgyKirichenko, line 50: There SHOULD NOT be any reference to controlplane!

modules/balancer/dataplane/types/session.h
```diff
@@ -0,0 +1,83 @@
...
+	 * stable_idx of the assigned real; see struct balancer_vs.
+	 */
+	uint64_t real_stable_idx;
+	uint32_t last_packet_timestamp;
+	uint32_t create_timestamp;
+	uint8_t timeout;
+};
+
+struct balancer_session_timeouts {
+	uint8_t tcp_syn_ack;
```
GeorgyKirichenko, line 28: I am afraid the type value range is not wide enough

modules/balancer/dataplane/dataplane.h
egnees: I wanted to separate ICMP and TCP/UDP processing. Also, i wanted to separated their counters. I had no idea how to unite TCP and UDP under the single name which clearly separate them from ICMP.

modules/balancer/controlplane/helpers/sessions.c
```diff
@@ -0,0 +1,154 @@
+#include <assert.h>
```
GeorgyKirichenko, line 1: Why the file is living here ant not in `api/`

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+	uint8_t client_len,
+	const struct net6 *src_net
+) {
+	struct rte_ipv6_hdr *outer = rte_pktmbuf_mtod_offset(
+		mbuf, struct rte_ipv6_hdr *, network_offset
+	);
+	uint8_t *dst = outer->src_addr;
+
+	if (client_len == NET6_LEN) {
+		__m128i v_dst = _mm_loadu_si128((__m128i *)dst);
```
GeorgyKirichenko, line 46: We should not rely on CPU version or even architecture as looking forward to support not only x86_64 only.
In cases when you are pretty sure SSE is strongly required please write a default implementation here and then use `ifunc` attribute to load any specific optimized alternative instead of a default one.

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+	uint16_t network_offset,
+	const uint8_t *client_src,
+	const struct net4 *src_net
+) {
+	struct rte_ipv4_hdr *outer = rte_pktmbuf_mtod_offset(
+		mbuf, struct rte_ipv4_hdr *, network_offset
+	);
+
+	uint32_t client, mask, addr;
+	__builtin_memcpy(&client, client_src, NET4_LEN);
```
GeorgyKirichenko, line 80: I do not like builtins here, if you really want them - please define an inline function inside common library and use it.

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+ * The operation per byte is:
+ *   outer_src[i] |= client_src[i] & ~mask[i]
+ *
+ * For 16-byte (IPv6) clients this maps to three 128-bit instructions:
+ *   andnot, or, store.
+ *
+ * For 4-byte (IPv4) clients a scalar fallback is used.
+ */
+static inline void
+embed_client_ip6(
```
GeorgyKirichenko, line 33: Both `embed` functions are excessive - why not evaluate randomized tunnel source address and pass it to the `encap_` functions&

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/*
+ * Update per-VS and per-real forwarding counters.
+ */
+static inline void
+update_tunnel_stats(struct l4_packet_context *pkt_ctx) {
+	struct balancer_vs_stats *vs_stats = pkt_ctx->matched_vs_stats;
+	struct balancer_real_stats *real_stats = pkt_ctx->resolved_real_stats;
+	uint64_t pkt_bytes = pkt_ctx->packet->mbuf->pkt_len;
```
GeorgyKirichenko, line 200: Please use `packet_size` routine

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+ */
+static void
+encapsulate_ipv6(struct packet *packet, struct balancer_real *real) {
+	struct rte_mbuf *mbuf = packet_to_mbuf(packet);
+	bool is_outer_ipv6 = real->flags & balancer_real_ipv6;
+
+	uint16_t inner_offset = packet->network_header.offset;
+
+	if (is_outer_ipv6) {
+		packet_ip6_encap(
```
GeorgyKirichenko, line 156: The function may fail, fails should be accounted and handled

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+
+		embed_client_ip6(
+			mbuf,
+			packet->network_header.offset,
+			client_src,
+			NET6_LEN,
+			&real->src.v6
+		);
+	} else {
+		packet_ip4_encap(
```
GeorgyKirichenko, line 174: The function may fail, fails should be accounted and handled

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+ */
+static void
+encapsulate_ipv4(struct packet *packet, struct balancer_real *real) {
+	struct rte_mbuf *mbuf = packet_to_mbuf(packet);
+	bool is_outer_ipv6 = real->flags & balancer_real_ipv6;
+
+	uint16_t inner_offset = packet->network_header.offset;
+
+	if (is_outer_ipv6) {
+		packet_ip6_encap(
```
GeorgyKirichenko, line 107: The function may fail, fails should be accounted and handled

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+
+		embed_client_ip6(
+			mbuf,
+			packet->network_header.offset,
+			client_src,
+			NET4_LEN,
+			&real->src.v6
+		);
+	} else {
+		packet_ip4_encap(
```
GeorgyKirichenko, line 125: The function may fail, fails should be accounted and handled

modules/balancer/dataplane/l4/gre.c
```diff
@@ -0,0 +1,88 @@
...
+
+void
+insert_gre_header(
+	struct packet *packet, bool is_outer_ipv6, bool is_inner_ipv4
+) {
+	struct rte_mbuf *mbuf = packet_to_mbuf(packet);
+	const uint16_t gre_size = sizeof(struct rte_gre_hdr);
+
+	if (unlikely(rte_pktmbuf_prepend(mbuf, gre_size) == NULL)) {
+		return;
```
GeorgyKirichenko, line 63: Please do not ignore fails here

modules/balancer/dataplane/l4/gre.c
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	uint16_t outer_ip_size = is_outer_ipv6 ? sizeof(struct rte_ipv6_hdr)
```
GeorgyKirichenko, line 66: What about ip options?

modules/balancer/dataplane/l4/gre.c
```diff
@@ -0,0 +1,88 @@
...
+	uint16_t cksum = ~hdr->hdr_checksum;
+	cksum = csum_minus(cksum, old_total_length);
+	cksum = csum_minus(cksum, old_ttl_proto);
+	cksum = csum_plus(cksum, hdr->total_length);
+	cksum = csum_plus(cksum, new_ttl_proto);
+	hdr->hdr_checksum = (cksum == 0xffff) ? cksum : ~cksum;
+}
+
+void
+insert_gre_header(
```
GeorgyKirichenko, line 56: I do not like the idea of inserting headers, I think it should be a sort of `encap` functions like `encap_gre` or even `encap_gre_ip`.
Also I think the encapsulation should live somewhere inside `lib/dataplane/packet`

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+	uint16_t vs_flags = pkt_ctx->matched_vs->flags;
+
+	/* Clamp MSS for IPv6 SYN packets before encapsulation. */
+	if (unlikely(vs_flags & balancer_vs_fix_mss)) {
+		fix_mss_ipv6(packet);
+	}
+
+	encapsulate_ipv6(packet, real);
+
+	if (unlikely(vs_flags & balancer_vs_gre)) {
```
GeorgyKirichenko, line 247: I do not think we should make a prejudgement here about service types.

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+tunnel_ipv4_packet(struct l4_packet_context *pkt_ctx) {
+	struct packet *packet = pkt_ctx->packet;
+	struct balancer_real *real = pkt_ctx->resolved_real;
+	uint16_t vs_flags = pkt_ctx->matched_vs->flags;
+
+	/* No MSS clamping for inner IPv4. */
+
+	encapsulate_ipv4(packet, real);
+
+	if (unlikely(vs_flags & balancer_vs_gre)) {
```
GeorgyKirichenko, line 221: I do not think we should make a prejudgement here about service types.

modules/balancer/dataplane/l4/tunnel.c
```diff
@@ -0,0 +1,253 @@
...
+		bool is_outer_ipv6 = real->flags & balancer_real_ipv6;
+		insert_gre_header(packet, is_outer_ipv6, true);
```
GeorgyKirichenko, line 223: I think the logic should be structured in the following manner:
`
if (use_gre)
  encap_gre();
encap_ip()
`
or
`
if (use_gre)
  encap_gre_ip();
else 
  encap_ip();
`

please refer to comments inside `l4/gre.c`

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+			tunnel_ipv6_packet(pkt_ctx);
+		} else {
+			tunnel_ipv4_packet(pkt_ctx);
+		}
+
+		packet_front_output(packet_front, pkt_ctx->packet);
+
+		context->l4_stats->outgoing_packets += 1;
+		context->common_stats->outgoing_packets += 1;
+		context->common_stats->outgoing_bytes +=
```
GeorgyKirichenko, line 398: Please do not invade into `dpdk` internals, there is `packet_size` routine for that purpose.

GeorgyKirichenko (changes requested):

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+ * 1. Batch VS lookup — classify all packets at once.
+ * 2. Group matched packets by VS.
+ * 3. Batch ACL per group — drop packets that fail.
+ *
+ * Returns the number of packets that passed both checks.
+ * is_ipv6 is a compile-time constant so the compiler
+ * eliminates the dead branches after inlining.
+ */
+static size_t
+match_and_filter(
```
GeorgyKirichenko, line 116: The name is confusing, maybe we should use something like `process_vs` or `service_lookup`

modules/balancer/dataplane/context.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <stdint.h>
+
+struct packet_front;
+struct balancer_packet_handler;
+struct counter_storage;
+
+struct worker_context {
```
GeorgyKirichenko, line 9: The name is confusing

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+	}
+
+	/* Drop unmatched, record indices and vs_ids of the rest. */
+	uint8_t order[packets_count];
+	uint32_t vs_ids[packets_count];
+	size_t matched_count = 0;
+
+	for (size_t i = 0; i < packets_count; ++i) {
+		uint32_t vs_id = vs_results[i];
+		if (unlikely(vs_id == FILTER_RULE_INVALID)) {
```
GeorgyKirichenko, line 150: I think the prejudgement is incorrect lets compiler to decide what is the best choice

modules/balancer/dataplane/l4/group.c
```diff
@@ -0,0 +1,60 @@
+#include "common/likely.h"
+#include <string.h>
+
+#include "group.h"
+
+/*
+ * Maximum number of distinct VS IDs we track per batch.
```
GeorgyKirichenko, line 7: On high loaded balancers with huge amount of services the amount of services  after grouping should be near the packet batch size.

modules/balancer/dataplane/l4/group.c
```diff
@@ -0,0 +1,60 @@
...
+#define MAX_GROUPS 16
+
+void
+group_by_id(uint32_t *vs_ids, uint8_t *order, size_t count) {
+	/* Pass 1: collect unique VS IDs and count per group. */
+	uint32_t unique_vs[MAX_GROUPS];
+	size_t counts[MAX_GROUPS];
+	size_t ngroups = 0;
+
+	for (size_t i = 0; i < count; ++i) {
```
GeorgyKirichenko, line 21: The another question - why not to just sort vs_ids(with according swap of order array)?

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+			ADDR_OF(&context->packet_handler->ipv4_vs_matcher),
+			ipv4_vs_matcher,
+			packets,
+			vs_results,
+			packets_count
+		);
+	}
+
+	/* Drop unmatched, record indices and vs_ids of the rest. */
+	uint8_t order[packets_count];
```
GeorgyKirichenko, line 144: There is an implicit dependency on BATCH_SIZE (with is 64 items now). In case of batching logic change and/or increase in the BATCH_SIZE value the code will fail.

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+	/* Run ACL per VS group. */
+	size_t total_passed = 0;
+	size_t pkt_idx = 0;
+	struct packet *group_pkts[matched_count];
+	while (pkt_idx < matched_count) {
+		uint32_t cur_vs_id = vs_ids[pkt_idx];
+
+		/* Collect contiguous packets for this VS. */
+		size_t group_size = 0;
+		while (pkt_idx < matched_count && vs_ids[pkt_idx] == cur_vs_id
```
GeorgyKirichenko, line 184: You already evaluated each group size so you have not any thoughts why not to reuse this knowledge here

modules/balancer/dataplane/types/vs.h
```diff
@@ -0,0 +1,199 @@
...
+	uint32_t reals_count;
+
+	uint64_t stable_idx;
+	uint64_t counter_id;
+
+	struct balancer_real_selector *selector;
+	struct filter *acl;
+
+	/*
+	 * Controlplane limits the number of rules per VS to billion,
```
GeorgyKirichenko, line 68: Please do not make any assumptions about controlpane behavior inside dataplane code

modules/balancer/dataplane/types/vs.h
```diff
@@ -0,0 +1,199 @@
...
+	 * so we can allocate a single array for all rules
+	 * in the shared memory without fragmentation issues.
+	 */
+	uint64_t *rule_counter_ids;
+
+	uint16_t flags;
+
+	/* ---Controlplane data --- */
+
+	/* To separate dataplane and controlplane cachelines. */
```
GeorgyKirichenko, line 78: Dataplane structures should not have any controlplane embedded values

modules/balancer/dataplane/types/vs.h
```diff
@@ -0,0 +1,199 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "common/network.h"
+
+#include "selector.h"
+
+enum balancer_vs_flags {
+	balancer_vs_pure_l3 = 1u << 0,
```
GeorgyKirichenko, line 12: Not a dataplane behavior flag

modules/balancer/dataplane/types/vs.h
```diff
@@ -0,0 +1,199 @@
...
+	balancer_vs_fix_mss = 1u << 1,
+	balancer_vs_gre = 1u << 2,
+	balancer_vs_ops = 1u << 3,
+	balancer_vs_wlc = 1u << 4,
```
GeorgyKirichenko, line 16: Not a dataplane behavior flag

modules/balancer/dataplane/types/vs.h
```diff
@@ -0,0 +1,199 @@
...
+	balancer_vs_removed = 1u << 5,
+	balancer_vs_round_robin = 1u << 6,
```
GeorgyKirichenko, line 18: Not a dataplane behavior flag

modules/balancer/dataplane/types/vs.h
```diff
@@ -0,0 +1,199 @@
...
+
+#include "selector.h"
+
+enum balancer_vs_flags {
+	balancer_vs_pure_l3 = 1u << 0,
+	balancer_vs_fix_mss = 1u << 1,
+	balancer_vs_gre = 1u << 2,
+	balancer_vs_ops = 1u << 3,
+	balancer_vs_wlc = 1u << 4,
+	balancer_vs_removed = 1u << 5,
```
GeorgyKirichenko, line 17: Not a dataplane behavior flag

modules/balancer/dataplane/types/vs.h
```diff
@@ -0,0 +1,199 @@
...
+};
+
+struct balancer_vs_allowed_source {
+	struct net *nets;
+	uint32_t nets_count;
+
+	struct filter_port_range *port_ranges;
+	uint32_t port_ranges_count;
+
+	char tag[balancer_vs_acl_max_tag_len + 1];
```
GeorgyKirichenko, line 37: I do not understand the tag meaning

GeorgyKirichenko: If it is about counter name length - use counters please - any counters should be accessible through standard counters API

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+	for (size_t i = 0; i < group_size; ++i) {
+		uint32_t rule_idx = acl_results[i];
+		if (unlikely(rule_idx == FILTER_RULE_INVALID)) {
+			vs_stats->packet_src_not_allowed += 1;
+			packet_front_drop(context->packet_front, group_pkts[i]);
+			continue;
+		}
+
+		vs_stats->incoming_packets += 1;
+		vs_stats->incoming_bytes += group_pkts[i]->mbuf->pkt_len;
```
GeorgyKirichenko, line 80: DPDK internals intrusion

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+	struct l4_packet_context *pkt_ctxs,
+	size_t pkt_ctx_count,
+	struct balancer_session_timeouts *timeouts,
+	bool is_ipv6
+) {
+	const size_t prefetch_distance = 2;
+
+	for (size_t pkt_idx = 0; pkt_idx < pkt_ctx_count; ++pkt_idx) {
+		if (pkt_idx + prefetch_distance < pkt_ctx_count) {
+			rte_prefetch0(rte_pktmbuf_mtod(pkt_ctxs[pkt_idx + prefetch_distance].packet->mbuf, void *));
```
GeorgyKirichenko, line 310: Justification of prefetch and prefetch value is required at least in form of a commentary

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+		);
+
+		/* Dont track stats for rules with empty tag. */
+		if (rule_counter != NULL) {
+			*rule_counter += 1;
+		}
+
+		pkt_ctxs[passed++] = (struct l4_packet_context){
+			.packet = group_pkts[i],
+			.matched_vs = vs,
```
GeorgyKirichenko, line 96: The only use of structure I found is obtaining `stable_idx` - why not to store this value?

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+		struct l4_packet_context *pkt_ctx = &pkt_ctxs[pkt_idx];
+		pkt_ctx->session_id.vs_stable_idx =
+			pkt_ctx->matched_vs->stable_idx;
+
+		/*
+		 * Zero client_ip + padding so the session ID hashes
+		 * deterministically. extract_network_metadata overwrites
+		 * the relevant prefix (4 bytes for IPv4, 16 for IPv6).
+		 */
+		__builtin_memset(
```
GeorgyKirichenko, line 322: Please use some wrapper or DPDK API function

GeorgyKirichenko, line 322: Implicit dependency on the structure internal layout so I would suggest to zero-initialize the whole structure - following is a subject of a compiler to optimize

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+}
+
+static void
+extract_ipv4_metadata(struct l4_packet_context *pkt_ctx) {
+	struct rte_ipv4_hdr *ipv4_hdr = rte_pktmbuf_mtod_offset(
+		pkt_ctx->packet->mbuf,
+		struct rte_ipv4_hdr *,
+		pkt_ctx->packet->network_header.offset
+	);
+	__builtin_memcpy(
```
GeorgyKirichenko, line 210: Wrapper

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+	struct rte_tcp_hdr *hdr = rte_pktmbuf_mtod_offset(
+		pkt_ctx->packet->mbuf,
+		struct rte_tcp_hdr *,
+		pkt_ctx->packet->transport_header.offset
+	);
+
+	uint8_t flags = hdr->tcp_flags;
+
+	pkt_ctx->session_id.client_port = hdr->src_port;
+	pkt_ctx->session_timeout = tcp_session_timeout(timeouts, flags);
```
GeorgyKirichenko, line 266: Code logic breakage - extract transport metadata should not update timeout and reschedule items.

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+) {
+	struct rte_udp_hdr *hdr = rte_pktmbuf_mtod_offset(
+		pkt_ctx->packet->mbuf,
+		struct rte_udp_hdr *,
+		pkt_ctx->packet->transport_header.offset
+	);
+
+	pkt_ctx->session_id.client_port = hdr->src_port;
+	pkt_ctx->session_timeout = timeouts->udp;
+	pkt_ctx->can_reschedule = true;
```
GeorgyKirichenko, line 284: In the current implementation UDP flag is an attribute of a service - you could have UDP client packet only if the service is UDP too

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+}
+
+static void
+extract_transport_metadata(
+	struct l4_packet_context *pkt_ctx,
+	struct balancer_session_timeouts *timeouts
+) {
+	if (pkt_ctx->packet->transport_header.type == IPPROTO_TCP) {
```
GeorgyKirichenko, line 292: This should be resolved directly from service attributes.

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+		pkt_ctx_count,
+		&context->packet_handler->session_timeouts,
+		is_ipv6
+	);
+
+	struct balancer_session_table *session_table =
+		ADDR_OF(&context->packet_handler->session_table);
+
+	rcu_t *reals_selector_guard = &context->packet_handler->rcu;
+	rcu_read_begin(reals_selector_guard, context->worker_idx);
```
GeorgyKirichenko, line 454: I don't like the idea of ​​replacing any configuration data under the feet of data processing workflows.

GeorgyKirichenko: The point is to form a new reals ring and then atomically swap the pointers. An then the old one could be freed after configuration generation change

modules/balancer/dataplane/l4/handle.c
```diff
@@ -0,0 +1,490 @@
...
+
+static void
+select_reals(
+	struct worker_context *context,
+	struct l4_packet_context *pkt_ctxs,
+	size_t pkt_ctx_count,
+	struct balancer_session_table *session_table,
+	uint64_t current_table_gen
+) {
+	const size_t prefetch_distance = 4;
```
GeorgyKirichenko, line 341: Justification is required

modules/balancer/dataplane/l4/handle.c
GeorgyKirichenko, line 1: The overall comment - there are many items encapsulated inside packet context what complicates internal dependency tracking - what and where is used or not. I would prefer to pass all such data explicitly.

modules/balancer/dataplane/l4/resolve.c
GeorgyKirichenko, line 1: The file name is confusing - it is not easy to understand what is resolved here - vs, rs, session?

modules/balancer/dataplane/l4/resolve.c
```diff
@@ -0,0 +1,297 @@
...
+	struct l4_packet_context *pkt_ctx,
+	struct balancer_vs *vs
+) {
+	uint32_t real_idx = selector_select(
+		ADDR_OF(&vs->selector),
+		context->worker_idx,
+		pkt_ctx->packet->hash
+	);
+	if (unlikely(real_idx == INVALID_REAL_IDX)) {
+		pkt_ctx->matched_vs_stats->no_reals += 1;
```
GeorgyKirichenko, line 151: I think the counter should be increased at upper level - when real lookup fails - without any dependency on a lookup algorithm used

modules/balancer/dataplane/l4/resolve.c
```diff
@@ -0,0 +1,297 @@
...
+		atomic_load_explicit(&selector->ring_id, memory_order_acquire);
+	struct balancer_ring *ring = &selector->rings[ring_id];
+
+	/*
+	 * Here branch predictor works well,
+	 * because use_rr is long-term variable.
+	 * So, we dont use ternary operator here.
+	 */
+	uint64_t idx;
+	if (selector->use_rr) {
```
GeorgyKirichenko, line 52: Confusing logic - in the one hand the routine is called from the `_ops` scheduler but hash is passed - in the other hand there is `is_rr` flag used.

modules/balancer/dataplane/l4/resolve.c
```diff
@@ -0,0 +1,297 @@
...
+static struct balancer_real *
+resolve_real_ops(
+	struct worker_context *context,
+	struct l4_packet_context *pkt_ctx,
+	struct balancer_vs *vs
+) {
+	uint32_t real_idx = selector_select(
+		ADDR_OF(&vs->selector),
+		context->worker_idx,
+		pkt_ctx->packet->hash
```
GeorgyKirichenko, line 148: I assume you should use increasing sequence instead of hash to use OPS.

modules/balancer/dataplane/l4/resolve.c
```diff
@@ -0,0 +1,297 @@
...
+	struct balancer_ring *ring = &selector->rings[ring_id];
+
+	/*
+	 * Here branch predictor works well,
+	 * because use_rr is long-term variable.
+	 * So, we dont use ternary operator here.
+	 */
+	uint64_t idx;
+	if (selector->use_rr) {
+		idx = selector->workers[worker].value++;
```
GeorgyKirichenko, line 53: What about size of RCU_WORKERS and worker_idx - we could have a huge amount of them

modules/balancer/dataplane/l4/resolve.c
```diff
@@ -0,0 +1,297 @@
...
+	 * At this point we need to assign a new real to the session slot.
+	 * This happens when:
+	 * - SESSION_CREATED: no prior session existed (new flow).
+	 * - SESSION_FOUND but real is stale (disabled/removed).
+	 *
+	 * Only connection-initiating packets may create or overwrite
+	 * sessions. Non-initiating packets (TCP non-SYN) are dropped
+	 * and the session slot is freed.
+	 */
+	if (unlikely(!pkt_ctx->can_reschedule)) {
```
GeorgyKirichenko, line 280: Why do you want to pessimise UDP services

modules/balancer/dataplane/l4/resolve.c
```diff
@@ -0,0 +1,297 @@
...
+		vs_stats->not_rescheduled_packets += 1;
+		st_remove_session(session_state);
```
GeorgyKirichenko, line 282: Why the session should be removed? Packet retransmission will change the logic.

modules/balancer/dataplane/l4/resolve.c
```diff
@@ -0,0 +1,297 @@
...
+	if (unlikely(
+		    session_state->real_stable_idx != real->stable_idx ||
+		    real_is_removed(real)
+	    )) {
+		vs_stats->real_is_removed += 1;
+		return NULL;
+	}
+
+	if (unlikely(!real_is_enabled(real))) {
+		struct balancer_real_stats *real_stats = real_get_stats(
```
GeorgyKirichenko, line 128: Some services allow to reuse disabled reals for established session

GeorgyKirichenko (changes requested):

lib/dataplane/time/clock.c
GeorgyKirichenko, line 1: Another PR

lib/controlplane/agent/agent.c
GeorgyKirichenko, line 1: Another PR

filter/compiler/net6_fast.c
GeorgyKirichenko, line 1: Another PR

controlplane/yncp/director.go
GeorgyKirichenko, line 1: Another PR

modules/acl/controlplane/metrics.go
GeorgyKirichenko, line 1: Another PR

modules/route/cli/route/src/main.rs
GeorgyKirichenko, line 1: Another PR

modules/balancer/dataplane/l4/resolve.c
GeorgyKirichenko, line 280: What if a service prohibits real reschedule but has no session?

GeorgyKirichenko: I understand that SYN packet enables reschedule somewhere in the code but the logic here looks confusing

GeorgyKirichenko (changes requested):

tests/functional/main/balancer_test.go
GeorgyKirichenko, line 1: Another PR

egnees:
Closing in favor of a from-scratch rewrite of modules/balancer2 (dataplane, control-plane, and CLI). The new implementation is going through code review now and will land as a stack of smaller, focused PRs instead of one large one. Thank you for all the review feedback here — a lot of it directly shaped the new implementation.

PR #662, opened 2026-04-21
fix(encap): propagate prepend errors and fix typo

egnees:
### Bug

`packet_ip4_encap` and `packet_ip6_encap` discarded the return value of
`packet_network_prepend` and unconditionally returned `0`. If
`rte_pktmbuf_prepend` failed (e.g. headroom exhaustion), callers could
not detect the failure and would proceed with a malformed packet.

### Changes

- **`packet_ip4_encap`** — return the result of `packet_network_prepend`
  directly instead of dropping it and returning `0`.
- **`packet_ip6_encap`** — same fix.
- **`packet_network_prepend`** — annotate the `rte_pktmbuf_prepend`
  failure path with `unlikely()` to match the hot-path assumption.
- **Comment typo** — `previos heade` → `previos header`.

### DIFF
--- a/lib/dataplane/packet/encap.c
+++ b/lib/dataplane/packet/encap.c
@@ -31,7 +31,7 @@ packet_network_prepend(
 ) {
 	struct rte_mbuf *mbuf = packet_to_mbuf(packet);
 
-	if (rte_pktmbuf_prepend(mbuf, size) == NULL)
+	if (unlikely(rte_pktmbuf_prepend(mbuf, size) == NULL))
 		return -1;
 	memmove(rte_pktmbuf_mtod(mbuf, char *),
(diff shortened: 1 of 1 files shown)


lib/dataplane/packet/encap.c
```diff
@@ -31,7 +31,7 @@ packet_network_prepend(
 ) {
 	struct rte_mbuf *mbuf = packet_to_mbuf(packet);
 
-	if (rte_pktmbuf_prepend(mbuf, size) == NULL)
+	if (unlikely(rte_pktmbuf_prepend(mbuf, size) == NULL))
```
3Hren, line 34: Out of the scope. And needs proofing.

egnees:
Reopened in [664](https://github.com/yanet-platform/yanet2/pull/664).

PR #663, opened 2026-04-21
feat(packet): add packet_fix_mss for IPv6 TCP SYN MSS fixup

egnees:
## Summary

- Introduce `lib/dataplane/packet/mss.{c,h}` with `packet_fix_mss(packet, clamp_mss, insert_mss)` — a single-pass MSS normalizer for IPv6 TCP SYN packets.
- Two independent MSS parameters (clamp ceiling vs. insert fallback).
- Structured `enum packet_fix_mss_result` so callers can distinguish `ok`, `malformed`, and `no_headroom` for metrics instead of a silent void return.

## Test plan

- [x] Clamp path: IPv6 TCP SYN with MSS > clamp → option rewritten, TCP checksum valid
- [x] No-op path: IPv6 TCP SYN with MSS ≤ clamp → packet unchanged
- [x] Insert path: IPv6 TCP SYN with no MSS → option inserted, TCP checksum + IPv6 payload length updated
- [x] Skip paths: IPv4, non-TCP, non-SYN, SYN+RST → untouched, returns `ok`
- [x] Malformed options (`len == 0`, declared length overruns data offset, mid-options EOL) → returns `malformed`
- [x] Zero-headroom mbuf with no MSS option → returns `no_headroom`

### DIFF
--- a/lib/dataplane/packet/mss.c
+++ b/lib/dataplane/packet/mss.c
@@ -1,0 +1,287 @@
+#include <netinet/in.h>
+#include <string.h>
+
+#include <rte_ether.h>
+#include <rte_ip.h>
+#include <rte_mbuf.h>
+#include <rte_tcp.h>
+
+#include "common/checksum.h"
+
+#include "lib/dataplane/packet/packet.h"
+#include "rte_branch_prediction.h"
+
+#include "mss.h"
+
+#define TCP_OPTION_KIND_EOL 0
+#define TCP_OPTION_KIND_NOP 1
+#define TCP_OPTION_KIND_MSS 2
+#define TCP_OPTION_MSS_LEN 4
+
+/* Minimum length of any variable-length TCP option: kind + len bytes. */
+#define TCP_OPTION_MIN_LEN 2
+
+/* Max TCP header length in bytes (data_off is 4 bits of 32-bit words: 15*4). */
+#define TCP_HDR_LEN_MAX 60
+
+/* One 32-bit-word step expressed in the raw data_off byte (high nibble). */
+#define TCP_DATA_OFF_ONE_WORD (1u << 4)
+
+struct tcp_option {
+	uint8_t kind;
+	uint8_t len;
+	uint8_t data[];
+} __attribute__((__packed__));
+
+/*
+ * Return the TCP header if the packet is IPv6 and carries a TCP SYN or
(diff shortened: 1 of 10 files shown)


lib/dataplane/packet/mss.h
```diff
@@ -0,0 +1,33 @@
...
+ *     left untouched.
+ *   - If an MSS option is present and its value exceeds `clamp_mss`, the
+ *     option is rewritten to `clamp_mss`. Lower values are kept as-is.
+ *   - If no MSS option is present, a new option with value `insert_mss`
+ *     is inserted immediately after the fixed TCP header.
+ *
+ * TCP checksum and IPv6 payload length are updated as needed.
+ */
+enum packet_fix_mss_result
+packet_fix_mss(struct packet *packet, uint16_t clamp_mss, uint16_t insert_mss);
```
GeorgyKirichenko: I would suggest to rename packet_fix_mss to packet_set_mss

lib/dataplane/packet/mss.c
```diff
@@ -0,0 +1,287 @@
...
+ * Return the TCP header if the packet is IPv6 and carries a TCP SYN or
+ * SYN+ACK (i.e. SYN is set and RST is clear); NULL otherwise.
+ *
+ * Precondition: parse_packet guarantees that whenever
+ * transport_header.type == IPPROTO_TCP, the mbuf holds at least
+ * `transport_header.offset + sizeof(struct rte_tcp_hdr)` bytes,
+ * so the fixed 20-byte TCP header is safe to dereference here.
+ */
+static struct rte_tcp_hdr *
+get_syn_tcp_header(struct packet *packet) {
```
GeorgyKirichenko: I think the name is a little bit confusing because the routine is for IPv6 packets only

GeorgyKirichenko (changes requested):

lib/dataplane/packet/mss.h
egnees: Renamed.

lib/dataplane/packet/mss.c
egnees: Renamed to `tcp_hdr`.

egnees: Leave explaining comments.

GeorgyKirichenko (approved):

PR #665, opened 2026-04-21
feat(encap): add packet_ip4/ip6_encap_gre

egnees:
## Summary
- Add `packet_ip4_encap_gre` and `packet_ip6_encap_gre` in `lib/dataplane/packet/encap.{h,c}`.
- Each routine prepends an outer IPv4/IPv6 header (next proto = GRE) followed by a GRE header in a single packed struct and a single `packet_network_prepend` call — no transient intermediate EtherType state, no extra copy of the outer IP header.
- Inner DSCP/TTL (for IPv4 outer) and traffic class/hop limit (for IPv6 outer) are inherited from the inner header; the GRE `proto` field is set from the inner EtherType (IPv4 or IPv6). The outer IPv4 checksum is computed.

## Test plan
- [x] Add unit tests covering all four combinations of inner × outer (v4/v6 × v4/v6), asserting:
  - outer IP `proto` / `next-header` = `IPPROTO_GRE`,
  - outer length fields include the GRE header,
  - GRE `proto` matches the inner EtherType,
  - outer IPv4 checksum is valid,
  - `packet->network_header.offset` / `transport_header.offset` are updated correctly.


GeorgyKirichenko (dismissed):

lib/dataplane/packet/encap.c
```diff
@@ -309,3 +346,62 @@ packet_ip6_encap_udp(
...
+
+	rte_memcpy(&outer.ip.src_addr, src, NET4_LEN);
+	rte_memcpy(&outer.ip.dst_addr, dst, NET4_LEN);
+	outer.ip.version_ihl = 0x45;
+
+	int inner_size = fill_outer_ip4_from_inner(&outer.ip, packet);
+	if (inner_size < 0) {
+		return -1;
+	}
+	outer.ip.total_length = rte_cpu_to_be_16(sizeof(outer) + inner_size);
```
GeorgyKirichenko, line 369: Should the `gre` header size be accounted here?

lib/dataplane/packet/encap.c
```diff
@@ -309,3 +346,62 @@ packet_ip6_encap_udp(
...
+	struct {
+		struct rte_ipv6_hdr ip;
+		struct rte_gre_hdr gre;
+	} __rte_packed outer;
+
+	rte_memcpy(&outer.ip.src_addr, src, NET6_LEN);
+	rte_memcpy(&outer.ip.dst_addr, dst, NET6_LEN);
+
+	int inner_size = fill_outer_ip6_from_inner(&outer.ip, packet);
+	outer.ip.payload_len = rte_cpu_to_be_16(sizeof(outer.gre) + inner_size);
```
GeorgyKirichenko, line 403: Should the `gre` header size be accounted here?

lib/dataplane/packet/encap.c
GeorgyKirichenko: Oh, you are right, sorry

GeorgyKirichenko (dismissed):

lib/dataplane/packet/encap.c
egnees: It is.

PR #694, opened 2026-04-29
feat(forward): per-rule metrics and interfaces

kt0ns:


modules/forward/controlplane/metrics.go
```diff
@@ -0,0 +1,80 @@
+package forward
+
+import (
+	"github.com/yanet-platform/yanet2/common/commonpb"
+)
+
+func (m *ForwardService) Metrics() ([]*commonpb.Metric, error) {
```
saushew, line 7: This method is currently unused

controlplane/ffi/agent.go
```diff
@@ -77,6 +77,40 @@ func (m ShmDeviceConfig) AsRawPtr() unsafe.Pointer {
...
+	ModuleAgent
+	FunctionAgent
+	PipelineAgent
+	DeviceAgent
+	CleanUp() error
+	Close() error
+}
+
+type ModuleAgent interface {
+	UpdateModules(modules []ModuleConfig) error
```
saushew, line 86: I don't see any cases where this abstraction could be useful. Some of the methods accept specific package's types (type `ModuleConfig` for example), which makes complete abstraction impossible.

controlplane/ffi/shm.go
```diff
@@ -136,24 +166,24 @@ func (m *SharedMemory) AgentsAttach(
...
 
 // DPConfig represents a handle to dataplane configuration.
-type DPConfig struct {
+type DPConfigImpl struct {
 	ptr *C.struct_dp_config
 }
 
-func NewDPConfigFromRaw(ptr unsafe.Pointer) *DPConfig {
-	return &DPConfig{ptr: (*C.struct_dp_config)(ptr)}
+func NewDPConfigFromRaw(ptr unsafe.Pointer) *DPConfigImpl {
```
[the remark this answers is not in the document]
saushew: AI is right. The rule of thumb: if you need the `Impl` suffix somewhere, just because the desired name already in use, then something going wrong.

modules/forward/controlplane/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+			for _, workerVals := range counter.Values {
+				if len(workerVals) > 0 {
+					cntPackets += workerVals[0]
+				}
+				if len(workerVals) > 1 {
+					bytes += workerVals[1]
+				}
+			}
+
+			if cntPackets == 0 && bytes == 0 {
```
saushew: Both packets and bytes can be equal to 0. These values are legal to push to the monitoring

modules/forward/controlplane/metrics.go
kt0ns: fixed

modules/forward/controlplane/metrics.go
kt0ns: fixed

controlplane/ffi/agent.go
kt0ns: Such interfaces are no longer implemented primarily for abstraction, but rather for separating areas of responsibility and facilitating code comprehension and understanding of problems that they solve.
High granularity in interface decomposition has been removed, but separation into `DPAgent` and `DPDataProvider` has been saved

controlplane/ffi/shm.go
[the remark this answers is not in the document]
kt0ns: fixed.
Now internal implementation is non-exportable

PR #700, opened 2026-04-30
refactor(balancer): Go bindings for controlplane API

egnees:
## Summary
This PR introduces thin Go bindings for the balancer controlplane API.


modules/balancer2/bindings/go/cbalancer2/cgo.go
```diff
@@ -0,0 +1,464 @@
...
+// memory. A chain references session tables but does not own them; the
+// referenced tables must outlive the chain.
+type SessionTableChain struct {
+	ptr *C.struct_balancer_session_table_chain
+}
+
+// Install installs the balancer handle in the dataplane. If a balancer with
+// the same name is already installed, it is replaced and the previous handle
+// becomes unused; the caller is responsible for freeing it.
+func (b *Balancer) Install(agent *ffi.Agent) error {
```
3Hren: Receiver naming: per `CLAUDE.md` ("Receiver names: always `m`. No type-letter mnemonics."), all receivers in this package should be `m`. Canonical bindings already follow this even for non-`ModuleConfig` types — e.g. `func (m ForwardMode) toC()` in `modules/forward/bindings/go/cforward/safe.go:22`, `func (m *ForwardRule) cBuild(...)` at `:65`.

This applies to every receiver in this file: `Balancer` (87, 102, 113, 138), `SessionTable` (162), `SessionTableChain` (176, 190, 206), `AllowedSources`/`cAllowedSources` (225, 239), `VSConfig`/`cVSConfig` (254, 299), `RealConfig` (306), `SessionTimeouts` (331). I'm leaving the comment on the first occurrence to avoid spamming.

modules/balancer2/bindings/go/cbalancer2/cgo.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+}
+
+// UpdateVSRealWeights updates per-real weights for the VS at the given index.
+// The weights slice must have length equal to the number of reals configured
+// for the VS and be indexed in the same order they were passed at VS
+// creation.
+func (b *Balancer) UpdateVSRealWeights(vsIdx uint32, weights []uint32) error {
+	if b.ptr == nil {
+		return errors.New("balancer handle is freed")
```
3Hren: Recommend dropping these guards entirely — canonical bindings (`forward`, `decap`, `route`) don't carry them on update methods. They just pass the (possibly nil) C pointer through and let the C side handle it (see `forward_module_config_update` use at `cforward/cgo.go:62`). Less code, no Go-vs-C divergence, and no risk of the wording inconsistency this PR currently has.

If kept anyway, the wording is inconsistent across guards on the same `<receiver>.ptr == nil` condition:

- `Install` (89) → `"balancer handle is nil"`
- `UpdateVSRealWeights` (115) → `"balancer handle is freed"`
- `UpdateVSRealStates` (140) → `"balancer handle is nil"`
- `SessionTableChain.PushFront` (192) → `"chain is freed"`
- `SessionTableChain.PopBack` (208) → `"chain is nil"`
- `createBalancer` (349) → `"chain is nil"`
- `createSessionTableChain` (413) → `"front table is nil"`

Pick one wording — `"… is freed"` is more accurate, since the only way `ptr` becomes nil after a successful constructor is via `Free`.

modules/balancer2/bindings/go/cbalancer2/cgo.go
egnees: Renamed all receivers to `m`.

modules/balancer2/bindings/go/cbalancer2/cgo.go
egnees: Removed nil guards. The C API must guarantee error on passing nil entity.

3Hren (approved):

PR #720, opened 2026-05-05
feat(acl): skip dataplane update when rules are unchanged

semenovad:
Skip ACL dataplane update when rules are unchanged

### DIFF
--- a/modules/acl/controlplane/service.go
+++ b/modules/acl/controlplane/service.go
@@ -8,6 +8,8 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"google.golang.org/protobuf/proto"
+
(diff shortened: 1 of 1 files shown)


modules/acl/controlplane/service.go
```diff
@@ -158,16 +172,20 @@ func (m *ACLService) UpdateConfig(
...
 	defer m.mu.Unlock()
 	defer tracker.Fix()
 
+	if existing, ok := m.configs[name]; ok && rulesEqual(existing.rules, req.Rules) {
+		return &aclpb.UpdateConfigResponse{}, nil
+	}
+
+	rules, err := convertRules(req.Rules)
+	if err != nil {
+		return nil, err
```
3Hren, line 186: `// TODO: invalid argument error here` was lost.

3Hren (approved):

PR #753, opened 2026-05-07
fix(ci): run build and functional tests on Rust changes

egnees:
The `Build YANET` workflow gates the functional tests but its `paths:` filter has no Rust patterns, so CLI-only PRs never exercise the release binaries against the functional suite.

- Add `**.rs`, `**/Cargo.toml`, `**/Cargo.lock` to the `push` and `pull_request` `paths:` in `.github/workflows/build.yml`.
- Trigger surface now matches `rust-check.yml`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,9 @@ on:
       - "go.mod"
       - "go.sum"
       - "Makefile"
+      - "**.rs"
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
   pull_request:
     branches: ["main"]
     paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -11,6 +11,9 @@ on:
       - "go.mod"
       - "go.sum"
       - "Makefile"
+      - "**.rs"
+      - "**/Cargo.toml"
```
3Hren, line 15: Seems like functional tests, which only depend on CLI, cannot be made broken from Cargo* changes, or am I wrong?

egnees: For example, we might remove a cli-crate from the members list in `Cargo.toml`, or add a sibling crate in one PR and replace the old crate with it in a follow-up — with no code changes, but the CLI binary can be changed. In theory, such PRs wouldn't need to touch related files like the `Makefile`.

3Hren (approved):

PR #768, opened 2026-05-12
fix(route): free cp_module resources on config data init failure

3Hren:
When `route_module_config_data_init` failed after a successful `cp_module_init`, the counter registry, registered counters, perf-counter table, and devices array allocated by `cp_module_init` were never released.

### DIFF
--- a/modules/route/tests/meson.build
+++ b/modules/route/tests/meson.build
@@ -1,0 +1,45 @@
+includes = include_directories('../../../', '../../../lib')
+
+lib_dependencies = [
+    lib_common_dep,
+    lib_logging_dep,
+    lib_errors_dep,
+    lib_config_cp_dep,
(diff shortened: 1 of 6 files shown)


modules/route/tests/meson.build
```diff
@@ -0,0 +1,45 @@
...
+    lib_agent_cp_dep,
+    lib_counters_dep,
+    lib_pipeline_dp_dep,
+    libdpdk_inc_dep,
+    lib_route_cp_dep,
+    lib_lib_utils_dep,
+    lib_yanet_mock_dep,
+]
+
+modules_dep = [
```
3Hren, line 17: Because of strange mock dependencies.

GeorgyKirichenko (approved):

PR #776, opened 2026-05-13
fix(acl): heap corruption on update error path and NULL deref in free

3Hren:


modules/acl/api/controlplane.c
```diff
@@ -147,11 +145,17 @@ acl_module_config_free(struct cp_module *cp_module) {
...
-	memory_bfree(
-		&agent->memory_context,
-		cp_module,
-		sizeof(struct acl_module_config)
-	);
+	struct agent *agent = ADDR_OF(&cp_module->agent);
+	// FIXME: remove the check as agent should be assigned
+	// cp_module_init can fail before agent is set; acl_module_config_init
+	// calls free on that path.
+	if (agent != NULL) {
```
3Hren, line 152: Better to block on `cp_module_fini` refactoring first.

PR #782, opened 2026-05-14
refactor: adapt cp_chain to memory pattern

3Hren:
Explicit new/free/init/fini. Keep allocator pointer inside.

### DIFF
--- a/lib/controlplane/config/cp_chain.c
+++ b/lib/controlplane/config/cp_chain.c
@@ -2,6 +2,7 @@
 
 #include "controlplane/config/zone.h"
 
+#include <assert.h>
 #include <stdio.h>
 
 static inline size_t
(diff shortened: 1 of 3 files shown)


lib/controlplane/config/cp_chain.c
```diff
@@ -11,8 +12,36 @@ cp_chain_alloc_size(uint64_t length) {
...
 
 struct cp_chain *
-cp_chain_create(
-	struct memory_context *memory_context,
+cp_chain_new(struct memory_context *memory_context, uint64_t length) {
+	size_t alloc_size = cp_chain_alloc_size(length);
+	struct cp_chain *self =
+		(struct cp_chain *)memory_balloc(memory_context, alloc_size);
+	if (self == NULL) {
+		return NULL;
```
3Hren, line 20: No `yanet_error_add` here, because in `*_new` functions this is the only reason of failure.

GeorgyKirichenko (approved):

PR #794, opened 2026-05-14
Non-terminating rules refactoring

GeorgyKirichenko:
The commit contains some followups for the terminating rules patch:
 - action list conversion between grpc and C-api
 - statistic counter handles pre-calculation
 - per-rule counters
 - action handling

### DIFF
--- a/modules/acl/api/controlplane.c
+++ b/modules/acl/api/controlplane.c
@@ -91,18 +91,21 @@ acl_module_config_init(
 		uint64_t size;
 		uint64_t *dst;
 	} counters[] = {
-		{"acl_no_match", 2, &config->no_match_counter_id},
-		{"acl_action_allow", 2, &config->action_allow_counter_id},
-		{"acl_action_deny", 2, &config->action_deny_counter_id},
-		{"acl_action_count", 2, &config->action_count_counter_id},
-		{"acl_action_check_state",
-		 2,
-		 &config->action_check_state_counter_id},
+		{"acl_no_match", 1, &config->no_match_counter_id},
+		{"acl_action_allow", 1, &config->action_allow_counter_id},
+		{"acl_action_deny", 1, &config->action_deny_counter_id},
+		{"acl_action_check_pass",
+		 1,
+		 &config->action_check_pass_counter_id},
+		{"acl_action_check_miss",
+		 1,
+		 &config->action_check_miss_counter_id},
 		{"acl_action_create_state",
-		 2,
+		 1,
 		 &config->action_create_state_counter_id},
-		{"acl_action_unknown", 2, &config->action_unknown_counter_id},
-		{"acl_state_miss", 2, &config->state_miss_counter_id},
+		{"acl_action_invalid", 1, &config->action_invalid_counter_id},
+		{"acl_action_non_term", 1, &config->action_non_term_counter_id},
+
 		{"acl_sync_sent", 2, &config->sync_sent_counter_id},
 	};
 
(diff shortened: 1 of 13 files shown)


modules/acl/api/controlplane.h
```diff
@@ -8,6 +8,13 @@
 
 #include "lib/errors/errors.h"
 
+#define ACL_RULE_ACTION_ALLOW 0
```
3Hren: In modules/acl/dataplane/config.h there are also the same defines with different names. Aren't they the same?

BTW why not using enums here? In that case it is possible to detect missing cases in switch statements

modules/acl/controlplane/ffi.go
```diff
@@ -96,6 +102,11 @@ func (m *AclRule) CBuild(pinner *runtime.Pinner) C.struct_acl_rule {
 	cRule := C.struct_acl_rule{}
 
 	CBuildActions(&cRule, m.Actions, pinner)
+
+	cCounter := C.CString(m.Counter)
+	C.strncpy(&cRule.counter[0], cCounter, C.COUNTER_NAME_LEN)
+	C.free(unsafe.Pointer(cCounter))
```
3Hren: Better to defer this.

modules/acl/controlplane/ffi.go
```diff
@@ -96,6 +102,11 @@ func (m *AclRule) CBuild(pinner *runtime.Pinner) C.struct_acl_rule {
 	cRule := C.struct_acl_rule{}
 
 	CBuildActions(&cRule, m.Actions, pinner)
+
+	cCounter := C.CString(m.Counter)
+	C.strncpy(&cRule.counter[0], cCounter, C.COUNTER_NAME_LEN)
```
3Hren: As I understand this, it's not guaranteed that `cRule.counter[0]` will have `\0` in case of cCounter >= COUNTER_NAME_LEN

3Hren (changes requested):

modules/acl/api/controlplane.c
```diff
@@ -91,18 +91,18 @@ acl_module_config_init(
...
 		uint64_t *dst;
 	} counters[] = {
-		{"acl_no_match", 2, &config->no_match_counter_id},
-		{"acl_action_allow", 2, &config->action_allow_counter_id},
-		{"acl_action_deny", 2, &config->action_deny_counter_id},
-		{"acl_action_count", 2, &config->action_count_counter_id},
-		{"acl_action_check_state",
-		 2,
-		 &config->action_check_state_counter_id},
+		{"acl_no_match", 1, &config->no_match_counter_id},
```
3Hren, line 94: `acl_no_match` is now registered with `size=1`, but `acl_handle_packets` still writes both `c[0]` and `c[1]` in the no-match branch (`dataplane.c:356-357`).

modules/acl/controlplane/service.go
```diff
@@ -66,31 +66,26 @@ func (m *ACLService) newHandlerTracker(name string) *handlerMetricTracker {
...
+			result[idx].Kind = ActionDeny
+		case aclpb.ActionKind_ACTION_KIND_COUNT:
+			result[idx].Kind = ActionCount
+		case aclpb.ActionKind_ACTION_KIND_CHECK_STATE:
+			result[idx].Kind = ActionCheckState
+		case aclpb.ActionKind_ACTION_KIND_CREATE_STATE:
+			result[idx].Kind = ActionCreateState
+		case aclpb.ActionKind_ACTION_KIND_LOG:
+			result[idx].Kind = ActionLog
+		}
```
3Hren, line 85: No `default` branch: `ACTION_KIND_SKIPTO` (defined in the proto and accepted in the Rust CLI) and any future/invalid enum value fall through, leaving `result[idx].Kind` at its zero value, which is `ActionAllow`. Previously `terminalAction` mapped the default to `Deny`.

modules/acl/dataplane/dataplane.c
```diff
@@ -221,137 +257,68 @@ acl_handle_packets(
...
-			uint64_t *counters = counter_get_address(
-				target->counter_id,
-				dp_worker->idx,
-				ADDR_OF(&module_ectx->counter_storage)
-			);
-			counters[0] += 1;
-			counters[1] += pkt_len;
-
 			enum sync_packet_direction push_sync_packet = SYNC_NONE;
+			bool allow = false;
```
3Hren, line 273: If an action list contains no terminal action (e.g. `[CreateState]` or `[CheckState, Log]` on miss), the loop falls through with `allow == false`, the packet is dropped at `apply:`, and no module counter is incremented. `CreateState`-only rules thus drop the very packet that installed the state, invisibly to metrics. Suggest either rejecting rules without a terminal in `acl_module_config_update`, or adding an `acl_action_no_terminal` counter and incrementing it here.

3Hren (changes requested):
Three correctness issues need fixing before merge.

3Hren (approved):

PR #809, opened 2026-05-18
refactor(counters): replace nested registries with tag-based lookup

egnees:
## Navigation

This PR replaces the deeply-nested, per-level counter storage registry (device → pipeline → function → chain → module) with a flat, tag-based storage model. Counters are now stored in a flat array of cp_counter_storage items, each annotated with a set of key-value tags, and looked up via tag-predicate matching.

### 1. New data model & core registry — start here
| File | What changed |
|------|-------------|
| [`lib/cp_counter.h`](lib/controlplane/config/cp_counter.h) | Introduces `cp_counter_tag`, `cp_counter_storage`, and the flat `cp_config_counter_storage_registry` (items array + count/capacity). Adds insert and find APIs that work with generic `counter_tag` predicates. Removes dependency on the old nested `registry` type. |
| [`lib/cp_counter.c`](lib/controlplane/config/cp_counter.c) | **Largest change (1248 → ~500 lines).** Implements tagged insert and find. Replaces five nested struct types and their per-level insert/lookup/free callbacks with a flat array with `_insert()` / `_find()` over sorted tags. All old insert and lookup functions now delegate to the generic tag API.

### 2. Public API adjustments
| File | What changed |
|------|-------------|
| [`api/counter.h`](api/counter.h) | Removes `shard` from supported tags |

### 3. Counter query functions rewritten to use tags
| File | What changed |
|------|-------------|
| [`lib/agent.c`](lib/controlplane/agent/agent.c) | **All per-level counter getters** (Implements new tag-based counters API. `yanet_get_module_counters`, `yanet_get_chain_counters`, `yanet_get_function_counters`, `yanet_get_pipeline_counters`, `yanet_get_device_counters`) now build a `counter_tag[]` array and delegate to the new `yanet_get_counters_by_tags()`.

### 4. Dead code removal
| File | What changed |
|------|-------------|
| [`lib/zone.h`](lib/controlplane/config/zone.h) | Removes 5 inline helpers which no longer needed |

### Suggested review order
1. [`lib/controlplane/config/cp_counter.h`](lib/controlplane/config/cp_counter.h) — new types & API surface
2. [`lib/controlplane/config/cp_counter.c`](lib/controlplane/config/cp_counter.c) — core implementation (tag validation, flat insert/find, legacy wrappers)
3. [`api/counter.h`](api/counter.h) — public API change (`const` removal, doc update)
4. [`lib/controlplane/agent/agent.c`](lib/controlplane/agent/agent.c) — consumer side: new `yanet_get_counters_by_tags()` + rewritten per-level getters
5. [`lib/controlplane/config/zone.h`](lib/controlplane/config/zone.h) — deleted helpers


lib/controlplane/agent/agent.c
```diff
@@ -1386,84 +1388,158 @@ yanet_get_module_counters(
...
+cp_counter_storage_copy_tags(const struct cp_counter_storage *storage) {
+	struct counter_tag *tags = malloc(storage->tag_count * sizeof(*tags));
+	if (tags == NULL) {
 		return NULL;
 	}
-	struct counter_storage *counter_storage = cs;
-	struct counter_registry *counter_registry =
-		ADDR_OF(&counter_storage->registry);
+	for (size_t i = 0; i < storage->tag_count; ++i) {
+		tags[i].key = storage->tags[i].key;
```
3Hren, line 1429: Potential UAF

egnees: This issue existed prior to the current PR changes. The `counter_handle` returned to the user contains a reference to the counter storage, which is invalidated during a controlplane update. Because this is a separate issue, it should be addressed in a future PR (likely by locking the controlplane when acquiring the counters list and unlocking it upon freeing).

GeorgyKirichenko (approved):

PR #815, opened 2026-05-20
feat(balancer): optionally pass real weights and states in config

egnees:
This PR adds support for optional `weight` and `enabled` fields in the real configuration. During a config update, these values can now either be inherited from the existing state or overwritten with new values


egnees:
Closing — this targets the pre-FQN balancerpb/config.proto path, which moved to balancerpb/v1/config.proto in #957. The same change (optional weight, new enabled field) will be resubmitted against the current layout.

PR #830, opened 2026-05-22
fix(route): IPv6 range-to-CIDR upper-half decomposition

3Hren:
Fix range-to-CIDR decomposition for the upper IPv6 half, which was emitting malformed prefixes at the all-ones boundary.

Also:
- Configs listing is returned sorted, so tabs and CLI output are deterministic across calls.
- Edit row button matches the save and delete header actions in size.
- Move the route module ahead of decap in the sidebar.


3Hren:
Changed my mind, prefixes is better (for now), will revert soon.

PR #855, opened 2026-05-24
feat: add YANET dashboard as application home page

3Hren:
Adds a new /builtin/dashboard page and makes it the application's landing page (root / redirects there). The page renders the real instance data as an isometric three.js topology where device boxes feed parallel pipeline lanes and function blocks sit on top of those lanes. Hovering a clickable element shows a pointer cursor; clicking opens a side inspector with rates, status and a trend sparkline.

- Camera is locked to the isometric view (no rotation, zoom or pan); pointer-move raycasts only to update cursor state.
- Scene rebuilds only when the structural data (device/pipeline/function identities) changes, never on counter ticks, so the WebGL context is created once per page lifetime. Live rates and trends feed the animation loop through a ref.
- Falls back to a friendly "WebGL unavailable" panel and an error boundary if the browser cannot create a WebGL context.
- Top region: system state strip (uptime / workers / controlplane status, with worker count inferred from the first device's counter response), compact KPI counts strip, and a throughput hero with aggregate bps/pps and a sparkline pinned to the card's bottom.
- Bottom region: dataplane modules grid with deep-link navigation to the corresponding module pages; modules without a page show the non-clickable cursor.
- All three top cards share a subtle ambient horizontal-scan overlay; sharp rectangular corners; warm dark-brown palette consistent with the /builtin/inspect HUD page; light-theme override scoped under .dashboard-page.
- The dashboard is not listed in the sidebar; the AsideHeader logo button navigates to it via React Router (with href fallback for middle-click open in new tab).
- three and @types/three added as dependencies; the page chunk is lazy-loaded.
- The existing /builtin/inspect HUD page remains unchanged.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -12,11 +12,13 @@
                 "@gravity-ui/navigation": "^3",
                 "@gravity-ui/uikit": "^7",
                 "@tanstack/react-virtual": "^3.13.12",
+                "@types/three": "^0.146.0",
                 "diff": "^9.0.0",
                 "js-yaml": "^4.1.0",
                 "react": "^19",
                 "react-dom": "^19",
-                "react-router-dom": "^7.9.6"
+                "react-router-dom": "^7.9.6",
+                "three": "^0.146.0"
             },
             "devDependencies": {
                 "@testing-library/jest-dom": "^6.9.1",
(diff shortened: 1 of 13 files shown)


3Hren:
Closing — opened prematurely, will iterate before re-opening.

PR #865, opened 2026-05-25
Acl convert counter

GeorgyKirichenko:


3Hren (commented):
Honest review — not approving yet. Big picture first:

**This PR bundles six unrelated changes** that should not share a review:

1. `value_table` storage refactor (chunked → per-row arrays) and `(h_dim, v_dim)` parameter swap.
2. net6 lookup cache locality (3-pass split).
3. New worker counters RPC (proto + Go + Rust CLI).
4. `ADDR_OF_NC` macro (unchecked variant) and its rollout in `lpm`/`value_table`.
5. `packet_front` reuse across `worker_loop_round` iterations.
6. ACL: propagate `Rule.Counter` through `convertRules`.

The PR title (`Acl convert counter`) only matches commit 6. The PR body is empty. Please split into focused PRs — at minimum: (a) `value_table` refactor + parameter swap, (b) `ADDR_OF_NC` + lpm/value_table adoption, (c) net6 cache locality, (d) worker counters end-to-end, (e) worker.c packet_front hoisting, (f) ACL counter field. Each is reviewable in isolation; bundled they obscure each other.

Most critical individual concern: the `*(uint64_t *)(0) = …` "assertion" in `value_table_get_ptr` is undefined behavior and may be optimized away under `-O2`. See inline comment.

The `(h_dim, v_dim)` → `(v_dim, h_dim)` parameter swap is invisible to the compiler (both `uint32_t`) and touches every classifier compiler/query. I traced all call sites and they are internally consistent (every caller flipped both arguments), but this is exactly the kind of change that needs a commit message explaining the rationale and a follow-up that gives the dimensions distinguishable types (rows/cols, or a tagged struct).

Detailed comments inline.

common/value.h
```diff
@@ -24,100 +22,86 @@ value_table_free(struct value_table *value_table) {
...
+	struct value_table *value_table, uint32_t v_idx, uint32_t h_idx
 ) {
-	uint32_t **values = ADDR_OF(&value_table->values);
-	uint64_t idx = (v_idx * value_table->h_dim) + h_idx;
-
-	return ADDR_OF(values + idx / VALUE_TABLE_CHUNK_SIZE) +
-	       idx % VALUE_TABLE_CHUNK_SIZE;
+	if (v_idx >= value_table->v_dim || h_idx >= value_table->h_dim) {
+		*(uint64_t *)(0) = v_idx + h_idx;
+	}
```
3Hren: This is undefined behavior — a write through a null pointer at constant address `0`. Under `-O2`/`-O3` the compiler is allowed to prove this branch is unreachable and delete the entire bounds check (because reaching it would mean executing UB). The result on a release build is silently corrupted reads, not a deterministic crash.

Replace with one of:
- `__builtin_trap();` — deterministic SIGILL, zero overhead, and the compiler treats it as `noreturn`.
- `abort();` — if you want a chance to flush logs.
- A real `assert(v_idx < value_table->v_dim && h_idx < value_table->h_dim);` plus build-mode discipline.

Using `(void)(v_idx + h_idx)` or similar to "use" the indices for the trap is fine, but please don't ship a null write as the panic mechanism.

common/value.h
```diff
@@ -24,100 +22,86 @@ value_table_free(struct value_table *value_table) {
...
 }
 
 static inline int
 value_table_init(
 	struct value_table *value_table,
 	struct memory_context *memory_context,
-	uint32_t h_dim,
-	uint32_t v_dim
+	uint32_t v_dim,
+	uint32_t h_dim
```
3Hren: The `(h_dim, v_dim)` → `(v_dim, h_dim)` swap is an API break for every caller (vlan/port/net4/net6/proto/proto_range/device/range_index/lpm/tests). I verified each call site is consistently flipped, so the data is laid out the way the writers wrote it, but the swap is invisible to the type system — both args are `uint32_t`.

Two follow-ups would make this safe to live with:
1. Rename to `rows`/`cols` (or `outer`/`inner`) so call sites read like English instead of `(1, N)`.
2. Spell out in a header comment which dimension is the outer-loop / cache-friendly one now (looks like `v_idx` selects a row pointer, `h_idx` indexes within the row → iterate inner over `h_idx`).

Without this, a future caller who swaps the two args back will pass the build and silently corrupt the table layout.

common/value.h
```diff
@@ -10,8 +10,6 @@
 #include "memory.h"
 #include "remap.h"
 
-#define VALUE_TABLE_CHUNK_SIZE 16384
-
 struct value_table {
```
3Hren, line 15: Storage model changed from one flat chunked array (`VALUE_TABLE_CHUNK_SIZE = 16384`) to a vector of independently-allocated rows. Worth a commit-message note on:

- Why the old chunking was abandoned (was the allocator now able to handle large allocations? was the chunk arithmetic hurting hot-path indexing?).
- The allocator implication: `v_dim` separate `memory_balloc` calls instead of one. For the net6 comb table the dimensions can be in the thousands — does `memory_balloc` cope with that many small allocations without fragmentation? The old code amortized across 16K-element chunks.

If this is a deliberate trade for predictable hot-path access (one pointer indirection, no division/modulo by `VALUE_TABLE_CHUNK_SIZE`), please say so.

common/memory_address.h
```diff
@@ -19,6 +19,13 @@
 							      : NULL));        \
 	})
 
+#define ADDR_OF_NC(OFFSET)                                                     \
```
3Hren: Please document this macro alongside `ADDR_OF`:

- What does `NC` stand for ("no check"? "no NULL check"?).
- The invariant the caller must uphold: `*(OFFSET)` is non-zero, OR the caller is fine with a result of `OFFSET` itself (which is what the math produces when offset_val is 0 — i.e., a pointer to the offset field, not NULL).
- Why it exists: presumably the `?:` in `ADDR_OF` blocks instruction-level parallelism on hot lookup paths (lpm, value_table). Saying that out loud helps future readers avoid "optimizing" `ADDR_OF` itself.

Without documentation, the next reader is going to use whichever name they typed first and the safety distinction will erode.

common/lpm.h
```diff
@@ -48,8 +48,8 @@ struct lpm {
 
 static inline struct lpm_page *
 lpm_page(const struct lpm *lpm, uint32_t page_idx) {
-	struct lpm_page **pages = ADDR_OF(&lpm->pages);
-	struct lpm_page *chunk = ADDR_OF(&pages[page_idx / LPM_CHUNK_SIZE]);
+	struct lpm_page **pages = ADDR_OF_NC(&lpm->pages);
```
3Hren: Swapping `ADDR_OF` → `ADDR_OF_NC` in the hot lookup path is fine *if* the invariant holds: `lpm->pages` is non-NULL after `lpm_init`, every populated chunk pointer is non-NULL, and every populated `value->page` offset on a non-terminal hop is non-NULL. Is there a single source of truth that asserts/maintains all three?

A one-line comment above `lpm_page` (and the analogous one in `lpm_lookup`) stating that invariant would protect against someone later introducing a sparse-chunk allocation that leaves some entries zero.

(Same comment applies to the `ADDR_OF_NC` call inside `lpm_lookup` further down.)

filter/query/device.h
```diff
@@ -13,6 +13,8 @@ FILTER_ATTR_QUERY_FUNC(device)(
 	struct value_table *t = (struct value_table *)data;
 	for (uint32_t idx = 0; idx < count; ++idx) {
 		uint64_t device_id = packets[idx]->module_device_id;
+		if (device_id >= t->h_dim)
```
3Hren: Silently remapping an out-of-range `device_id` to `0` is dangerous: it turns a real bug (control plane and dataplane disagreeing on the device registry) into "this packet got the rules meant for device 0", which can mean "accept" in some ACL setups.

If you genuinely need a guard against the OOB write that `value_table_get_ptr` now traps on (see the value.h null-write comment), prefer one of:

- An explicit unmatched/deny marker: `result[idx] = FILTER_RULE_INVALID; continue;`.
- A drop signal that the surrounding pipeline understands.
- An `assert` / `LOG(ERROR, ...)` so the failure mode is loud, not silent.

Bonus: this guard exists because the trap in `value_table_get_ptr` would otherwise fire — that argues for the trap being expressed as something callers can probe, not a hard `*(uint64_t*)0` write.

filter/query/net6.h
```diff
@@ -24,14 +27,28 @@ FILTER_ATTR_QUERY_FUNC(net6_dst)(
...
 		);
 
-		uint32_t hi = lpm8_lookup(
+		hi_values[idx] = lpm8_lookup(
 			&c->hi, (const uint8_t *)ipv6_hdr->dst_addr
 		);
-		uint32_t lo = lpm8_lookup(
+	}
+
+	for (uint32_t idx = 0; idx < count; ++idx) {
```
3Hren: The `dst_addr` 3-pass split duplicates the entire `packet_to_mbuf` + `rte_pktmbuf_mtod_offset` recomputation in passes 1 and 2 — the compiler will *probably* CSE it, but you're betting against the alias analysis on `mbuf->buf_addr` here.

Cleaner options:

- Cache the eight `dst_addr` bytes in a small stack array on pass 0, then run two lpm loops over that array, then the merge loop. Lets you keep the hot inputs in L1 between the two lpm passes.
- Or split out a `static inline const uint8_t *net6_dst_ptr(const struct packet *)` helper and call it once per packet in pass 1, store the pointer in a `uint8_t *addrs[count]`, reuse in pass 2.

Also `net6_src` is a copy of `net6_dst`. The two should share a single helper parameterized on `src_addr`/`dst_addr` (or on the byte offset).

filter/tests/meson.build
```diff
@@ -173,14 +173,15 @@ combo_net6_port_src = executable('combo_net6_port_src',
...
-test('combo_net4_port_proto_dst', combo_net4_port_proto_dst, suite: ['filter'])
+# blacklist as invalid - proto lookup generates unclassified values
+#combo_net4_port_proto_dst = executable('combo_net4_port_proto_dst',
+#  ['combo_net4_port_proto_dst.c'], 
+#  c_args: yanet_test_c_args,
+#  link_args: yanet_link_args,
+#  dependencies: test_deps,
+#  link_language: 'c'
+#)
+#test('combo_net4_port_proto_dst', combo_net4_port_proto_dst, suite: ['filter'])
```
3Hren: Commenting out a test with `#` lines (no tracking issue, no `skip(): "…"` reason in a `test()` call) effectively deletes it from CI without leaving a paper trail. The reason "blacklist as invalid — proto lookup generates unclassified values" sounds like the test caught a real defect.

Please either:

- File an issue, reference it in a single-line comment, and **delete** the dead lines; or
- Use Meson's `test(..., should_fail: true)` if the failure is the assertion being tested; or
- Fix the underlying "unclassified values" problem in this PR (since this PR is already touching the value-table layer).

Leaving commented-out executable + `test()` declarations rots fast.

dataplane/worker.c
```diff
@@ -314,15 +314,15 @@ worker_loop_round(struct dataplane_worker *worker) {
...
+	struct packet_front schedule_input[device_count];
+	for (uint64_t idx = 0; idx < device_count; ++idx)
+		packet_front_init(schedule_input + idx);
 
-		struct packet_front schedule_input[device_count];
-		for (uint64_t idx = 0; idx < device_count; ++idx)
-			packet_front_init(schedule_input + idx);
+	struct packet_front schedule_output[device_count];
+	for (uint64_t idx = 0; idx < device_count; ++idx)
+		packet_front_init(schedule_output + idx);
```
3Hren, line 322: Hoisting `schedule_input` / `schedule_output` out of the `while (1)` loop changes a per-iteration init into a per-call init. Combined with the change in `packet_front.h` that resets all four src lists during `packet_front_merge`, this works — but the contract is now: "the per-device fronts must be drained back into the central `packet_front` every iteration, or stale state leaks."

Today `packet_front_output` only appends to `front->output`, so the `if (packet_list_first(&schedule_input[idx].output) == NULL) continue;` guard correctly skips devices that received nothing. But that invariant is implicit. Two suggestions:

1. Add a 2-line comment above the hoisted arrays: "these are reused across iterations; `packet_front_merge` is responsible for resetting per-device fronts, and `packet_front_output` only touches `.output`."
2. In debug builds, add `assert(packet_list_first(&schedule_input[idx].pending_input) == NULL && ...)` at the top of each iteration — cheap, and it traps the day someone adds a write to a different list.

Also: `device_count` is read from `cp_config_gen` *once* at function entry, which is fine for current code but please make sure no other path can grow the device registry while a worker is in this loop — the VLA size is fixed for the whole call.

lib/dataplane/module/packet_front.h
```diff
@@ -37,6 +37,11 @@ packet_front_merge(struct packet_front *dst, struct packet_front *src) {
 	packet_list_concat(&dst->drop, &src->drop);
 	packet_list_concat(&dst->pending_input, &src->pending_input);
 	packet_list_concat(&dst->pending_output, &src->pending_output);
+
+	packet_list_init(&src->pending_input);
+	packet_list_init(&src->pending_output);
```
3Hren: Resetting all four src lists inside `merge` is a behavior change that any other caller of `packet_front_merge` will see, not just the worker loop. Worth confirming none of them relied on inspecting `src` post-merge (a quick grep should suffice — I did one and the only consumers I see treat `src` as scratch).

Alternative phrasing that makes the intent explicit at the call site: a separate `packet_front_reset(src)` called by `worker_loop_round` after the merge. That way `merge` stays "move lists from src into dst" and reset is a deliberate caller decision.

controlplane/builtin/counters.go
```diff
@@ -150,6 +150,21 @@ func (m *Counters) Module(
 	return response, nil
 }
 
+// Device returns device counters.
```
3Hren: Doc comment is a copy-paste from `Device`: it should describe `Worker`. Per CLAUDE.md the brief should be a single sentence ending with a period, e.g. `// Worker returns per-worker counters.`

Also the handler ignores `ctx` and `request`. `ctx` is fine to drop quietly (no downstream RPC), but if `WorkerCountersRequest` will never carry fields, document that — otherwise the next reader will assume you forgot to read e.g. an instance/worker filter.

controlplane/ynpb/counters.proto
```diff
@@ -6,6 +6,7 @@ option go_package = "github.com/yanet-platform/yanet2/controlplane/ynpb;ynpb";
 
 service CountersService {
   // Inspect retrieves the current configuration state.
```
3Hren, line 8: Two small things:

- The dangling comment `// Inspect retrieves the current configuration state.` predates this PR and is already misleading (there is no `Inspect` rpc in this service). Since you're touching the service block, please either remove it or replace with per-rpc comments.
- Adding `Worker` as the first rpc reorders the wire-visible method index — fine for gRPC (it's name-based) but worth knowing if anything downstream assumes index ordering.

controlplane/ynpb/counters.proto
```diff
@@ -14,6 +15,8 @@ service CountersService {
   rpc Perf(PerfCountersRequest) returns (PerfCountersResponse) {}
 }
 
+message WorkerCountersRequest {}
```
3Hren: Empty request messages are valid but inflexible — any future per-worker filter (instance id, worker id, name pattern) becomes a breaking change because you'd add a field with a meaning to a message currently treated as `{}`.

If you anticipate filtering by worker, add e.g. `string worker = 1;` (optional, empty = all) now, even if the server currently ignores it. The cost is one proto field; the benefit is wire compatibility.

cli/modules/counters/src/main.rs
```diff
@@ -46,6 +48,10 @@ pub enum ModeCmd {
     Perf(PerfCmd),
 }
 
+#[derive(Debug, Clone, Parser)]
+pub struct WorkerCmd {
+}
```
3Hren: Nits in this block:

- Empty struct with braces: prefer `pub struct WorkerCmd;` over `pub struct WorkerCmd { }`.
- Drop the trailing blank line at line 53; matches the style of the other `Cmd` structs below.
- At the call site: `WorkerCountersRequest { }` → `WorkerCountersRequest {}` (no inner space). `cargo +nightly fmt` should clean both.

cli/modules/counters/src/main.rs
```diff
@@ -137,6 +143,7 @@ async fn run(cmd: Cmd) -> Result<(), Box<dyn Error>> {
     let mut service = CountersService::new(&cmd.connection).await?;
 
     match cmd.mode {
+        ModeCmd::Worker(_cmd) => service.show_worker().await?,
```
3Hren: Since `WorkerCmd` carries no fields, the match arm currently binds `_cmd` and drops it. Once you switch to `pub struct WorkerCmd;` you can write `ModeCmd::Worker(_) => service.show_worker().await?,` and skip the unused-binding suppression.

modules/acl/controlplane/service.go
```diff
@@ -130,6 +130,7 @@ func convertRules(reqRules []*aclpb.Rule) ([]AclRule, error) {
 
 		rule := AclRule{
 			Actions:       convertActions(reqRule.Actions),
+			Counter:       reqRule.Counter,
```
3Hren, line 133: This single-line fix (propagating `Counter` through `convertRules`) is a real bug fix and the only thing in this PR that matches the title `Acl convert counter`. It should be its own PR — easy to backport, easy to revert, doesn't carry the risk of the dataplane refactor.

Also worth a test: a round-trip in `service_test.go` that submits a rule with a `Counter` name and checks it survives `convertRules` — without it, this exact regression can recur silently.

filter/compiler/helper.c
```diff
@@ -60,12 +60,12 @@ merge_and_set_registry_values(
 		return -1;
 	}
 
-	for (uint64_t v_idx = 0; v_idx < value_registry_capacity(registry2);
+	for (uint64_t v_idx = 0; v_idx < value_registry_capacity(registry1);
```
3Hren: The original loop bounds were swapped relative to the registry capacities used in `value_table_init` immediately above — `v_idx` was iterating to `capacity(registry2)` while writes were keyed by `(h_idx=v1<cap1, v_idx=v2<cap2)`. Pre-PR that was either a real OOB bug (if `cap1 != cap2`) or accidentally-correct when the two registries had matching capacities.

This fix is right, but it would be much more reassuring with: (a) a commit message explaining what bug this was masking and how it was found, and (b) a test in `tests/common/value_table.c` that exercises a non-square `(v_dim, h_dim)` init followed by full-table init-to-`FILTER_RULE_INVALID`. Right now the only value-table test is `(1, 10)`, which is symmetric in the failing dimension.

PR #880, opened 2026-05-26
feat(acl): add config hash 

semenovad:
Config hash allows operator to check whether rules differ from the controlplane state without transferring the full ruleset

### DIFF
--- a/modules/acl/cli/src/main.rs
+++ b/modules/acl/cli/src/main.rs
@@ -423,10 +423,12 @@ impl ACLService {
         let request = UpdateConfigRequest {
             name: cmd.config_name.clone(),
             rules: config.rules,
+            rules_hash: cmd.rules_hash,
         };
         log::trace!("UpdateConfigRequest: {request:?}");
         let response = self.client.update_config(request).await?.into_inner();
         log::debug!("UpdateConfigResponse: {response:?}");
+        println!("modified={} rules_hash={}", response.modified, response.rules_hash);
         Ok(())
     }
 
--- a/modules/acl/controlplane/service.go
+++ b/modules/acl/controlplane/service.go
@@ -2,6 +2,8 @@ package acl
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"sync"
 
 	"go.uber.org/zap"
@@ -59,6 +61,7 @@ type ACLService struct {
 
 type aclConfig struct {
 	rules       []*aclpb.Rule
+	rulesHash   string
 	acl         ModuleHandle
 	fwstateName string
 }
@@ -172,6 +175,22 @@ func rulesEqual(a, b []*aclpb.Rule) bool {
 	return true
 }
 
+func computeRulesHash(rules []*aclpb.Rule) string {
+	h := sha256.New()
+	opts := proto.MarshalOptions{Deterministic: true}
+	for _, rule := range rules {
+		data, err := opts.Marshal(rule)
+		if err != nil {
+			continue
+		}
+		// Write 4-byte big-endian length prefix to avoid ambiguity between messages
+		length := uint32(len(data))
+		h.Write([]byte{byte(length >> 24), byte(length >> 16), byte(length >> 8), byte(length)})
+		h.Write(data)
+	}
+	return hex.EncodeToString(h.Sum(nil))
+}
+
 func (m *ACLService) UpdateConfig(
 	ctx context.Context,
 	req *aclpb.UpdateConfigRequest,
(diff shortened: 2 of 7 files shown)


modules/acl/controlplane/aclpb/acl.go
```diff
@@ -13,3 +13,11 @@ func (m *UpdateConfigRequest) AsLogValue() any {
 		return nil
 	})
 }
+
+// AsLogValue implements xgrpc.ProtoLogValue for compact logging.
```
3Hren: What's the problem with default logging format?

modules/acl/controlplane/aclpb/acl.proto
```diff
@@ -20,6 +20,8 @@ service ACLService {
   rpc ListConfigs(ListConfigsRequest) returns (ListConfigsResponse);
   // Returns module-level metrics
   rpc GetMetrics(GetMetricsRequest) returns (GetMetricsResponse);
+  // Returns the SHA-256 hash of the current rules for a given config
+  rpc GetConfigHash(GetConfigHashRequest) returns (GetConfigHashResponse);
```
3Hren: I thought it will be better to provide hash as an optional argument to UpdateConfigRequest. UpdateConfigResponse could return calculated hash to be sent later.
No hash in UpdateConfigRequest means always compile. Existing hash means check and then either compile or return "not modified" under lock.

Thus, we don't need separate method for obtaining hashes nor exposing hash algo.

3Hren (changes requested):

3Hren:
If you want to avoid transferring ruleset, for example, for just checking it's time to upload, you could change mechanics of `UploadRulesetRequest` to include hash and NOT include rules.

modules/acl/controlplane/service.go
```diff
@@ -181,8 +200,22 @@ func (m *ACLService) UpdateConfig(
...
+	if clientHash := req.GetRulesHash(); clientHash != "" {
+		if existing, ok := m.configs[name]; ok && existing.rulesHash == clientHash {
+			return &aclpb.UpdateConfigResponse{
+				Modified:  false,
+				RulesHash: existing.rulesHash,
+			}, nil
+		}
+	}
+
+	newHash := computeRulesHash(req.Rules)
```
3Hren, line 227: This makes a hash-only check unsafe when the client hash does not match. `rules` has no presence bit, so a caller that sends only `name` + `rules_hash` to avoid transferring the ruleset will fall through here with an empty `req.Rules`, compute the empty-rules hash, and then publish an empty ACL config below. In the hash-only mismatch case this should return a "different/upload needed" response or require rules to be present before compiling/publishing.

modules/acl/cli/src/main.rs
```diff
@@ -33,6 +33,7 @@ pub mod aclpb {
 #[derive(Serialize)]
 struct SerializableShowConfigResponse {
     name: String,
+    rules_hash: String,
```
3Hren: nit: move rules_hash after rules

modules/acl/controlplane/service.go
```diff
@@ -59,6 +61,7 @@ type ACLService struct {
 
 type aclConfig struct {
 	rules       []*aclpb.Rule
+	rulesHash   string
```
3Hren, line 64: This new field also has to be preserved by the FWState relink paths. `ACLAdapter.RelinkConfigs` and `ACLAdapter.LinkConfigs` rebuild `aclConfig` in `acl_adapter.go`, but only copy `rules`, `acl`, and `fwstateName`. After a link/relink, `rulesHash` becomes empty, so `ShowConfig` reports an empty hash and a hash-only check with the previous correct hash is treated as a mismatch. Please copy `oldConfig.rulesHash` in both reconstructed configs.

modules/acl/controlplane/service.go
```diff
@@ -181,8 +200,34 @@ func (m *ACLService) UpdateConfig(
...
+		if len(req.Rules) == 0 {
+			return nil, status.Error(codes.InvalidArgument,
+				"rules_hash provided without rules on hash mismatch: "+
+					"omit rules_hash to force-publish an empty ruleset")
+		}
+	}
+
+	newHash := computeRulesHash(req.Rules)
+
+	if req.RulesHash == nil {
```
3Hren: This still makes no-hash requests behave like conditional updates. The review thread above defined `rules_hash` absence as “always compile”, and the new error text says to omit `rules_hash` to force-publish an empty ruleset. With an existing empty ACL, omitting `rules_hash` and sending empty `rules` reaches this branch and returns `Modified=false` instead of republishing.

3Hren (commented):
Left the two review notes inline.

3Hren:
@codex review

modules/acl/controlplane/service.go
```diff
@@ -181,10 +200,27 @@ func (m *ACLService) UpdateConfig(
...
+	// If the server hash differs rules are compiled and published
+	// If rules_hash without rules is rejected when the hashes differ
+	// (omit rules_hash to force-publish an empty ACL)
+
+	if clientHash := req.GetRulesHash(); clientHash != "" {
+		if existing, ok := m.configs[name]; ok && existing.rulesHash == clientHash {
+			return &aclpb.UpdateConfigResponse{
+				Modified:  false,
+				RulesHash: existing.rulesHash,
+			}, nil
```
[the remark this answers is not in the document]
semenovad: rules_hash is a token previously issued by the server via UpdateConfig or ShowConfig.  When the caller echoes it back, compare it to the stored hash before doing any work: a match means the installed config is already up to date and compilation can be skipped. Callers that want to force an update must omit rules_hash

PR #883, opened 2026-05-26
Support for fragmented IP packets

GeorgyKirichenko:


filter/query/ipfrag.h
```diff
@@ -0,0 +1,21 @@
...
+
+static inline void
+FILTER_ATTR_QUERY_FUNC(ip_frag)(
+	void *data, struct packet **packets, uint32_t *result, uint32_t count
+) {
+	struct value_table *t = (struct value_table *)data;
+
+	for (uint32_t idx = 0; idx < count; ++idx) {
+		uint32_t id =
+			packets[idx]->flags & (1 << PACKET_FLAG_FRAGMENTED) ? 0
```
3Hren: This lookup id is reversed relative to the compiler. `filter/compiler/ipfrag.c` stores `FILTER_IP_FRAG_NONE` in value-table slot 0 and `FILTER_IP_FRAG_FRAG` in slot 1, but query maps fragmented packets to slot 0 and non-fragmented packets to slot 1. That makes `fragment: None` and `fragment: Frag` match the opposite traffic.

bindings/go/filter/types.go
```diff
@@ -158,3 +158,12 @@ func NewProtoRange(proto uint8, subtype Subtype) ProtoRange {
 func NewProto(proto uint8, subtype uint8) uint16 {
 	return uint16(proto)<<8 | uint16(subtype)
 }
+
+type Fragment uint32
+
+// Represent Fragmented attribute
+const (
+	FragmentNone Fragment = iota
```
3Hren: This enum is written directly into `struct_acl_rule.fragment` in `cacl.AclRule.cBuild`, but its numeric values do not match the C ABI in `filter/rule.h` (`ANY=0`, `NONE=1`, `FRAG=2`). As a result `None` is sent as `Any`, `Frag` as `None`, and default `Any` as `Frag`, so fragment ACL rules are compiled with the wrong match semantics. Please mirror the C constants here or map explicitly before crossing CGO.

common/filterpb/convert.go
```diff
@@ -241,3 +241,18 @@ func ToVlanRanges(pb []*VlanRange) (filter.VlanRanges, error) {
...
+		return filter.FragmentAny, nil
+	}
+	switch pb.Kind {
+	case FragmentKind_None:
+		return filter.FragmentNone, nil
+	case FragmentKind_Frag:
+		return filter.FragmentFrag, nil
+	}
+
+	return filter.FragmentAny, nil
```
3Hren: Unknown `FragmentKind` values should be rejected instead of silently becoming `Any`. Otherwise a malformed or newer client value broadens an ACL rule and can allow both fragmented and non-fragmented packets. Please return an `InvalidArgument` error here, similar to the strict action-kind conversion.

PR #906, opened 2026-05-28
fix: add NAT64 functional dataplane coverage

3Hren:
- Added service-driven NAT64 functional coverage for translation, drop, extension-header, fragmentation, and ICMP error paths.
- Fixed NAT64 dataplane handling for IPv6 extension headers, non-first fragments, translated IPv4 length, and ICMP Packet Too Big MTU fields.
- Removed the legacy direct-CGO Go dataplane test after moving its useful ICMP echo assertions into functional coverage.
- Kept C dataplane tests as low-level malformed, bounds, and fuzzer regression coverage.
- Validated with Go NAT64 tests, NAT64 Meson compile targets, NAT64 C unittest, formatting checks, and diff whitespace checks.


modules/nat64/dataplane/nat64dp.c
```diff
@@ -1728,13 +1726,15 @@ nat64_handle_v6(
 	if (is_fragmented) {
 		uint16_t total_len = rte_be_to_cpu_16(ipv6_header->payload_len);
 		uint16_t frag_size = total_len - ext_hdrs_len;
+		uint16_t frag_offset_bytes = frag_offset * 8;
```
3Hren, line 1729: `RTE_IPV6_GET_FO()` returns the IPv6 fragment offset in 8-byte units, not bytes. Checking `% 8` after extraction would reject valid offsets such as `2` (16 bytes); convert to bytes only for validation and keep the 8-byte-unit value for the IPv4 fragment field.

3Hren (commented):
Added a clarification about IPv6 fragment offset units.

moonug (approved):
lgtm

PR #916, opened 2026-05-29
fix(fwstate): accumulate flags and packet counters on in-layer updates

sakateka:
fwmap_copy_value_fwstate() used to overwrite the stored fw_state_value
when the key already existed in the active layer, preserving only
created_at. Successive sync frames for the same flow therefore lost
accumulated TCP flags and reset packets_forward/packets_backward.

Make the in-layer update path merge with the previous value: OR flags,
sum per-direction counters, take updated_at/external from the new frame,
keep created_at.

Also rename the two value callbacks for clarity:
  copy_value  -> update_value  (in-layer; old lives in *dst)
  merge_value -> promote_value (cross-layer; old lives in stale layer)

Add Go regression tests covering both the in-layer accumulation and the
cross-layer promotion behavior.

### DIFF
--- a/lib/fwstate/fwmap.h
+++ b/lib/fwstate/fwmap.h
@@ -35,12 +35,12 @@ typedef enum {
 	FWMAP_RAND_DEFAULT,
 	FWMAP_RAND_SECURE,
 	FWMAP_COPY_KEY_DEFAULT,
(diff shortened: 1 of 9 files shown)


sakateka (commented):
LGTM

3Hren (approved):

PR #1047, opened 2026-06-03
refactor(dataplane): reuse per-device packet fronts

egnees:
- Part of a larger refactoring aimed at reducing variable-length array usage in the dataplane processing path.
- Move the per-device input and output scheduling fronts out of the per-round stack arrays into persistent device context state, reusing them across pipeline rounds instead of reinitializing them every iteration.
- Shard the per-device scheduling fronts per worker so that concurrently running workers each own an isolated front, avoiding data races and false sharing on the hot path.
- Resolve device contexts once per round rather than on every iteration.
- Packet processing results are unchanged.


lib/dataplane/pipeline/econtext.h
```diff
@@ -107,6 +108,8 @@ struct device_ectx {
 	struct counter_storage *counter_storage;
 	struct device_entry_ectx *input_pipelines;
 	struct device_entry_ectx *output_pipelines;
+	struct packet_front pending_input;
+	struct packet_front pending_output;
```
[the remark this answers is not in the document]
egnees: Fixed.

lib/dataplane/packet/packet.h
```diff
@@ -95,6 +95,12 @@ packet_list_concat(struct packet_list *dst, struct packet_list *src) {
 	dst->count += src->count;
 }
 
+static inline void
+packet_list_move(struct packet_list *dst, struct packet_list *src) {
```
GeorgyKirichenko: The function name is confusing, you are not really moving but concatenating the values. Should we update the `concat` function instead?

lib/dataplane/module/packet_front.h
```diff
@@ -39,6 +42,14 @@ packet_front_merge(struct packet_front *dst, struct packet_front *src) {
 	packet_list_concat(&dst->pending_output, &src->pending_output);
 }
 
+static inline void
+packet_front_move(struct packet_front *dst, struct packet_front *src) {
```
GeorgyKirichenko: Why not to update merge routine?

lib/dataplane/pipeline/pipeline.c
```diff
@@ -286,20 +287,26 @@ device_ectx_process_input(
 	struct device_ectx *device_ectx,
 	struct packet_front *packet_front
 ) {
+	struct packet_front *pending =
```
GeorgyKirichenko: Why do you wish to change the routine parameter semantic?
Before the PR `process_input` handled the packet front passed via function argument but now it works with an internal structure.

GeorgyKirichenko (changes requested):

lib/dataplane/pipeline/pipeline.c
egnees: Fixed.

lib/dataplane/module/packet_front.h
egnees: After last changes, it is not required.

lib/dataplane/packet/packet.h
egnees: After the last changes, it is not required.

PR #1049, opened 2026-06-03
refactor(web): share command-palette open-state hook

3Hren:
Extracts the command-palette open state and its keyboard wiring (Cmd/Ctrl+K toggles, Escape closes) into a shared hook, replacing the identical boilerplate previously duplicated in the Forward, operators Route, and Neighbours pages. No behaviour change.


3Hren:
Superseded by #1045, which already extracted the command-palette keyboard wiring into a shared `usePaletteShortcut` hook on main.

PR #1059, opened 2026-06-03
feat(web): add command palette to the FWState page

3Hren:
- Add the command palette (⌘K / Ctrl+K) to the FWState page, matching the route operator and forward pages.
- Surface add/save/delete config, switch config, sub-tab navigation, state address-family and include-expired filters, and open the ACL module as palette actions.


3Hren:
Addressed in the latest revision: "Open ACL module" now routes through a guarded `handleOpenAclModule` that prompts for confirmation when there are unsaved changes. The pre-existing Links-tab button was switched to the same guarded handler.

PR #1067, opened 2026-06-04
feat(gateway): distinguish in-process and out-of-process backends

3Hren:
Built-in and in-process gateway services never heartbeat, so their last-seen timestamp is frozen at gateway startup and they appear perpetually stale in `gateway list`. Record how each backend is hosted and surface it.

- Add a `BackendKind` to the backend registry, set at registration: built-in and in-process services (shared loopback) register as in-process, while module backends from the Register RPC register as out-of-process. The heartbeat renew path leaves the kind untouched.
- Extend the `Gateway.ListServices` proto and Go service with a `BackendKind kind` field (additive, no renames).
- Show a Kind column in `yanet-cli gateway list` and render the heartbeat age only for out-of-process backends; in-process backends show a dash, since their liveness equals the gateway's. JSON output carries the kind as a self-describing string.

### DIFF
--- a/cli/modules/gateway/src/main.rs
+++ b/cli/modules/gateway/src/main.rs
@@ -16,7 +16,7 @@ use ync::{
     errors::Error,
     output::{self, CommonFormat},
 };
-use ynpb::pb::{ListServicesRequest, RegisteredBackend, gateway_client::GatewayClient};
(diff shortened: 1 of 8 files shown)


3Hren:
Superseded by the reworked three-kind classification (built-in / in-process / external); the in/out-of-process split in this PR was inaccurate because director-launched module services run their own listener inside the gateway process.

PR #1070, opened 2026-06-04
feat(gateway): classify registered backends by kind

3Hren:
Built-in and in-process gateway services never heartbeat, so their last-seen timestamp is frozen at registration and they appear perpetually stale in `gateway list`. The gateway also could not tell a service launched in its own process apart from one registered by a separate process. Record how each backend is hosted and surface it.

- Add a `BackendKind` to the registry with three values: built-in (compiled into the gateway, shares its gRPC server), in-process (launched in the gateway process by the director, registers once over its own listener), and external (registered from a separate process and refreshed on a heartbeat interval).
- Let registrants declare whether they run inside the gateway process via a new `RegisterRequest.in_process` flag. The director's in-process service runner sets it; separate-process registrants such as operators leave it false and become external. The gateway marks its own services built-in.
- Extend `Gateway.ListServices` with the resolved kind.
- Show a Kind column in `yanet-cli gateway list` and report a heartbeat age only for external backends; built-in and in-process backends show a dash, since they register once and their liveness equals the gateway's. JSON output carries the kind as a self-describing string.

### DIFF
--- a/cli/modules/gateway/src/main.rs
+++ b/cli/modules/gateway/src/main.rs
@@ -16,7 +16,7 @@ use ync::{
     errors::Error,
     output::{self, CommonFormat},
 };
-use ynpb::pb::{ListServicesRequest, RegisteredBackend, gateway_client::GatewayClient};
+use ynpb::pb::{BackendKind, ListServicesRequest, RegisteredBackend, gateway_client::GatewayClient};
 
 const GATEWAY_SERVICE: &str = "ynpb.Gateway";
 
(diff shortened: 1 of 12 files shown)


3Hren:
Addressed the renewal finding: the resolved `BackendKind` is now computed before the renew fast path and threaded through both `Renew` and the same-endpoint case of `registerBackend`, so a same-endpoint re-registration with a changed hosting mode updates the stored kind (and clears the stale heartbeat age) in both directions. Added `TestBackendRegistry_RenewUpdatesKind` covering external↔in-process changes.

3Hren:
Addressed the backward-compat finding: `last_seen_cell` now treats `Unspecified` like `External` and shows `format_age(ts)`, so a newer CLI talking to an older gateway that omits the `kind` field still surfaces heartbeat ages. Only the known no-heartbeat kinds (`built-in`, `in-process`) render a dash. Updated the unit test accordingly.

PR #1091, opened 2026-06-04
feat(dataplane): pin instance storage to its NUMA node

egnees:
Allocate each dataplane instance's storage on its configured NUMA node by faulting in its pages under a bind memory policy. Allocation is strict: if the node cannot back the instance's storage, initialization fails.

### DIFF
--- a/dataplane/dataplane.c
+++ b/dataplane/dataplane.c
@@ -1,9 +1,7 @@
 #include "dataplane.h"
 
 #include "config.h"
-#include "logging/log.h"
-
-#include <stdint.h>
+#include "numa.h"
 
 #include <dlfcn.h>
 #include <pthread.h>
@@ -14,7 +12,6 @@
 #include "dpdk.h"
 
 #include "common/data_pipe.h"
-#include "common/exp_array.h"
 #include "common/strutils.h"
 
 #include "common/hugepages.h"
@@ -33,12 +30,12 @@
 #include "lib/dataplane/config/topology.h"
 #include "lib/dataplane/packet/data.h"
 #include "lib/dataplane/packet/packet.h"
-#include "lib/errors/errors.h"
+#include "lib/logging/log.h"
 
 #include <unistd.h>
 
-#include "sys/mman.h"
 #include <fcntl.h>
+#include <sys/mman.h>
 
 static int
 dataplane_worker_connect(
@@ -293,7 +290,15 @@ dataplane_init(
 		close(mem_fd);
 		return -1;
 	}
+
+	long page_size = file_page_size(mem_fd);
 	close(mem_fd);
+	if (page_size <= 0) {
+		LOG(ERROR, "failed to get storage page size");
+		return -1;
+	}
+
+	assert((uintptr_t)storage % page_size == 0);
 
 	off_t instance_offset = 0;
 	for (uint32_t instance_idx = 0;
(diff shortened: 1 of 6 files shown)


dataplane/dataplane.c
```diff
@@ -211,6 +213,70 @@ dataplane_create_devices(
...
+			err,
+			"numa index %u exceeds maximum %zu",
+			numa_idx,
+			sizeof(uint64_t) * 8 - 1
+		);
+		return -1;
+	}
+	uint64_t numa_mask = 1ull << numa_idx;
+
+	const size_t page_size = sysconf(_SC_PAGESIZE);
```
GeorgyKirichenko: I think this is not fully correct because we may use huge pages for shared memory.

dataplane/dataplane.c
```diff
@@ -211,6 +213,70 @@ dataplane_create_devices(
...
+	if ((uintptr_t)start % page_size != 0) {
+		yanet_error_add(err, "start address is not page-aligned");
+		return -1;
+	}
+	if (size % page_size != 0) {
+		yanet_error_add(err, "size is not a multiple of page size");
+		return -1;
+	}
+
+	if (set_mempolicy(MPOL_BIND, &numa_mask, sizeof(numa_mask) * 8) != 0) {
```
GeorgyKirichenko: Should we use MPOL_F_STATIC_NODES with the MPOL_BIND?

dataplane/dataplane.c
```diff
@@ -211,6 +213,70 @@ dataplane_create_devices(
...
+		);
+		return -1;
+	}
+
+	// TODO: intercept SIGBUS
+	for (size_t i = 0; i < size; i += page_size) {
+		((char *)start)[i] = 0;
+	}
+
+	if (set_mempolicy(MPOL_DEFAULT, NULL, 0) != 0) {
```
GeorgyKirichenko: Please use `get_mempolicy` before updating and restore original value instead of a default one

dataplane/dataplane.c
```diff
@@ -305,6 +371,23 @@ dataplane_init(
 			config->instances + instance_idx;
 
 		LOG(INFO, "initialize storage for instance %u", instance_idx);
+		yanet_error *err = NULL;
+		if (allocate_instance_on_numa(
```
GeorgyKirichenko: Should we call this inside `dp_storage_init`?

GeorgyKirichenko (changes requested):

dataplane/dataplane.c
egnees: Using page_size here is an optimization — we could memset the entire region instead. Stepping by the actual page size isn't necessary.

dataplane/dataplane.c
egnees: I don't think so, since `dp_storage_init` is responsible only for populating the dp and cp configs. Moreover, it's shared with the unit tests library, where the arena is aligned to 64 bytes rather than the 4096-byte alignment required by `allocate_instance_on_numa`.

dataplane/dataplane.c
egnees: I think the alignment of dataplane_instance storages should be addressed in separate PRs. For now, it's required to pass cp_memory and dp_memory values that are divisible by the (huge)page size.

GeorgyKirichenko: sysconf(_SC_PAGESIZE) is equal to 4kb where actual page size may be 4kb, 2mb or either 1gb (on x86_64)
also `memset` is Ok here - this is just an initialization stage

dataplane/dataplane.c
GeorgyKirichenko: I do not see how 4096 alignment is required for `set_mempolicy`
Also if you want to have storage configuration initialization separated - why not to combine storage space allocation and populating inside one routine

dataplane/dataplane.c
egnees: Fixed.

dataplane/dataplane.c
egnees: Yes, you are right. Fixed.

dataplane/dataplane.c
egnees: Added utility for getting page size of the mmap-ed storage. Use true page size now.

dataplane/dataplane.c
egnees: We need to check the alignment because misaligned instance storages can share the same (huge)page, making it impossible to pin storages to different NUMA nodes

egnees: Currently, `dp_storage_init` and its containing library know nothing about the hardware and don't link against `libnuma`, which is a good thing. The dataplane library, on the other hand, already has extensive hardware knowledge and links with `libnuma`. So I think it's quite convenient to keep `dp_storage_init` separate from NUMA-aware allocation, which aligns with the already established pattern.

GeorgyKirichenko (approved):

PR #1138, opened 2026-06-06
perf(dataplane): make packet list byte accounting O(1)

3Hren:
Maintain a running byte total on the packet list, updated by its mutators, so the per-list byte sum is read in constant time instead of walking the list on every call. The sum runs about ten times per pipeline round, so this removes an O(n) cost from the hot path while keeping the reported byte counts identical.

The header stays free of DPDK includes (the per-packet length is read through a thin accessor defined in the implementation file), preserving the existing boundary for the Go bindings and public API.

Add a correctness test asserting the maintained total matches an independent brute-force sum across a mix of add, concat, and pop, validated under AddressSanitizer.

### DIFF
--- a/lib/dataplane/packet/packet.c
+++ b/lib/dataplane/packet/packet.c
@@ -555,13 +555,14 @@ logtrace_rte_mbuf(struct rte_mbuf *mbuf) {
 #endif // ENABLE_TRACE_LOG
 }
 
+uint16_t
+packet_accounted_len(const struct packet *packet) {
+	return packet_data_len((struct packet *)packet);
+}
+
 uint64_t
 packet_list_bytes_sum(struct packet_list *list) {
-	uint64_t bytes = 0;
-	for (struct packet *pkt = list->first; pkt != NULL; pkt = pkt->next) {
-		bytes += packet_data_len(pkt);
-	}
-	return bytes;
+	return list->bytes;
 }
 
 void
--- a/lib/dataplane/packet/packet.h
+++ b/lib/dataplane/packet/packet.h
@@ -56,13 +56,19 @@ struct packet_list {
 	struct packet *first;
 	struct packet **last;
 	uint64_t count;
+	uint64_t bytes;
 };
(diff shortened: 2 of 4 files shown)


3Hren:
Benchmarked this change with the new Go `testing.B` + `benchstat` dataplane harness (now on `main`). Summary: **the O(1) accounting is throughput-neutral, not a measurable win**, so I'm closing this rather than merging it.

### Measurements (both builds `debugoptimized -O2`, n=10)

- `BenchmarkPipelineRound` (empty pipeline, list-operation dominated): the apparent regression (geomean +4.9% sec/op) is **host noise + a loop-alignment artifact**, not a real per-packet cost. Proof: a reproducible −5% *improvement* at `batch=8` shows up in both candidate runs — impossible for a genuine monotonic per-packet tax. A baseline-vs-itself control on this shared 40-core host swings ±3.5% per batch with low p-values, so anything within ~±5% here is noise.
- `BenchmarkACL{Allow,Deny}` (representative module work): **NEUTRAL** (geomean +0.7% sec/op, within noise).

### Why not pursue the inline variant

The running counter calls `packet_accounted_len()` on every `packet_list_add`/`pop`. That call is **non-inlinable by design**: live `data_len` lives in `rte_mbuf`, and `packet.h` is the cgo-boundary DPDK-free header that cannot include `rte_mbuf`. The only inline alternative — caching `data_len` in `struct packet` at parse time — would change the byte counters from **current length** to **ingress length**, mis-counting `output`/`drop` bytes after decap/nat64 shrink the packet (the counters in `pipeline.c` are taken post-module-processing). So there is no inline *and* correctness-preserving variant.

### Conclusion

Packet lists are short in practice, so O(n)-vs-O(1) for the byte sum doesn't matter, while the change adds a non-inline call on the hot `add`/`pop` path and grows `struct packet_list` by 8 bytes. Net: no measurable benefit, small structural cost. Keeping the simple O(n) sum.

PR #1139, opened 2026-06-06
chore(dataplane): remove stat.log debug logging

3Hren:
Remove the dead `stat.log` debug-logging path from the dataplane.

- Delete the `stat_thread` worker that opened `stat.log` and wrote per-second device and extended statistics deltas.
- Drop the `pthread_create` launch in `dataplane_start` and the now-unused `pthread` include.

### DIFF
--- a/dataplane/dataplane.c
+++ b/dataplane/dataplane.c
@@ -6,7 +6,6 @@
 #include <stdint.h>
 
 #include <dlfcn.h>
-#include <pthread.h>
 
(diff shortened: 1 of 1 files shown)


3Hren:
Superseded by ec282e84 (feat(dataplane): per-port statistics counters): the stat.log debug path no longer exists on main — stat_thread was reworked to collect device xstats into counter storage instead of writing a file.

PR #1142, opened 2026-06-06
feat(acl): add rule-scale and diverse-flow benchmarks

3Hren:
Add two ACL dataplane benchmarks on top of the Go-driven harness so the hot path can be measured under more realistic conditions than the existing single-flow benches.

- Add \`BenchmarkACLAllowRuleScale\`, sweeping the compiled rule count from 1 to 10000 and driving a batch whose flows spread across the installed rules, exercising the filter lookup across the full rule set rather than one hot entry.
- Add \`BenchmarkACLAllowDiverseFlow\`, sweeping the batch size while sending distinct per-packet flow tuples, isolating the per-packet header-resolution cost.
- Use a bench-local harness with enlarged memory regions for the 10000-rule case, leaving the shared functional-test constants untouched.

### DIFF
--- a/modules/acl/tests/functional/acl_bench_test.go
+++ b/modules/acl/tests/functional/acl_bench_test.go
@@ -6,13 +6,17 @@ import (
 	"net/netip"
 	"testing"
 
+	"github.com/c2h5oh/datasize"
 	"github.com/gopacket/gopacket"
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #1185, opened 2026-06-11
refactor(cli/acl): migrate to the output framework

3Hren:
- Migrates the ACL module CLI to the shared output framework: a global format flag (human by default, JSON for scripting), structured errors with endpoint and service context, and kind-specific exit codes.
- Removes the local format flag from the metrics subcommand — the global flag now drives all output, with the existing grouped tables as the human rendering.
- Renders a shown config as YAML in human mode, so the output round-trips directly into the update command, while JSON mode keeps the full response.
- Reports delete and update results through structured success messages instead of silence.

Note: the default output for list, show, and metrics changes from raw JSON to human-readable, matching every CLI already migrated to the framework. Pass the format flag explicitly to keep JSON in scripts. Functional tests are unaffected — they only invoke update and do not parse output.

### DIFF
--- a/modules/acl/cli/src/args.rs
+++ b/modules/acl/cli/src/args.rs
@@ -41,14 +41,6 @@ pub struct ShowCmd {
     pub config_name: String,
 }
 
-#[derive(Debug, Clone, clap::ValueEnum, PartialEq, Default)]
-pub enum OutputFormat {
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1191, opened 2026-06-11
refactor(web): share delete and discard confirmation dialogs between card headers

3Hren:
- Extracts the duplicated delete and discard confirmation dialog pair from the functions and pipelines card headers into a shared component.
- The shared component reproduces every dialog title, message and action ordering exactly, so there is no visual or behavioral change.


3Hren:
@codex review

PR #1218, opened 2026-06-12
fix(route-operator): flag every member of the per-source best group as best

3Hren:
- Show and lookup responses now mark a route as best when it belongs to its source's best-cost group, matching the set that is actually installed into the FIB, so all ECMP members are shown as selected instead of only the first one.
- The web lookup drawer renders the server flag instead of assuming the first row is best; the routes table and CLI already used the flag and pick the change up automatically.
- Covered by RIB mask tests and service-level tests for static ECMP and unequal-cost BIRD routes.


3Hren:
@codex review

PR #1226, opened 2026-06-13
feat(common): link memory contexts into a parent-child tree

3Hren:
Give the shared-memory allocator context a parent/child/sibling tree so memory consumption can later be attributed to the subsystem that allocated it. This is the foundation for surfacing per-subsystem memory in inspect.

- Add parent, first_child and next_sibling offset links to the memory context, maintained on initialisation and teardown.
- Make context finalisation order-independent by detaching any remaining children, closing a latent use-after-free when a parent is finalised before its children.
- Cover the link, unlink and ordering invariants with unit tests.

### DIFF
--- a/common/memory.h
+++ b/common/memory.h
@@ -8,7 +8,6 @@
 #include "memory_block.h"
 #include "strutils.h"
 
-// TODO: link parent and child context
 struct memory_context {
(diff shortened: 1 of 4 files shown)


3Hren:
@codex review

PR #1228, opened 2026-06-13
refactor(common): remove unused recursive spinlock

3Hren:
The recursive spinlock variant in the shared spinlock header has no consumers anywhere in the tree. Removing it deletes dead code and drops the header's dependency on the unistd and syscall system headers, which keeps it safe to include from low-level, widely-included headers.


3Hren:
@codex review

PR #1229, opened 2026-06-13
refactor(filter): rename open/close point enum constants

3Hren:
The point classifier's enum used the constant names open and close, which shadow the libc open and close functions. That breaks compilation the moment any header on the include path pulls in unistd. Rename the constants with a type prefix to remove the latent clash. The unrelated classifier struct field of the same name is left untouched.


3Hren:
@codex review

PR #1235, opened 2026-06-15
test(xnetip): add property-based tests for IP range and CIDR helpers

3Hren:
- Add property-based tests (using pgregory.net/rapid) for the xnetip range and CIDR helpers.
- Cover the round-trip between a prefix and its address range, and the structural invariants of the range-to-CIDR decomposition: masked output, matching address family, contiguous coverage, and exact endpoints.
- Introduce pgregory.net/rapid as a test-only dependency.


3Hren:
@codex review

PR #1264, opened 2026-06-19
feat(web): uniform keyboard control for tables

3Hren:
Unifies keyboard control across every editable and detail table in the web UI so they all behave the same way.

- Arrow keys move the highlighted row; Enter or a click opens that row's editor; Escape closes it.
- Tab moves focus from the table into the editor, and Ctrl/Cmd+Enter saves and returns to the table.
- The per-row hover edit button is removed and whole rows are clickable; since in-table text selection is dropped, every column's data is now shown inside the editor (including read-only route attributes and live forward counters).
- Applying an editor without making changes no longer marks a configuration as having unsaved changes.


3Hren:
@codex review

PR #1265, opened 2026-06-19
fix(web): keep nested controls from activating table rows

3Hren:
Follow-up to #1264, which landed a Codex-flagged issue: clicking a counter play/pause button inside an ACL rule row also opened the rule drawer, because the shared table row handler received the bubbled click.

- Row activation now ignores clicks that originate from nested interactive controls (buttons, links, menu items), so those controls act independently.
- The guard mirrors the table's existing keyboard activation behaviour, keeping mouse and keyboard consistent.
- Adds a unit test covering the guard.


3Hren:
@codex review

PR #1268, opened 2026-06-20
fix(filter): prevent double-free and NULL-deref in net4 classifier init

3Hren:
The net4 attribute compiler published its prefix-tree pointer before validating the allocation and left it published after a build failure.

- On allocator exhaustion the prefix-tree pointer was dereferenced without a NULL check, crashing the control plane while compiling a filter.
- On a build failure the value collector already released the prefix tree, but the pointer stayed published, so the filter teardown path released it a second time.
- Reset the published pointer and free the wrapper struct on failure, matching the device, vlan and ip-fragment compilers, and validate the allocation before use.

### DIFF
--- a/lib/filter/compiler/net4.c
+++ b/lib/filter/compiler/net4.c
@@ -128,7 +128,7 @@ collect_net4_values(
 		}
 	}
 	if (lpm_init(lpm, memory_context)) {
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #1273, opened 2026-06-20
chore(raii-sweep): refine lifecycle convention and add structural scanner

3Hren:
- Restate the RAII convention as four strictly orthogonal primitives: new allocates the struct, init fills fields via a single error label that calls fini, fini releases field memory idempotently, and free deallocates the struct.
- Split the sweep into a naming axis and a structural-correctness axis, each driven by its own read-only scanner.
- Add a structural scanner that flags error-path, NULL-safety, new/free orthogonality, and fini idempotency smells the naming scanner cannot see.
- Expand the triage rules with the validated violation classes, cross-cutting lifecycle invariants, and false-positive criteria.

### DIFF
--- a/.claude/skills/raii-sweep/SKILL.md
+++ b/.claude/skills/raii-sweep/SKILL.md
@@ -1,21 +1,25 @@
 ---
 name: raii-sweep
 description: >-
-  Weekly maintenance sweep that restores RAII symmetry in the YANET2 C code:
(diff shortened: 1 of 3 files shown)


3Hren:
Both review findings are addressed in #1275: the scanner now accepts `init` as a lifecycle verb (so init error-path and idempotency checks are live, and the filter-compiler init seeds are visible) and matches `static` two-line definitions. The scanned-definition count rose from 99 to 150.

PR #1277, opened 2026-06-20
fix(web): unify filter toolbar row height across pages

3Hren:
Page toolbar rows rendered at slightly different heights between pages: those with the segmented filter control sat a few pixels taller than pages whose only control was the filter search input. This grows the toolbar filter input to match the segmented filter control, so every toolbar row keeps the same height across pages while the rows stay flexibly sized.

Verified live across all affected pages: every toolbar row now measures the same height and the filter inputs match the segmented control, with the page-header search and form inputs left unchanged.


3Hren:
@codex review

PR #1279, opened 2026-06-21
refactor(web): route shared imports through a core path alias

3Hren:
First step toward making the web UI modular like the rest of the repo: shared code is now imported through a single core path alias instead of deep relative paths, so pages become position-independent ahead of co-locating per-module web sources with their backend.

- Route every page and the app shell to the shared web code through one stable alias rather than long relative paths.
- Add the matching bundler and type-checker configuration so the alias resolves in the build, the test runner, and the dev server.
- Widen the dev server's allowed file roots and pin a single React copy so a later phase can build co-located module sources from one bundle.
- Pure import rewrite with no runtime behavior change; build, unit tests, and dev server all verified.


3Hren:
@codex review

PR #1280, opened 2026-06-21
refactor(web): relocate shared code into src/core

3Hren:
Second step of the web UI modularization: the shared code now lives in a dedicated core directory, cleanly separated from the application shell, ahead of co-locating per-module sources.

- Move the shared web modules into a dedicated core directory under the app source.
- Repoint the core alias and the type-checker path mapping at the new location.
- Adjust the handful of internal references that pointed outside the shared code.
- Pure relocation: consumers already import through the alias, so there are no call-site or behavior changes; build, unit tests, and dev server all verified.


3Hren:
@codex review

PR #1281, opened 2026-06-21
refactor(web): derive pages from a discovered manifest registry

3Hren:
Third step of the web UI modularization: pages now register themselves through a build-time manifest registry instead of three hand-maintained lists, which is the discovery mechanism the per-module web subtrees will plug into.

- Replace the hand-edited route table, sidebar list, and idle-prefetch list with a single registry that discovers a per-page manifest at build time and derives all three from it.
- Each page declares its route, section, title, icon, redirects, and lazily-loaded body in a small manifest beside it; adding a page no longer touches central files.
- Keep the page-id tuple as the hand-authored source of truth and guard the page-to-manifest correspondence with a discovery test.
- Pure refactor: routes, redirects, sidebar order, prefetching, and per-page code-splitting are preserved; build, unit tests, and dev server all verified.


3Hren:
@codex review

PR #1284, opened 2026-06-21
refactor(web): co-locate the neighbours web UI with its operator

3Hren:
Relocates the Neighbours operator page from `web/src/pages/operators/neighbours` to `operators/neighbours/web`, co-locating the UI with its operator the same way the decap page now lives under `modules/decap/web`.

The move is a pure rename. The manifest registry already discovers sibling web sources via the `operators/*/web/manifest.ts` glob, so the page's id, route, and nav entry are unchanged and the registry bijection test stays satisfied.


3Hren:
@codex review

3Hren:
Addressed the P2: co-locating the spec moved it outside Vitest's web-relative test root, so it silently dropped from CI. Added explicit `test.include` globs for the sibling `modules|operators|devices/*/web` roots in `web/vite.config.ts`; the neighbours spec now runs again (test files 21 → 22, +25 assertions locally). This also future-proofs every later co-located move.

@codex review

3Hren:
@codex review

PR #1285, opened 2026-06-21
refactor(web): co-locate the route-operator web UI with its operator

3Hren:
Relocates the route-operator RIB page from `web/src/pages/operators/route` to `operators/route/web`, placing the UI alongside the operator's existing `cli`, `cmd`, `internal`, and `operatorpb` code, following the neighbours and decap moves.

The move is a pure rename. The manifest registry already discovers sibling web sources via the `operators/*/web/manifest.ts` glob, so the page's id, route, and nav entry are unchanged and the registry bijection test stays satisfied. The co-located `utils.test.ts` keeps running in CI through the sibling test-include globs added previously.


3Hren:
@codex review

PR #1286, opened 2026-06-21
refactor(web): co-locate the fwstate module web UI with its module

3Hren:
Relocates the fwstate page from `web/src/pages/modules/fwstate` to `modules/fwstate/web`, co-locating the UI with its module like the decap, neighbours, and route moves.


3Hren:
@codex review

PR #1287, opened 2026-06-21
refactor(web): co-locate the pdump module web UI with its module

3Hren:
Relocates the pdump page from `web/src/pages/modules/pdump` to `modules/pdump/web`, co-locating the UI with its module like the preceding moves.


3Hren:
@codex review

PR #1288, opened 2026-06-21
refactor(web): co-locate the route module web UI with its module

3Hren:
Relocates the route module FIB page from `web/src/pages/modules/route` to `modules/route/web`, co-locating the UI with its module. This is the route module page (FIB), distinct from the route operator page relocated earlier.


3Hren:
@codex review

PR #1289, opened 2026-06-21
refactor(web): promote shared module-page UI into core

3Hren:
The acl and forward pages shared a small UI helper set that lived under `web/src/pages/modules/_shared`. Since those pages are about to relocate into their own module directories, the shared code is promoted into core (`web/src/core/components`) so it has a single position-independent home and both pages can import it through the `@yanet/core` alias.

The moved files are unchanged; only the import paths in acl and forward are updated. No behavior changes.


3Hren:
@codex review

PR #1290, opened 2026-06-21
refactor(web): co-locate the acl module web UI with its module

3Hren:
Relocates the acl page from `web/src/pages/modules/acl` to `modules/acl/web`, co-locating the UI with its module.

The move is a pure rename; the page's shared helpers were already promoted to core in an earlier change. The manifest registry discovers the page via the `modules/*/web/manifest.ts` glob, so its id, route, and nav entry are unchanged and the registry bijection test stays satisfied. The co-located YAML-import web worker is referenced relatively and travels with the move.


3Hren:
@codex review

PR #1291, opened 2026-06-21
refactor(web): co-locate the forward module web UI with its module

3Hren:
Relocates the forward page from `web/src/pages/modules/forward` to `modules/forward/web`, co-locating the UI with its module and completing the per-module relocations.

The move is a pure rename; the page's shared helpers were already promoted to core earlier. The manifest registry discovers the page via the `modules/*/web/manifest.ts` glob, so its id, route, and nav entry are unchanged and the registry bijection test stays satisfied.


3Hren:
@codex review

PR #1292, opened 2026-06-21
refactor(web): co-locate the built-in pages under web/builtin

3Hren:
Relocates the built-in pages (dashboard, devices, functions, inspect, pipelines) and their shared helpers from `web/src/pages/builtin` to `web/builtin/<svc>`, completing the modular co-location and emptying `web/src/pages` entirely. The manifest registry now discovers built-in pages through a `web/builtin/*` glob alongside the module, device, and operator roots.

The page moves are renames. The move also surfaced a layering quirk: a core hook served only the inspect page yet reached back into it, so it was relocated next to its single consumer and dropped from the core surface, removing the core-to-page dependency. Page ids, routes, and nav entries are unchanged and the registry bijection test stays satisfied.


3Hren:
@codex review

PR #1308, opened 2026-06-23
feat(route): select best routes per gateway from its local RIB

egnees:
The route operator pushes a forwarding table to several gateways, but each gateway can forward only through its own devices. Choosing the best route globally could hand a gateway a route whose egress device it lacks, leaving the prefix blackholed there.

- Each gateway selects its best routes only among the routes reachable through its own devices.
- A gateway with no path to the globally best route falls back to the best route it can actually use, instead of dropping the prefix.

### DIFF
--- a/operators/route/internal/operator/fib_build.go
+++ b/operators/route/internal/operator/fib_build.go
@@ -34,40 +34,19 @@ type FIBBuildStats struct {
 	NeighbourNotFound int
 	HardwareRoutes    int
 	PrefixesAdded     int
-	// FilteredRoutes counts routes excluded by the per-source best-group filter.
+	// FilteredRoutes counts eligible routes dropped because a better route
+	// of the same source exists.
 	FilteredRoutes int
-}
-
-// filterFIBEntries returns the subset of entries whose nexthops include at
-// least one device present in the given set.
-//
-// When the device set is empty every entry is returned unchanged.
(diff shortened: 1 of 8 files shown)


operators/route/internal/operator/fib_build.go
```diff
@@ -34,43 +34,27 @@ type FIBBuildStats struct {
...
+// A gateway forwards only through its own devices, so a route is eligible
+// only when its nexthop resolves to a neighbour reachable on one of the
+// given devices; an empty device set imposes no such restriction. For each
+// prefix the best routes per source among the eligible ones are kept, so a
+// gateway can use a lower-priority route on its own device rather than a
+// globally best one it cannot reach.
 func BuildFIB(
 	ribDump maptrie.MapTrie[netip.Prefix, netip.Addr, rib.RoutesList],
 	neighbours neigh.NexthopCacheView,
+	devices map[string]struct{},
```
3Hren: If this is optional - make options. Thus, no `nil` arguments will be passed in tests

GeorgyKirichenko: I think we should just filter neighbours out

egnees: Fixed.

sakateka (approved):
LGTM

GeorgyKirichenko (approved):

PR #1309, opened 2026-06-24
feat(forward): per-rule metrics

kt0ns:


modules/forward/controlplane/forwardpb/v1/forward.proto
```diff
@@ -20,6 +21,9 @@ service ForwardService {
   // Allows to delete forward config if it is not referenced
   // by any pipeline.
   rpc DeleteConfig(DeleteConfigRequest) returns (DeleteConfigResponse);
+
+  // GetMetrics returns a snapshot of forward module metrics.
+  rpc GetMetrics(GetMetricsRequest) returns (GetMetricsResponse);
```
3Hren, line 29: Separate service

3Hren (changes requested):

modules/forward/controlplane/service.go
```diff
@@ -25,6 +27,10 @@ type Backend interface {
 	UpdateModule(name string, rules []cforward.ForwardRule) (ModuleHandle, error)
 	// DeleteModule removes a module config.
 	DeleteModule(name string) error
+	// DPConfig returns a handle to the dataplane configuration used to
+	// read module counters. It may return nil if no dataplane configuration
+	// is attached.
+	DPConfig() *ffi.DPConfig
```
3Hren: ffi leak. What tests/mocks should return?

controlplane/ffi/shm.go
```diff
@@ -900,3 +900,43 @@ func (m *DPConfig) AllModulePositions(moduleType string) iter.Seq[ModuleReferenc
 		}
 	}
 }
+
+type CountersConfig struct {
```
saushew: I don't see any real value in this abstraction, while it creates the false impression that counters are an object separate from the dataplane.

modules/forward/controlplane/service.go
```diff
@@ -26,7 +28,6 @@ type Backend interface {
 	// DeleteModule removes a module config.
 	DeleteModule(name string) error
 }
-
```
saushew, line 29: fmt pls

saushew (changes requested):

PR #1310, opened 2026-06-24
fix(dataplane): cap the worker pending-mbuf ring to prevent mempool exhaustion

3Hren:
The worker's pending-mbuf backpressure guard compared a counter against itself, so the cap never engaged and the ring of in-flight mbufs could grow without bound whenever a consumer worker's transmit completions lagged the producer.

- Once the ring passed its capacity, slot reuse aliased still-pending entries and silently dropped mbuf pointers, leaking them out of the mempool until the receive path stalled. This is reachable under ordinary traffic with a back-pressured egress (link down, full hardware transmit ring, or a slow virtual device), with no malformed packet required.
- Restores the depth check so the ring is bounded and excess packets are dropped instead of leaked.
- Adds a focused regression test that fails against the broken guard and passes with the fix.

### DIFF
--- a/dataplane/pending_ring.h
+++ b/dataplane/pending_ring.h
@@ -1,0 +1,38 @@
+#pragma once
+
+#include <stdint.h>
+
+struct rte_mbuf;
+
+// One slot in the pending-mbuf ring.
+//
+// Stores the mbuf and the refcount baseline at push time so the sweep can
+// decide when the consumer has released its reference.
+struct worker_pending_mbuf {
+	struct rte_mbuf *mbuf;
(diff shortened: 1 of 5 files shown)


3Hren:
Closing in favour of #1302, which fixes the same dead backpressure guard (the pending-mbuf depth check compared a counter against itself) as part of a broader per-connection pending-queue redesign. #1302 is the maintainer's architectural fix and supersedes this one-line patch; no reason to carry a competing PR. The discriminating regression test from this branch can be ported onto #1302's per-pipe structure separately.

PR #1311, opened 2026-06-24
feat(web): redesign Devices page and live trafgen generator controls

3Hren:
Redesigns the Devices page to match the new design. The device list collapses to an icon rail with hover previews, the type filter becomes a segmented control, metric cards are taller with bottom-anchored sparklines, and per-device counters move into the detail header next to a tinted device icon and rounded kind badges.

Adds live generator controls for trafgen devices. The transmit rate and the pcap frame set are now applied to the running device through their own actions, independent of Save, which now persists only pipeline assignments. The transmit rate field is validated and Apply is enabled only for natural numbers.

Lets a pcap be loaded and previewed locally before it is uploaded to the device.

Makes the frames table collapsible and aligned with its toggle, and keeps the RX and TX labels fixed in the device list so they no longer shift with value width.


3Hren:
@codex review

PR #1313, opened 2026-06-25
refactor(dataplane): empty the source list on packet-list concat

3Hren:
Concatenating two packet lists now leaves the source list empty: the move resets the source after splicing its packets onto the destination. Every moved packet therefore has a single owner, which removes a class of latent stale-alias bugs where a drained source kept pointing at packets it no longer owned.

This folds a reset that callers previously performed by hand into the move itself, dropping the redundant per-call-site cleanup across the pipeline, the packet-front helpers, and the dataplane test harness.

The change is zero-cost on the per-packet hot path, validated at the assembly level. Where the source is a dead local — the inlined fast paths — the compiler eliminates the added reset entirely: the generated code is byte-for-byte identical to before at -O1, -O2, and -O3 (gcc 13.3, -march=haswell, matching the project's optimization level). As a negative control, when the source outlives the call the reset stores survive exactly as they must, and at those escaping sites the stores were already emitted by the old hand-written reset — so nothing is added there either.

The dataplane build and the single-chain dispatch regression tests pass.

### DIFF
--- a/lib/dataplane/module/packet_front.h
+++ b/lib/dataplane/module/packet_front.h
@@ -52,11 +52,9 @@ packet_front_drop(struct packet_front *packet_front, struct packet *packet) {
 static inline void
 packet_front_switch(struct packet_front *packet_front) {
 	packet_list_concat(&packet_front->input, &packet_front->output);
-	packet_list_init(&packet_front->output);
 }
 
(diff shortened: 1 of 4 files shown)


3Hren:
@codex review

PR #1320, opened 2026-06-30
feat(forward): per-rule metrics

kt0ns:


modules/forward/api/controlplane.c
```diff
@@ -82,6 +82,11 @@ forward_module_config_free(struct cp_module *cp_module) {
 	);
 }
 
+struct dp_config *
+forward_module_dp_config(struct cp_module *cp_module) {
```
saushew: Seems that this code is unused

modules/forward/controlplane/service.go
```diff
@@ -38,6 +42,7 @@ type ForwardService struct {
 	mu      sync.Mutex
 	backend Backend
 	configs map[string]forwardConfig
+	metrics *grpcmetrics.ServerMetrics
```
saushew: This field is not initialised in the `ForwardService` constructor by now

saushew (changes requested):

modules/forward/api/controlplane.c
kt0ns: Fixed. Leftover from old implementation

modules/forward/controlplane/service.go
kt0ns: Fixed. This was a leftover from the ACL module template. Since module collects only per-rule metrics, it has been removed

saushew (approved):

3Hren (approved):

PR #1323, opened 2026-07-01
feat(vlan): unify device CLI output via the shared output layer

3Hren:
- Adopt the shared unified-output layer in the VLAN device CLI so it gains a machine-readable JSON output mode alongside consistent human and error rendering.
- Replace the ad-hoc logging-based success and error reporting with the shared structured output and error-reporting path.
- Remove the unused output-format scaffolding and the dependencies that only supported it.


3Hren:
@codex review

PR #1325, opened 2026-07-01
feat(decap): unify CLI output via the shared output layer

3Hren:
- Adopt the shared unified-output layer in the decap module CLI so its list and show commands gain a machine-readable JSON output mode alongside consistent human and error rendering.
- Report the update command as a structured success and route all failures through the shared error-reporting path.
- Replace the per-command output-format switch with the global format flag and drop the now-unused direct JSON dependency.


3Hren:
@codex review

PR #1326, opened 2026-07-01
refactor(cli): bundle client, endpoint and error mapping into one wrapper

3Hren:
- Introduce a shared wrapper in the CLI core that pairs a gRPC client with its endpoint and service name, so a CLI no longer threads the endpoint through its own struct or re-declares the per-command error-mapping helpers.
- Convert the decap and trafgen CLIs to the wrapper as the reference; behaviour is unchanged.
- Leave the remaining single-client CLIs and the two-client CLIs (which share one channel across clients) for a follow-up consistency pass.


3Hren:
@codex review

PR #1327, opened 2026-07-01
refactor(plain): adopt the shared client wrapper in the device CLI

3Hren:
- Move the plain device CLI onto the shared client/endpoint wrapper introduced in the CLI core, dropping its own endpoint field and per-command error-mapping helpers.
- Behaviour is unchanged.


3Hren:
@codex review

PR #1333, opened 2026-07-01
refactor(gateway): adopt the shared client wrapper in the gateway CLI

3Hren:
- Move the gateway CLI onto the shared client/endpoint wrapper introduced in the CLI core, dropping its own endpoint field and per-command error-mapping helpers.
- Connection-failure errors now report the uniform connect action used by the other migrated CLIs; all other behaviour is unchanged.


3Hren:
@codex review

PR #1335, opened 2026-07-01
refactor(acl): adopt the shared client wrapper in the acl CLI

3Hren:
- Move the acl CLI onto the shared client wrapper: its two clients now each hold their own wrapper over a single shared channel, dropping the endpoint field and per-command error-mapping helper.
- Behaviour is unchanged, including reporting metrics errors under the acl service name.


3Hren:
@codex review

PR #1336, opened 2026-07-01
refactor(route): adopt the shared client wrapper in the route operator CLI

3Hren:
- Move the route operator CLI onto the shared client wrapper: the route and readiness clients now each hold their own wrapper over a single shared channel, dropping the endpoint field and per-command error-mapping helper.
- Behaviour is unchanged.


3Hren:
@codex review

PR #1337, opened 2026-07-01
refactor(function): adopt the shared client wrapper in the function CLI

3Hren:
- Move the function CLI onto the shared client wrapper, dropping its own endpoint field.
- Expose the wrapper's endpoint so the CLI's friendly not-found rewriting keeps working unchanged.
- Behaviour is unchanged.


3Hren:
@codex review

PR #1338, opened 2026-07-01
refactor(pipeline): adopt the shared client wrapper in the pipeline CLI

3Hren:
- Move the pipeline CLI onto the shared client wrapper, dropping its own endpoint field.
- Keep the CLI's friendly not-found rewriting via the wrapper's endpoint accessor.
- Behaviour is unchanged.


3Hren:
@codex review

PR #1339, opened 2026-07-01
refactor(pipeline): route the operator metrics command through the shared wrapper

3Hren:
- Route the pipeline operator's metrics command through the shared client wrapper, so the whole CLI uses it consistently instead of a hand-rolled connection.
- Behaviour is unchanged.


3Hren:
@codex review

PR #1342, opened 2026-07-01
feat(dscp): unify CLI output via the shared output layer

3Hren:
- Adopt the shared output stack and client wrapper in the dscp module CLI, so it gains a machine-readable JSON mode alongside consistent human and error rendering.
- Replace the per-command output-format switch with the global format flag, and drop the now-unused direct JSON dependency.
- Request behaviour is unchanged.


3Hren:
@codex review

PR #1343, opened 2026-07-01
feat(route-mpls): unify CLI output via the shared output layer

3Hren:
- Adopt the shared output stack and client wrapper in the route-mpls module CLI, so it gains a machine-readable JSON mode alongside consistent human and error rendering.
- Move the CLI onto the shared endpoint and authentication flags in place of its hand-rolled connection.
- Drop the now-unused direct dependencies. Request behaviour is unchanged.


3Hren:
@codex review

PR #1346, opened 2026-07-01
feat(ttlmap): add incremental bucket iterator

egnees:
- Adds a bucket-at-a-time iterator (ttlmap_bucket_iter / TTLMAP_ITER_NEXT) that copies one bucket under lock per call, for callers that need to stream a map's contents across multiple steps instead of holding one long lock over the whole pass.
- Exposes the bucket entry count as a public constant (TTLMAP_BUCKET_SIZE) for callers sizing per-call buffers.

### DIFF
--- a/common/ttlmap/detail/bucket.h
+++ b/common/ttlmap/detail/bucket.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <stdalign.h>
+
 #include "key_value.h"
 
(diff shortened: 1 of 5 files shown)


egnees:
@codex review

moonug (approved):

GeorgyKirichenko (approved):

PR #1347, opened 2026-07-01
refactor(cli): add a shared connection primitive and collapse the service constructors

3Hren:
- Introduce a shared connection that establishes the channel once and wraps one service per gRPC client, so multi-service CLIs (acl, the route operator) stop hand-rolling the connect boilerplate.
- Make the raw client wrapper internal, leaving one clear entry point per shape: single-service CLIs connect a service directly, multi-service CLIs spawn several from one connection.
- Move the function and pipeline CLIs onto the shared entry point too; request behaviour is unchanged, and their connection-failure errors now report the uniform connect action like every other CLI.

### DIFF
--- a/cli/core/src/client.rs
+++ b/cli/core/src/client.rs
@@ -74,6 +74,30 @@ pub async fn connect(args: &ConnectionArgs) -> Result<LayeredChannel, Connection
     Ok(auth.layer(channel))
 }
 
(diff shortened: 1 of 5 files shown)


3Hren:
@codex review

3Hren:
@codex review

PR #1354, opened 2026-07-02
feat(grpcmetrics): add a stream server interceptor

3Hren:
- Adds a stream server interceptor recording the same started, handled and handling-duration metric families as the unary one, with the stream kind derived from the handler flags.
- Generalizes the internal series machinery over the call type and exports the service label name for consumers that build retention predicates.
- For streams the extra-label hook is invoked without a request message, which is now documented.


3Hren:
@codex review

3Hren:
@codex review

PR #1359, opened 2026-07-02
fix(pdump): resolve duplicate short flag in the read command

3Hren:
- Fixes a startup panic in the pdump read command caused by two options sharing the same short flag.
- The packet count option is now long-only, keeping the short flag for the config name consistent with the other subcommands.


3Hren:
@codex review

PR #1361, opened 2026-07-03
feat(acl): add dataset-driven classification benchmark

3Hren:
- Adds a standalone benchmark measuring ACL classification throughput on a large production-shaped ruleset under flow-diverse traffic.
- The ruleset and traffic are synthesized deterministically by default, so the benchmark runs without any external files; real rule dumps (JSON) and traffic samples (pcap/pcapng) can be supplied via environment variables and are never committed to the repository.
- Batch sizes sweep the concurrent-flow working set to expose cache effects on production-scale tables that single-flow benchmarks cannot see.


3Hren:
@codex review

PR #1364, opened 2026-07-03
fix(dataplane_ut): add optional payload reset between benchmark rounds

3Hren:
- The benchmark runner recycles one packet set across all rounds and restored only the routing fields, so modules that rewrite packet contents in place drifted off the measured path: with the route module, packets exhausted their TTL after the first 63 rounds and the benchmark silently measured the early-drop path instead of forwarding (58.6 Mpps reported versus 31.1 Mpps of actual forwarding on the same setup).
- Bench now takes the packet slice plus functional options; the new WithPayloadReset option snapshots every payload at capture time and restores it between rounds. Benchmarks without the option keep byte-identical behavior and timing. Only payload bytes are restored, so in-place header rewrites are covered while geometry-changing modules stay out of scope, as documented.
- Adds a route forwarding benchmark using the new option — the first benchmark for the route module; it was built to measure a proposed route optimization and correctly refuted it (#1363).
- The shared route test helpers now accept any test type, following the ACL test package precedent.

### DIFF
--- a/bindings/go/dataplane_ut/bench_test.go
+++ b/bindings/go/dataplane_ut/bench_test.go
@@ -84,7 +84,7 @@ func BenchmarkPipelineRound(b *testing.B) {
 				require.NoError(b, err)
 				pkts[idx] = pkt
 			}
-			h.Bench(b, pkts...)
+			h.Bench(b, pkts)
 		})
 	}
 }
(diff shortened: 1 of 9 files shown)


3Hren:
@codex review

3Hren:
@codex review

PR #1376, opened 2026-07-07
feat(dataplane): guard .so module plugins by ABI version

3Hren:
The dataplane loads packet-processing modules as external .so plugins from a configured directory, but the loader performed no version check — a plugin built against a different dataplane layout could silently corrupt shared memory.

Each plugin now embeds an ABI-version symbol that the loader verifies on load, refusing any plugin with a missing or mismatched version.

A compile-time assertion over the shared dataplane structures fails the build whenever that ABI changes without a version bump, so the version cannot silently fall out of date.

Plugin load errors now fail dataplane startup instead of being silently skipped.


3Hren:
@codex review

PR #1377, opened 2026-07-07
feat(fwstate): metrics service

saushew:
Adds a full metrics service for the `fwstate` module: dataplane packet/byte counters, per-config/per-AF map-statistics gauges, and gRPC call metrics. Metrics are exposed through a new `MetricsService` gRPC service and rendered as structured tables by the `yanet-cli-fwstate` CLI.

### DIFF
--- a/modules/acl/controlplane/mod.go
+++ b/modules/acl/controlplane/mod.go
@@ -65,7 +65,7 @@ func NewACLModule(cfg *Config, log *zap.Logger) (*ACLModule, error) {
 		agent,
 		aclAdapter,
 		fwstate.WithLog(log),
-		fwstate.WithMetrics(),
+		fwstate.WithMetrics(fwstate.NewMetricsFactory()),
 	)
 	fwstateMetricsService := fwstate.NewMetricsService(fwstateService)
 
--- a/modules/acl/controlplane/mod.go
+++ b/modules/acl/controlplane/mod.go
@@ -14,10 +14,8 @@ import (
 )
 
 const (
-	agentName                 = "acl"
-	serviceName               = "modules.acl.controlplane.aclpb.v1.ACLService"
-	fwstateServiceName        = "modules.fwstate.controlplane.fwstatepb.v1.FWStateService"
-	fwstateMetricsServiceName = "modules.fwstate.controlplane.fwstatepb.v1.MetricsService"
+	agentName   = "acl"
+	serviceName = "modules.acl.controlplane.aclpb.v1.ACLService"
 )
 
 // ACLModule is a control-plane component for ACL (Access Control List) module
@@ -95,8 +93,8 @@ func (m *ACLModule) ServicesNames() []string {
 	return []string{
 		serviceName,
 		aclpb.MetricsService_ServiceDesc.ServiceName,
-		fwstateServiceName,
-		fwstateMetricsServiceName,
+		fwstate.FWStateServiceName,
+		fwstate.FWStateMetricsServiceName,
 	}
 }
 
--- a/modules/acl/controlplane/mod.go
+++ b/modules/acl/controlplane/mod.go
@@ -14,21 +14,23 @@ import (
 )
 
 const (
-	agentName          = "acl"
-	serviceName        = "modules.acl.controlplane.aclpb.v1.ACLService"
-	fwstateServiceName = "modules.fwstate.controlplane.fwstatepb.v1.FWStateService"
+	agentName                 = "acl"
+	serviceName               = "modules.acl.controlplane.aclpb.v1.ACLService"
+	fwstateServiceName        = "modules.fwstate.controlplane.fwstatepb.v1.FWStateService"
+	fwstateMetricsServiceName = "modules.fwstate.controlplane.fwstatepb.v1.MetricsService"
 )
 
 // ACLModule is a control-plane component for ACL (Access Control List) module
 // with integrated firewall state management.
 type ACLModule struct {
-	cfg            *Config
-	shm            *ffi.SharedMemory
-	agent          *ffi.Agent
-	aclService     *ACLService
-	metricsService *MetricsService
-	fwstateService *fwstate.FWStateService
-	log            *zap.Logger
+	cfg                   *Config
+	shm                   *ffi.SharedMemory
+	agent                 *ffi.Agent
+	aclService            *ACLService
+	metricsService        *MetricsService
+	fwstateService        *fwstate.FWStateService
+	fwstateMetricsService *fwstate.MetricsService
+	log                   *zap.Logger
 }
 
 // NewACLModule creates a new ACL module instance.
@@ -61,16 +63,23 @@ func NewACLModule(cfg *Config, log *zap.Logger) (*ACLModule, error) {
 	metricsService := NewMetricsService(aclService)
 
 	aclAdapter := NewACLAdapter(aclService)
-	fwstateService := fwstate.NewFWStateService(agent, aclAdapter, log)
+	fwstateService := fwstate.NewFWStateService(
+		agent,
+		aclAdapter,
+		fwstate.WithLog(log),
+		fwstate.WithMetrics(),
+	)
+	fwstateMetricsService := fwstate.NewMetricsService(fwstateService)
 
 	return &ACLModule{
-		cfg:            cfg,
-		shm:            shm,
-		agent:          agent,
-		aclService:     aclService,
-		metricsService: metricsService,
-		fwstateService: fwstateService,
-		log:            log,
+		cfg:                   cfg,
+		shm:                   shm,
+		agent:                 agent,
+		aclService:            aclService,
+		metricsService:        metricsService,
+		fwstateService:        fwstateService,
+		fwstateMetricsService: fwstateMetricsService,
+		log:                   log,
 	}, nil
 }
 
@@ -83,23 +92,31 @@ func (m *ACLModule) Endpoint() string {
 }
 
 func (m *ACLModule) ServicesNames() []string {
-	return []string{serviceName, aclpb.MetricsService_ServiceDesc.ServiceName, fwstateServiceName}
+	return []string{
+		serviceName,
+		aclpb.MetricsService_ServiceDesc.ServiceName,
+		fwstateServiceName,
+		fwstateMetricsServiceName,
+	}
 }
 
 func (m *ACLModule) RegisterService(server *grpc.Server) {
 	aclpb.RegisterACLServiceServer(server, m.aclService)
 	aclpb.RegisterMetricsServiceServer(server, m.metricsService)
 	fwstatepb.RegisterFWStateServiceServer(server, m.fwstateService)
+	fwstatepb.RegisterMetricsServiceServer(server, m.fwstateMetricsService)
 }
 
 // UnaryServerInterceptors returns the gRPC unary interceptors for this module.
 func (m *ACLModule) UnaryServerInterceptors() []grpc.UnaryServerInterceptor {
-	si := m.aclService.UnaryServerInterceptor()
-	if si == nil {
-		return nil
+	var interceptors []grpc.UnaryServerInterceptor
+	if si := m.aclService.UnaryServerInterceptor(); si != nil {
+		interceptors = append(interceptors, si)
 	}
-
-	return []grpc.UnaryServerInterceptor{si}
+	if si := m.fwstateService.UnaryServerInterceptor(); si != nil {
+		interceptors = append(interceptors, si)
+	}
+	return interceptors
 }
 
 // ACLAdapter returns an adapter for fwstate module integration.
--- a/modules/fwstate/cli/src/main.rs
+++ b/modules/fwstate/cli/src/main.rs
@@ -18,7 +18,7 @@ use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::{Status, codec::CompressionEncoding};
 use ync::{
-    client::{ConnectionArgs, LayeredChannel, Service},
+    client::{Connection, ConnectionArgs, LayeredChannel, Service},
     display::print_table_from_entries,
     errors::Error,
     output::{self, CommonFormat},
@@ -75,19 +75,17 @@ pub struct FWStateService {
 
 impl FWStateService {
     pub async fn new(connection: &ConnectionArgs) -> Result<Self, Error> {
-        let service = Service::connect(connection, SERVICE_NAME, |channel| {
+        let conn = Connection::connect(connection).await?;
+        let service = Service::new(&conn, SERVICE_NAME, |channel| {
             FwStateServiceClient::new(channel)
                 .send_compressed(CompressionEncoding::Gzip)
                 .accept_compressed(CompressionEncoding::Gzip)
-        })
-        .await?;
-
-        let metrics = Service::connect(connection, METRICS_SERVICE_NAME, |channel| {
+        });
+        let metrics = Service::new(&conn, METRICS_SERVICE_NAME, |channel| {
             MetricsServiceClient::new(channel)
                 .send_compressed(CompressionEncoding::Gzip)
                 .accept_compressed(CompressionEncoding::Gzip)
-        })
-        .await?;
+        });
 
         Ok(Self { service, metrics })
     }
--- a/modules/fwstate/cli/src/main.rs
+++ b/modules/fwstate/cli/src/main.rs
@@ -384,14 +384,17 @@ impl FWStateService {
             .await
             .map_err(self.metrics.status("metrics"))?
             .into_inner();
-        let label_filters: Vec<(&str, &str)> = cmd
-            .labels
-            .iter()
-            .filter_map(|s| {
-                let mut it = s.splitn(2, '=');
-                Some((it.next()?, it.next()?))
-            })
-            .collect();
+        let mut label_filters: Vec<(&str, &str)> = Vec::with_capacity(cmd.labels.len());
+        for s in &cmd.labels {
+            match s.split_once('=') {
+                Some(kv) => label_filters.push(kv),
+                None => {
+                    return Err(self
+                        .metrics
+                        .invalid("metrics", format!("invalid label filter {s:?}, expected KEY=VALUE")));
+                }
+            }
+        }
 
         let metrics: Vec<Metric> = response
             .metrics
@@ -755,62 +758,66 @@ fn print_metrics_table(metrics: &[Metric]) {
         let mut pair_order: Vec<String> = Vec::new();
         let mut pair_map: HashMap<String, CounterPair> = HashMap::new();
 
+        // Generic counters (the default arm of emitCounterMetrics) are all
+        // exported under the shared names fwstate_counter_packets /
+        // fwstate_counter_bytes and distinguished only by the "counter" label.
+        // After stripping the fwstate_ prefix and the _packets/_bytes suffix
+        // they all collapse to the same base "counter", so the pair key must
+        // include the "counter" label value to keep them as separate rows
+        // instead of overwriting each other (last write wins). Dedicated
+        // counters carry no "counter" label, so the suffix is empty and their
+        // base is unaffected.
         for m in counters {
             let val = m.value.unwrap_or(0.0) as u64;
             let stripped = m.name.strip_prefix("fwstate_").unwrap_or(&m.name);
-            if let Some(base) = stripped.strip_suffix("_packets") {
-                let pair = pair_map.entry(base.to_string()).or_insert_with(|| {
-                    pair_order.push(base.to_string());
-                    CounterPair {
-                        display: metric_display_name(base),
-                        packets: None,
-                        bytes: None,
-                        entries: None,
-                    }
-                });
-                pair.packets = Some(val);
-            } else if let Some(base) = stripped.strip_suffix("_bytes") {
-                let pair = pair_map.entry(base.to_string()).or_insert_with(|| {
-                    pair_order.push(base.to_string());
-                    CounterPair {
-                        display: metric_display_name(base),
-                        packets: None,
-                        bytes: None,
-                        entries: None,
-                    }
-                });
-                pair.bytes = Some(val);
-            } else if let Some(base) = stripped.strip_suffix("_entries") {
-                // State-table entry counters (e.g. sync insert counters)
-                // count frames, not packets/bytes; render under Entries.
-                let pair = pair_map.entry(base.to_string()).or_insert_with(|| {
-                    pair_order.push(base.to_string());
-                    CounterPair {
-                        display: metric_display_name(base),
-                        packets: None,
-                        bytes: None,
-                        entries: None,
-                    }
-                });
-                pair.entries = Some(val);
+            let counter_label = m.label_value("counter").unwrap_or("");
+
+            // Determine the semantic suffix and the base name shared by the
+            // packets/bytes series of the same counter. The pair key and the
+            // display name must be built from the base (suffix stripped),
+            // otherwise fwstate_rx_packets and fwstate_rx_bytes get different
+            // keys and end up on two separate rows instead of one.
+            let (base, is_packets, is_bytes) = if let Some(b) = stripped.strip_suffix("_packets") {
+                (b, true, false)
+            } else if let Some(b) = stripped.strip_suffix("_bytes") {
+                (b, false, true)
             } else {
-                let pair = pair_map.entry(stripped.to_string()).or_insert_with(|| {
-                    pair_order.push(stripped.to_string());
-                    CounterPair {
-                        display: metric_display_name(stripped),
-                        packets: None,
-                        bytes: None,
-                        entries: None,
-                    }
-                });
+                (stripped, false, false)
+            };
+
+            let display = if counter_label.is_empty() {
+                metric_display_name(base)
+            } else {
+                metric_display_name(counter_label)
+            };
+            let pair_key = format!("{base}\0{counter_label}");
+
+            let pair = pair_map.entry(pair_key.clone()).or_insert_with(|| {
+                pair_order.push(pair_key.clone());
+                CounterPair {
+                    display,
+                    packets: None,
+                    bytes: None,
+                    entries: None,
+                }
+            });
+
+            if is_packets {
+                pair.packets = Some(val);
+            } else if is_bytes {
+                pair.bytes = Some(val);
+            } else {
+                // State-table entry counters (e.g. sync insert counters) and
+                // any unsuffixed counter count frames/items, not
+                // packets/bytes; render under Entries.
                 pair.entries = Some(val);
             }
         }
 
         let rows: Vec<CounterRow> = pair_order
             .iter()
-            .map(|base| {
-                let p = &pair_map[base];
+            .map(|pair_key| {
+                let p = &pair_map[pair_key];
                 CounterRow {
                     counter: p.display.clone(),
                     packets: p.packets.map(format_number).unwrap_or_else(|| "-".into()),
--- a/modules/fwstate/cli/src/main.rs
+++ b/modules/fwstate/cli/src/main.rs
@@ -1,26 +1,31 @@
 use core::{fmt, net::Ipv6Addr, time::Duration};
-use std::time::UNIX_EPOCH;
+use std::{collections::HashMap, time::UNIX_EPOCH};
 
-use args::{DeleteCmd, DirectionArg, EntriesCmd, LinkCmd, ModeCmd, ShowCmd, StatsCmd, UpdateCmd};
+use args::{DeleteCmd, DirectionArg, EntriesCmd, LinkCmd, MetricsCmd, ModeCmd, ShowCmd, StatsCmd, UpdateCmd};
 use clap::{ArgAction, CommandFactory, Parser};
 use clap_complete::CompleteEnv;
 use commonpb::pb::IpAddress;
 use fwstatepb::{
-    DeleteConfigRequest, Direction, GetStatsRequest, LinkFwStateRequest, ListConfigsRequest, ListEntriesRequest,
-    ShowConfigRequest, UpdateConfigRequest, fw_state_service_client::FwStateServiceClient,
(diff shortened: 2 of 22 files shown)


sakateka (dismissed):

sakateka (dismissed):

modules/fwstate/controlplane/metrics.go
```diff
@@ -0,0 +1,246 @@
...
+			if packets == 0 && bytes == 0 {
+				continue
+			}
+
+			switch counter.Name {
+			case "fwstate_sync_packets":
+				result = append(result,
+					makeCounter("fwstate_sync_packets_packets", packets, baseLabels...),
+					makeCounter("fwstate_sync_packets_bytes", bytes, baseLabels...),
+				)
```
3Hren: Doubled `_packets` suffix.

The C counter is already named `fwstate_sync_packets`, so the exported series become `fwstate_sync_packets_packets` and `fwstate_sync_packets_bytes`.

Every other case in this switch derives `<base>_packets` / `<base>_bytes` from a base that does not itself end in `_packets`. Either rename the C counter to `fwstate_sync` (`modules/fwstate/api/fwstate_cp.c`, `modules/fwstate/dataplane/config.h`) or strip the suffix here.

modules/fwstate/controlplane/metrics.go
```diff
@@ -0,0 +1,246 @@
...
+					makeCounter("fwstate_sync_v6_inserted_packets", packets, baseLabels...),
+				)
+			case "fwstate_sync_v4_insert_failed":
+				result = append(result,
+					makeCounter("fwstate_sync_v4_insert_failed_packets", packets, baseLabels...),
+				)
+			case "fwstate_sync_v6_insert_failed":
+				result = append(result,
+					makeCounter("fwstate_sync_v6_insert_failed_packets", packets, baseLabels...),
+				)
```
3Hren: `_inserted` and `_insert_failed` count state-table entries, not packets.

`fwstate_process_sync_v4` / `fwstate_process_sync_v6` do `inserted_cnt[0] += 1` once per invocation, and they are invoked from the per-frame loop in `modules/fwstate/dataplane/dataplane.c` (see the other comment). A single sync packet carries `udp_payload_len / sizeof(struct fw_state_sync_frame)` frames, so these four series count frames, not packets.

Exporting them as `..._packets` is therefore wrong, and `yanet-cli-fwstate metrics` compounds it by rendering them under a `Packets` column. Suggest naming them after what they measure (e.g. `fwstate_sync_v4_inserted_total`) and giving them their own CLI column.

modules/fwstate/controlplane/metrics.go
```diff
@@ -0,0 +1,246 @@
...
+				result = append(result,
+					makeCounter("fwstate_external_dropped_packets", packets, baseLabels...),
+					makeCounter("fwstate_external_dropped_bytes", bytes, baseLabels...),
+				)
+			case "fwstate_internal_forwarded":
+				result = append(result,
+					makeCounter("fwstate_internal_forwarded_packets", packets, baseLabels...),
+					makeCounter("fwstate_internal_forwarded_bytes", bytes, baseLabels...),
+				)
+			}
```
3Hren: The switch has no `default:` arm, so any counter whose name is not one of the eight listed above is silently dropped.

`cp_module_init` registers `rx`, `tx`, `rx_bytes` and `tx_bytes` on every module (`lib/controlplane/config/cp_module.c`), and `ModuleCounters` returns them here — none of them reach the metrics endpoint. Any counter added to `fwstate_module_config_new` later will be dropped the same way, with no build or test failure to catch it.

`modules/acl/controlplane/metrics.go` handles this with a `default:` arm that emits a generic pair labelled with the counter name.

modules/fwstate/dataplane/dataplane.c
```diff
@@ -331,7 +377,9 @@ fwstate_handle_packets(
 					sync_frame,
 					is_external,
 					now,
-					&fwstate_config->sync_config.timeouts
+					&fwstate_config->sync_config.timeouts,
+					sync_v4_inserted_cnt,
+					sync_v4_insert_failed_cnt
```
3Hren, line 382: Per-frame call site backing the naming comment on `modules/fwstate/controlplane/metrics.go`.

`fwstate_process_sync_v4` increments `inserted_cnt[0]` / `insert_failed_cnt[0]` once per call, and this loop runs once per `fw_state_sync_frame` carried by the packet. So `fwstate_sync_v4_inserted` is a frame counter, and exporting it as `fwstate_sync_v4_inserted_packets` misreports it whenever a packet carries more than one frame.

3Hren (commented):
Three related problems with the metric names exported by `collectDataplaneMetrics`. Details inline.

None of them are build- or test-visible: the C counters, the exported series and the CLI columns are three independent hardcoded lists, and nothing cross-checks them.

modules/fwstate/controlplane/service.go
```diff
@@ -77,20 +121,67 @@ type FWStateService struct {
...
-		log:         log,
+		log:         opts.Log,
+	}
+	if opts.MetricsOpts != nil {
+		metricsOpts := make([]grpcmetrics.Option, 0, len(opts.MetricsOpts)+2)
+		metricsOpts = append(metricsOpts, grpcmetrics.WithLabeler(labeler))
+		metricsOpts = append(metricsOpts, opts.MetricsOpts...)
+		metricsOpts = append(metricsOpts, grpcmetrics.WithRetention(m.retention))
+		m.metrics = grpcmetrics.New(metricsOpts...)
+	}
```
3Hren: Missing `grpcmetrics.WithServiceFilter`.

`controlplane/gateway/runner.go` builds one `grpc.Server` per module and chains every interceptor returned by `UnaryServerInterceptors()` onto it. The ACL module registers four services on that server, so this `ServerMetrics` observes `ACLService` and acl `MetricsService` calls too, and records them under `grpc_service="...aclpb.v1.ACLService"`.

Reproduce: issue any `yanet-cli-acl` call, then run `yanet-cli-fwstate metrics` — the gRPC section lists ACL methods.

`grpcmetrics.WithServiceFilter` (`common/go/grpcmetrics/options.go`) exists for exactly this, and the gateway already uses it. Suggest binding it next to the labeler:

```go
metricsOpts = append(metricsOpts, grpcmetrics.WithServiceFilter(
        func(service string) bool {
                return service == fwstateServiceName ||
                        service == fwstateMetricsServiceName
        },
))
```

Those two constants currently live in `modules/acl/controlplane/mod.go`; they belong in this package, ideally derived from `fwstatepb.FWStateService_ServiceDesc.ServiceName` and `fwstatepb.MetricsService_ServiceDesc.ServiceName` rather than hardcoded.

modules/acl/controlplane/mod.go
```diff
@@ -83,23 +92,31 @@ func (m *ACLModule) Endpoint() string {
...
+	var interceptors []grpc.UnaryServerInterceptor
+	if si := m.aclService.UnaryServerInterceptor(); si != nil {
+		interceptors = append(interceptors, si)
 	}
-
-	return []grpc.UnaryServerInterceptor{si}
+	if si := m.fwstateService.UnaryServerInterceptor(); si != nil {
+		interceptors = append(interceptors, si)
+	}
+	return interceptors
```
3Hren, line 117: Both metrics interceptors now run for every unary RPC on this server.

`RegisterService` above puts `ACLService`, acl `MetricsService`, `FWStateService` and fwstate `MetricsService` on one `grpc.Server`, and `grpc.ChainUnaryInterceptor` applies the slice returned here to all of them. Neither `ServerMetrics` is scoped to its own services, so each collector counts the other's calls, and every RPC now pays two full metric-recording interceptors.

The ACL side of this predates the PR (its `grpcmetrics.NewFactory` call also omits `WithServiceFilter`) and is tracked separately — but this change adds a symmetric second offender, so please scope the fwstate one here.

modules/fwstate/controlplane/metrics.go
```diff
@@ -0,0 +1,246 @@
...
+	for name := range m.configs {
+		configNames[name] = struct{}{}
+	}
+	m.mu.Unlock()
+
+	return func(id metrics.MetricID) bool {
+		config := id.Labels["config"]
+		if config == "" {
+			return true
+		}
```
3Hren, line 319: Foreign series are never pruned.

This predicate keeps any series without a `config` label. The fwstate `labeler` type-switches on `*fwstatepb.*Request` and returns `nil` for ACL requests, so the ACL series picked up per the other comment carry no `config` label and survive every `Collect` for the lifetime of the process.

A service filter fixes this at the source; the predicate itself is fine once only fwstate calls reach the collector.

3Hren (commented):
The new fwstate `ServerMetrics` has no service filter, so the fwstate metrics endpoint reports gRPC call metrics for the ACL services that share its `grpc.Server`, and never prunes them. Details inline.

modules/fwstate/controlplane/metrics.go
```diff
@@ -0,0 +1,246 @@
...
+//   - grpc_code:     gRPC status code string (grpc_server_handled_total only)
+func (m *FWStateService) Metrics() ([]*commonpb.Metric, error) {
+	// Collect map-stat gauges and dataplane counters under a single m.mu lock
+	// so both halves observe the same set of configs. The lock is released
+	// before collecting gRPC metrics: m.metrics.Collect() invokes the retention
+	// callback, which takes m.mu itself.
+	m.mu.Lock()
+	metrics := m.collectMapStats()
+	dpMetrics, err := m.collectDataplaneMetrics()
+	m.mu.Unlock()
```
3Hren: `m.mu` is taken without `defer Unlock`.

`collectDataplaneMetrics` crosses cgo (`m.agent.DPConfig()`, `ModuleCounters`). If anything on that path panics, the mutex stays held and every subsequent fwstate RPC that needs it blocks forever. `modules/acl/controlplane/metrics.go` guards the equivalent code with `m.mu.Lock(); defer m.mu.Unlock()` inside the helper.

The comment is also inaccurate: `collectDataplaneMetrics` never reads `m.configs` — only `collectMapStats` does. So the lock does not make "both halves observe the same set of configs"; it is only needed for the first call, and holding it across the second one blocks `UpdateConfig`/`DeleteConfig` for the duration of a full counter sweep across all workers.

Suggest locking inside a defer-guarded `collectMapStats` and dropping the second half out of the critical section.

Unrelated, same lines: the local `metrics` on line 76 shadows the imported package `common/go/metrics`, which this file uses in `retention()`.

modules/fwstate/controlplane/metrics.go
```diff
@@ -0,0 +1,246 @@
...
+		{Name: "af", Value: af},
+	}
+
+	return []*commonpb.Metric{
+		makeGauge("fwstate_index_size", float64(stats.IndexSize), labels...),
+		makeGauge("fwstate_extra_bucket_count", float64(stats.ExtraBucketCount), labels...),
+		makeGauge("fwstate_max_chain_length", float64(stats.MaxChainLength), labels...),
+		makeGauge("fwstate_layer_count", float64(stats.LayerCount), labels...),
+		makeGauge("fwstate_total_elements", float64(stats.TotalElements), labels...),
+		makeGauge("fwstate_max_deadline_ns", float64(stats.MaxDeadline), labels...),
```
3Hren: `fwstate_max_deadline_ns` is an absolute timestamp, not a duration.

`fwmap_max_deadline` returns `max(now + ttl)` over the per-worker counters, where `now` is the dataplane's monotonic clock. Exporting it under an `_ns` suffix makes `format_gauge_value` in `modules/fwstate/cli/src/main.rs` print it as a duration.

Either rename it so the CLI stops duration-formatting it (e.g. `fwstate_max_deadline_timestamp_ns`), or convert it to the remaining TTL, which is probably what an operator actually wants to see.

modules/fwstate/controlplane/service.go
```diff
@@ -11,10 +11,54 @@ import (
...
+
+// options holds the optional parameters for FWStateService construction.
+type options struct {
+	// MetricsOpts holds the extra grpcmetrics options (e.g. custom
+	// histogram buckets) to apply when collecting gRPC call metrics.
+	// A non-nil slice enables metrics collection; the labeler and
+	// retention predicate are bound by the service itself.
+	MetricsOpts []grpcmetrics.Option
+	Log         *zap.Logger
+}
```
3Hren, line 27: Two conventions here.

`MetricsOpts` abbreviates `Options` - use `MetricsOption`.

The field doc crams two separate ideas onto consecutive `//` lines. House style is a one-line brief, a blank `//` line, then the detail paragraph.

modules/fwstate/controlplane/service.go
```diff
@@ -11,10 +11,54 @@ import (
...
+// the service itself, so callers do not need to know about them. When this
+// option is unset, no gRPC call metrics are collected.
+func WithMetrics(opts ...grpcmetrics.Option) Option {
+	return func(o *options) {
+		// Always initialize to a non-nil slice so that a later
+		// MetricsOpts != nil check reliably detects that WithMetrics
+		// was called, even when no extra options were passed.
+		o.MetricsOpts = make([]grpcmetrics.Option, 0, len(opts))
+		o.MetricsOpts = append(o.MetricsOpts, opts...)
+	}
```
3Hren: This reimplements `grpcmetrics.Factory` with a nil-vs-empty-slice sentinel.

`grpcmetrics.NewFactory` exists for exactly this and is documented as the way to break the service<->collector cycle; `modules/acl/controlplane/service.go` already uses it (`options.Metrics grpcmetrics.Factory` + `m.metrics = opts.Metrics(m.retention)`). The comment inside this function describes the workaround, which is usually a sign the shape is wrong.

There is also a real asymmetry in `NewFWStateService`: the built-in `WithLabeler` is appended *before* `opts.MetricsOpts`, so a caller-supplied labeler overrides it, while the built-in `WithRetention` is appended *after*, so a caller-supplied retention is silently ignored. The doc comment claims both are "bound by the service itself".

Adopting `grpcmetrics.Factory` removes the sentinel and the asymmetry at once.

modules/fwstate/cli/src/main.rs
```diff
@@ -73,7 +82,14 @@ impl FWStateService {
         })
         .await?;
 
-        Ok(Self { service })
+        let metrics = Service::connect(connection, METRICS_SERVICE_NAME, |channel| {
+            MetricsServiceClient::new(channel)
+                .send_compressed(CompressionEncoding::Gzip)
+                .accept_compressed(CompressionEncoding::Gzip)
+        })
+        .await?;
```
3Hren: This second `Service::connect` opens a second connection.

Each `Service::connect` runs `Connection::connect`, i.e. a fresh `Channel::connect().await` plus `auth::create_layer(&args.auth).await`. Together with the one on line 78, every `yanet-cli-fwstate` invocation — `show`, `update`, `entries`, none of which touch metrics — now pays two TCP connections and two auth handshakes.

`cli/core/src/client.rs` documents the multi-service path: connect a `Connection` once, then `Service::new` per service. `modules/acl/cli/src/main.rs` does it that way.

```rust
let conn = Connection::connect(connection).await?;
let service = Service::new(&conn, SERVICE_NAME, |channel| { /* ... */ });
let metrics = Service::new(&conn, METRICS_SERVICE_NAME, |channel| { /* ... */ });
```

modules/fwstate/cli/src/main.rs
```diff
@@ -539,6 +593,315 @@ fn print_entry(entry: &fwstatepb::FwStateEntry) {
...
+            format!("{:.2} KB", value / 1024.0)
+        } else if value < 1024.0 * 1024.0 * 1024.0 {
+            format!("{:.2} MB", value / (1024.0 * 1024.0))
+        } else {
+            format!("{:.2} GB", value / (1024.0 * 1024.0 * 1024.0))
+        }
+    } else {
+        format_number(value as u64)
+    }
+}
```
3Hren, line 677: Unit issues in the gauge formatter.

The `_ns` branch assumes every `_ns` gauge is a duration. `fwstate_max_deadline_ns` is an absolute monotonic timestamp (see the other comment on `metrics.go`), so it renders as a nonsensical multi-day duration.

The `_bytes` branch divides by 1024 but labels the result `KB`/`MB`/`GB`. Those are the decimal prefixes; the binary ones are `KiB`/`MiB`/`GiB`.

modules/fwstate/tests/dataplane/fwstate.go
```diff
@@ -39,7 +41,27 @@ test_fwstate_handle_packets(
...
+	}
+
+	struct counter_storage *storage = counter_storage_spawn(
+		&cp_module->memory_context, NULL, registry
+	);
+	SET_OFFSET_OF(&module_ectx.counter_storage, storage);
+
 	fwstate_handle_packets(dp_worker, &module_ectx, packet_front);
+
+	counter_storage_free(storage);
```
3Hren, line 74: Error paths in the cgo test harness.

- On `counter_registry_link` failure this calls `fwstate_handle_packets` anyway, with `module_ectx.counter_storage` still zero. `ADDR_OF_NONNULL` on a zero offset yields the address of the field itself, so `counter_get_address` then reads and writes through a bogus stack pointer. The branch is currently unreachable — `counter_registry_link` with `src == NULL` always returns 0 — so this is a shape defect rather than a live bug, but the handler should not run after a setup failure.
- `err` is never freed on that path.
- `counter_storage_spawn` can return NULL on allocation failure; it is passed to `SET_OFFSET_OF` unchecked, which re-arms the same bogus-pointer path. (`counter_storage_free(NULL)` is safe, so the last line is fine.)
- The registry is re-linked and the storage re-spawned and freed on **every** call, so counter values cannot accumulate across invocations. Any future test asserting a counter value would silently read zero. Spawning once at config setup and freeing at teardown would make the counters observable — which is what the missing tests need.

modules/fwstate/tests/dataplane/fwstate.go
```diff
@@ -85,12 +107,23 @@ cp_module_init(
 	// Set agent offset
 	SET_OFFSET_OF(&cp_module->agent, agent);
 
+    // Initialize the counter registry. fwstate_module_config_new registers
+    // module-level counters right after cp_module_init, and
+    // counter_registry_register dereferences registry->memory_context (an
+    // offset pointer).
```
3Hren: This comment block is indented with spaces while the surrounding code uses tabs. clang-format does not reach inside cgo preambles, so it has to be fixed by hand.

modules/acl/controlplane/mod.go
```diff
@@ -14,21 +14,23 @@ import (
 )
 
 const (
-	agentName          = "acl"
-	serviceName        = "modules.acl.controlplane.aclpb.v1.ACLService"
-	fwstateServiceName = "modules.fwstate.controlplane.fwstatepb.v1.FWStateService"
+	agentName                 = "acl"
+	serviceName               = "modules.acl.controlplane.aclpb.v1.ACLService"
+	fwstateServiceName        = "modules.fwstate.controlplane.fwstatepb.v1.FWStateService"
+	fwstateMetricsServiceName = "modules.fwstate.controlplane.fwstatepb.v1.MetricsService"
```
3Hren: `fwstateMetricsServiceName` is hardcoded.

`ServicesNames()` below already derives acl's own from `aclpb.MetricsService_ServiceDesc.ServiceName`, and `modules/forward/controlplane/mod.go` does the same. A hand-written FQN drifts silently if the proto package ever moves.

Both fwstate names would also be better placed in the fwstate package than in acl's — the fwstate service currently does not know its own service name, which is why the service filter suggested in the other review has nothing to reference.

3Hren (commented):
A batch of non-blocking findings, separate from the two correctness reviews above.

One that has no line to anchor to: **the eight new counters and the whole of `metrics.go` have no test coverage.** `service_test.go` exists but only covers pure helpers (`TestClampBatchSize`, `TestValidateSyncPorts`, `TestValidateSyncConfigDstEther`); `modules/fwstate/tests/dataplane/fwstate_test.go` runs packets through the handler but asserts only Output/Drop, never counter values. `labeler` is a pure function and `retention` needs nothing but a populated `configs` map, so both are cheap to cover. For comparison, `f65863a2` (`feat(forward): per-rule metrics`) extended `service_test.go` in the same commit.

modules/fwstate/controlplane/metrics.go
saushew: Fixed at https://github.com/yanet-platform/yanet2/pull/1377/changes/69ea4656c6fa7a5b3eea0163a7405906e6a1bcca

modules/fwstate/controlplane/metrics.go
saushew: Fixed at https://github.com/yanet-platform/yanet2/pull/1377/changes/583c18f3e70e3dc936b68b59b230213a4d8ec246

modules/fwstate/dataplane/dataplane.c
saushew: The increment counter behavior is fixed by renaming here https://github.com/yanet-platform/yanet2/pull/1377/changes/583c18f3e70e3dc936b68b59b230213a4d8ec246

modules/fwstate/controlplane/service.go
saushew: Fixed at https://github.com/yanet-platform/yanet2/pull/1377/changes/15f188931db6507ae0b9d3ed851d2a7101b00c6e

modules/fwstate/controlplane/metrics.go
saushew: Fixed by `grpcmetrics.WithServiceFilter` here https://github.com/yanet-platform/yanet2/pull/1377/changes/15f188931db6507ae0b9d3ed851d2a7101b00c6e

modules/acl/controlplane/mod.go
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/93f130aed84c3cd98620c3a8a8e4508532ed1f85

modules/fwstate/controlplane/metrics.go
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/0ea4e3162c1ea48beb51dd0ec886e31effdfb5e9

modules/fwstate/controlplane/service.go
saushew: This is outdated since the comment was deleted here https://github.com/yanet-platform/yanet2/pull/1377/changes/93f130aed84c3cd98620c3a8a8e4508532ed1f85

modules/fwstate/controlplane/metrics.go
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/db947293324f1ba088ead07a8ab848c7e4def283

modules/fwstate/controlplane/service.go
saushew: Fixed here https://github.com/yanet-platform/yanet2/commit/93f130aed84c3cd98620c3a8a8e4508532ed1f85

modules/fwstate/cli/src/main.rs
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/966bd8e3ab1cfd5e572ff29a13a65159a56885f6

modules/fwstate/cli/src/main.rs
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/5b5bc861af21d49d7889d4c9a44dfc67e4536b2c

modules/fwstate/tests/dataplane/fwstate.go
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/b4b2e4d7c2d379514d15602d04bf32a341d1642b

modules/acl/controlplane/mod.go
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/15f188931db6507ae0b9d3ed851d2a7101b00c6e

modules/fwstate/tests/dataplane/fwstate.go
saushew: Partially fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/4f974c062c52caeec845eacfc6fad57a7feb4566, but I'd like to extract CGo preamble to it's own `harness.h`/`harness.c` files to make implementation more clear. This approach will probably make other issues more visible, and they will be fixed in future changes.

modules/fwstate/controlplane/metrics.go
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/7ec3be1f51ecaef5160b96b318325da19af86808

modules/fwstate/controlplane/metrics.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+	case "rx_bytes":
+		return []*commonpb.Metric{
+			makeCounter("fwstate_rx_bytes", packets, baseLabels...),
+		}
+	case "tx_bytes":
+		return []*commonpb.Metric{
+			makeCounter("fwstate_tx_bytes", packets, baseLabels...),
+		}
+	default:
```
3Hren, line 263: This block was written against the pre-45d5026b counter scheme and no longer matches main.

Since 45d5026b (merged 2026-07-11), `cp_module_init` registers `rx` and `tx` as size-2 vectors (`[packets, bytes]`), and `rx_bytes`/`tx_bytes` no longer exist. It also registers `drop`, `pending_input`, `pending_output` (size 2) and `hist_0`..`hist_5` (histogram vectors, written on every batch in `lib/dataplane/pipeline/pipeline.c`).

After merge:
- `case "rx"` / `case "tx"` export only `workerVals[0]` — the byte halves are silently dropped;
- the `rx_bytes` / `tx_bytes` arms are dead code;
- the "single-valued" comment above is no longer true;
- `drop` falls into the default arm, though for fwstate it is meaningful (external sync packets are dropped) and deserves a dedicated pair like the others;
- `hist_0..5` reach the default arm and are exported as garbage — bucket[0]/bucket[1] sums labelled as packets/bytes. They need an explicit skip; exporting them as proper histograms is a separate task.

None of this is test-visible: `metrics_test.go` feeds synthetic `CounterInfo`, and CI builds the merge ref but nothing exercises the real counter set. Needs a rebase onto current main and a rework of this switch against the new scheme.

modules/fwstate/cli/src/main.rs
```diff
@@ -539,6 +591,333 @@ fn print_entry(entry: &fwstatepb::FwStateEntry) {
...
+                    m.label_value("config").unwrap_or(""),
+                    m.label_value("device").unwrap_or(""),
+                    m.label_value("pipeline").unwrap_or(""),
+                    m.label_value("function").unwrap_or(""),
+                    m.label_value("chain").unwrap_or(""),
+                );
+                if !counter_map.contains_key(&key) {
+                    counter_keys.push(key.clone());
+                }
+                counter_map.entry(key).or_default().push(m);
```
3Hren, line 743: The grouping key omits the `counter` label, so all generic series collapse into one row.

Every counter from the default arm of `emitCounterMetrics` is exported under the same pair of names (`fwstate_counter_packets` / `fwstate_counter_bytes`) and is distinguished only by the `counter` label. Here they land in the same position group, then `strip_suffix("_packets")` yields the same base `"counter"` for all of them, and `pair.packets = Some(val)` below overwrites — last write wins.

On current main this is a live case, not a hypothetical: after the counters rework every position carries `drop`, `pending_input`, `pending_output` and `hist_0..5`, so the human output shows a single meaningless "Counter" row. Include the `counter` label value in the pair key (JSON output is unaffected — labels survive).

modules/fwstate/cli/src/metric.rs
```diff
@@ -0,0 +1,119 @@
+use commonpb::pb;
```
3Hren, line 1: This file is a near-verbatim copy of `modules/acl/cli/src/metric.rs` — the only differences are import paths and the cumulative-percentile fix. The table renderer in `main.rs` clones acl's as well: `CounterRow`/`GaugeRow`/`GrpcCallRow`/`GrpcLatRow`, `format_number`, `format_gauge_value`, `metric_display_name`, `print_metrics_table`.

The divergence has already started: the cumulative-percentile fix and the KiB fix exist only in this copy, so `yanet-cli-acl` computes p50/p95/p99 incorrectly right now. That is exactly why the renderer must not be cloned per module — move `Metric`/`Kind`/`from_proto`/`histogram_percentile` (ideally the table helpers too) into a shared crate under `common/rust`/`ync` and point both CLIs at it.

If the extraction is deferred to a follow-up, the percentile and KiB fixes must be ported to the acl CLI immediately.

3Hren (commented):
Re-reviewed the current head. All findings from the previous rounds check out as fixed — including the TTL export: the dataplane clock is TSC-derived wall time (`lib/dataplane/time/clock.h`), so the comparison against `UnixNano` is valid. Three blocking items remain, all inline: the rx/tx handling was written against the pre-45d5026b counter scheme, the CLI collapses generic counter series into one row, and the metrics CLI code is a diverging clone of the acl CLI.

modules/fwstate/controlplane/metrics.go
saushew: Rebased and fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/10b2fe8b03e5a5e744ae15800c292f016ea96a70

modules/fwstate/cli/src/main.rs
saushew: Fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/10b2fe8b03e5a5e744ae15800c292f016ea96a70

modules/fwstate/cli/src/metric.rs
saushew: ACL datasize dimentions are fixed here https://github.com/yanet-platform/yanet2/pull/1377/changes/0c8dc7fc0b741686fb953b00481913f6e26335ac. Shared crate will be done in further changes.

3Hren (approved):

PR #1381, opened 2026-07-07
feat(route): track FIB entry count in croute bindings

egnees:
Adds FIBRangeCount() to croute.ModuleConfig, reusing the same FIB iterator DumpFIB walks internally but skipping nexthop resolution and entry allocation.

This gives a count that matches len(DumpFIB()) exactly, without materializing the full FIB. An earlier version of this PR tracked a separately-incremented counter of successful AddPrefix calls, but that diverges from the DumpFIB range count whenever prefixes overlap or share a nexthop set, so it was replaced with this iterator-based approach.

#1379 currently computes its route_fib_entries metric via len(DumpFIB()), which materializes the whole FIB under a read lock on every scrape. Once this merges, that PR will switch to FIBRangeCount() instead.

### DIFF
--- a/modules/route/bindings/go/croute/safe.go
+++ b/modules/route/bindings/go/croute/safe.go
@@ -120,3 +120,20 @@ func (m *ModuleConfig) DumpFIB() ([]FIBEntry, error) {
 
 	return entries, nil
 }
+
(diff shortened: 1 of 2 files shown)


egnees:
@codex review

sakateka (approved):

PR #1385, opened 2026-07-07
feat(dataplane): export module constructors from .so plugins

3Hren:
The dataplane loads packet-processing modules as external .so plugins, whose dataplane sources are compiled with hidden visibility. Add the export attribute that marks a module constructor for default visibility so the plugin loader can resolve it from the loaded object.

This completes the export side of the module plugin ABI, following the .so ABI-version guard.


3Hren:
@codex review

PR #1386, opened 2026-07-07
feat(dataplane_ut): load modules from .so plugins at runtime

3Hren:
The in-process dataplane test harness could only run statically-linked modules. Thread the shared-memory plugin loader through the harness so a test can point it at a directory of module .so plugins and have them resolved at runtime, taking priority over the statically-linked built-ins.

An empty plugin directory keeps the previous behaviour: only the statically-linked built-in modules are resolved, so existing harness tests are unaffected.


3Hren:
@codex review

PR #1389, opened 2026-07-08
chore(cargo): discover module CLI crates via a glob workspace member

3Hren:
Module CLI crates were listed one by one in the workspace, so an optional or out-of-tree module's CLI needed a Cargo.toml edit to be built. Match them with a glob member instead, so a module's CLI crate is picked up by a workspace build automatically — the Cargo counterpart to the meson extra_modules option.

The route CLI crate is nested, so its parent directory is excluded to keep the glob from treating that non-crate directory as a member. The resolved workspace member set and Cargo.lock are unchanged for the public checkout.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,18 +17,8 @@ members = [
     "cli/modules/ready",
     "cli/modules/metrics",
     "cli/modules/device",
-    "modules/acl/cli",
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #1393, opened 2026-07-09
fix(fwstate): validate sync ports before uint16 conversion

### LINKED ISSUE #899: fix(fwstate): validate sync ports before uint16 conversion
FWState sync ports are defined as `uint32` in protobuf but are silently truncated to `uint16` during conversion to C config.

Affected paths:
- `modules/fwstate/controlplane/converters.go`
- `modules/fwstate/controlplane/service.go`
- `modules/fwstate/controlplane/fwstatepb/fwstate.proto`

Problem:
- `SyncConfig.port_multicast` and `port_unicast` are `uint32` in proto.
- `ConvertPbToCSyncConfig()` casts them to `uint16` before host-to-network conversion.
- Validation does not reject values greater than 65535.
- A request with port `70000` is accepted and used as a different wire port.

Expected outcome:
- Validate ports are in the TCP/UDP range before conversion.
- Return `InvalidArgument` for out-of-range values.
- Add converter or service tests covering the boundary values.

3Hren:
Sync configuration ports are declared as 32-bit values but were silently truncated to 16 bits when converting to the dataplane config, so a caller could set a port such as 70000 and have it accepted as a different wire port.

This adds validation that rejects out-of-range ports with `InvalidArgument` before any conversion, keeping the accepted set aligned with the valid TCP/UDP range.

Closes #899.

### DIFF
--- a/modules/fwstate/controlplane/service.go
+++ b/modules/fwstate/controlplane/service.go
@@ -24,6 +24,10 @@ const (
 	// ListEntries round-trip to prevent unbounded allocation under the
 	// service mutex.
 	maxListEntriesBatchSize uint32 = 10000
+
+	// maxSyncPort is the highest value accepted for port_multicast and
+	// port_unicast, matching the width of the C-side uint16 port field.
+	maxSyncPort uint32 = 65535
 )
 
 // clampBatchSize returns a batch size that is within the allowed range:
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1394, opened 2026-07-09
chore(gitignore): ignore private modules by default, whitelist public ones

3Hren:
Ignore everything under the modules directory by default and re-include the tracked public modules explicitly.

This lets a new out-of-tree private module be added without touching this file, while keeping the set of public modules an explicit, reviewable whitelist.


3Hren:
@codex review

PR #1395, opened 2026-07-09
refactor(metrics): unify GetMetrics messages on the shared commonpb pair

3Hren:
Every module and operator that exposes a metrics RPC used to redefine its own identical `GetMetricsRequest`/`GetMetricsResponse` pair, while an identical canonical pair sat unused in commonpb. This points all of them at the shared commonpb messages and deletes the per-package copies.

The metrics service itself stays defined per package, because the gateway routes by fully-qualified service name to each backend — only the message types are unified. The generic metrics probe already spoke the commonpb types, so its contract is unchanged.

The change is wire-compatible: the request stays empty and the response still carries the same repeated metric list. Because the referenced message type names change, the Buf breaking check reports the nominal rename for the affected packages even though no bytes on the wire change.


3Hren:
@codex review

PR #1400, opened 2026-07-09
fix(functional): stabilize VM snapshot fallback in CI

moonug:
## Summary

- rebuild the booted VM snapshot on each runner instead of restoring non-portable live QEMU state from cache
- select guest paths from the snapshot actually chosen by the VM pool, preserving correct local and 9P paths during fallback
- use the selected guest paths for baseline configuration files and add regression coverage for baseline-to-booted fallback

## Testing

- `go test ./framework`
- 40 functional CI runs: 39 passed; one run failed with a separate `TestForward/Test_ICMP6_Echo` packet timeout


moonug:
## A/B result

The resource-exhaustion hypothesis is not supported by these runs.

- `pool=1` passed the full functional suite. Peak observed QEMU RSS was ~654 MiB, peak QEMU CPU ~148%, minimum `MemAvailable` ~13.7 GiB, and peak load1 1.11.
- `pool=5` passed on rerun. Peak observed aggregate QEMU RSS was ~3.23 GiB, peak QEMU CPU ~349%, minimum `MemAvailable` ~11.6 GiB, peak load1 5.04, and swap remained unused.
- The first `pool=5` attempt failed before the five-VM pool was created: the single `baseline-prep-main` VM panicked in guest kernel 6.8.0-124 at `kvm_kick_cpu+0x61/0x90`. At that point host load was below 1 and about 14.2 GiB memory remained available.

This isolates the observed `mount -a` timeout as a consequence of a guest nested-KVM kernel panic while restoring the cached `booted` snapshot, rather than host CPU/RAM exhaustion caused by pool size. Pool size can increase load, but the successful pool=5 run had substantial memory headroom and no swap/OOM evidence.

Run: https://github.com/yanet-platform/yanet2/actions/runs/29085376798

3Hren (approved):

PR #1401, opened 2026-07-10
fix(dataplane): drop output from input entry point processing

GeorgyKirichenko:
The input entry point must not transmit packets directly — it may only classify and re-route them via pending queues. Enforce this by dropping the output list after device_ectx_process_input, leaving pending_input and pending_output as the only paths to survival.

Update all functional test harnesses to append a catch-all forward sink with ModeOut to each chain, so pass-through packets reach the output stage via pending_output.


GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

PR #1403, opened 2026-07-11
feat(web): support multiple UI instances with different backends

GeorgyKirichenko:
Add runtime config.json mechanism so each deployment can pre-configure its default backend URL and seed additional gateways. The SPA fetches /config.json on startup; if missing it falls back to same-origin.

Replace VITE_BACKEND_PORT with VITE_BACKEND_URL (full URL) and add VITE_DEV_PORT so multiple dev servers can run side by side, each proxying to a different backend.

Extract deriveBaseUrl from GatewayDrawer into its own module so runtimeConfig and the drawer share one implementation. Wire GatewayProvider to accept a runtimeConfig prop that overrides the seed gateways and builtin gateway base URL.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -17,7 +17,23 @@ Run the development server (from the repo root, or `cd web` and drop the flag):
 npm run dev -w web
 ```
 
-The development server will run on `http://localhost:3000` and proxy API requests to the backend.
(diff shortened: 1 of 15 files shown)


GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

PR #1409, opened 2026-07-13
fix(bird-adapter): reject unknown config keys and align the default endpoint

3Hren:
A renamed configuration key (`gateway_endpoint` → `route_operator_endpoint`) was silently dropped on load, because config parsing ignored unknown YAML keys. The endpoint therefore kept its built-in default, which disagreed with the shipped configuration file and happened to match the adapter's own listen address on a deployed host. The adapter opened its RIB stream to itself, and BIRD route updates never reached the route operator — with no error at startup.

- Add opt-in strict YAML decoding to the shared config loader, so an unknown or renamed key becomes a startup error instead of a silent fallback to a default.
- Enable strict decoding for the BIRD adapter only, leaving every other component on the existing lenient behaviour so that hosts carrying older configuration files keep starting.
- Align the built-in defaults with the shipped configuration file and point both at the gateway address the other operators already use. The previous default targeted a port nothing listens on.
- Pin the shipped configuration file to the built-in defaults with a test, so the two can no longer drift apart unnoticed.

### DIFF
--- a/common/go/xcfg/load.go
+++ b/common/go/xcfg/load.go
@@ -1,7 +1,10 @@
 package xcfg
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"reflect"
 	"strings"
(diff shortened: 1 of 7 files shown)


3Hren:
@codex review

sakateka (approved):

PR #1416, opened 2026-07-14
refactor(cli): table-free ready output

3Hren:
The readiness table jumped in width between runs and refreshes, which made it unusable for eyeballing readiness or watching a rollout.

Column widths were derived from content, and the reason column held unbounded free text — a single long apply error forced every other column to reflow. The streaming mode re-rendered a whole table for each stream message, even though the server sends only the scopes that changed after the first snapshot, so every frame had a different shape.

The table is replaced with a status list. The fields that precede the age are fixed-width, the reason text moves onto indented continuation lines instead of being a column, and the only content-derived width is computed once and held. The streaming mode now prints one snapshot followed by an append-only transition log.

Ages are rendered through the duration formatter the crate already depends on, rather than a hand-rolled one.

A scope is reported as stale only when it is demonstrably heartbeated, so a scope that is set once and never re-observed is never flagged. Readiness sources heartbeat on their own differing cadences, so the threshold is deliberately generous and can be tuned or disabled.

Machine-readable output is unchanged, as are the exit codes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3566,11 +3566,12 @@ dependencies = [
 name = "yanet-cli-ready"
 version = "0.1.0"
 dependencies = [
+ "chrono",
  "clap",
  "clap_complete",
  "colored",
+ "humantime",
  "prost-types",
- "tabled",
  "tokio",
  "yanet-cli",
  "yanet-readinesspb",
--- a/cli/core/src/display.rs
+++ b/cli/core/src/display.rs
@@ -40,6 +40,14 @@ pub fn fit_terminal_width(table: &mut Table) {
     }
 }
 
+/// Returns the current terminal width in columns, detected from stdout.
+///
+/// Returns `None` when stdout is not a TTY (piped or redirected), matching
+/// the same detection [`fit_terminal_width`] uses.
+pub fn terminal_width() -> Option<usize> {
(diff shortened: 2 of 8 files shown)


3Hren:
Thanks — addressed both findings.

**Watch drops later reasons (P2, watch.rs)** — fixed. The transition log now formats and emits every entry of `Scope.reasons`, reusing the same helper as the snapshot renderer, so the two can no longer disagree. (Today's tracker stores at most one reason per scope, so this was unreachable in practice, but the asymmetry between the two renderers was real.)

**Cap overlong scope names (P2, layout.rs)** — not taking this one, deliberately. The clamp is a padding bound, not a truncation, and that is the intended behaviour: a name longer than the bound overflows **its own row only** and shifts nothing on any other row, because the name column precedes every other fixed field. Truncating instead would trade a one-row ragged edge for a permanently unidentifiable scope, which is the worse failure for a diagnostic tool — the scope name is the primary key an operator acts on.

The bound is also generous relative to reality: scope names are `fib:<gateway>:<module>`, `pipeline:<gateway>`, `config:<gateway>`, `neighbours`, `rib`, `bird-session`, `gateway`. On a live stand the longest is 16 characters. Reaching the bound would take a ~30-character gateway name, and even then the cost is one ragged row, not a reflow.

Also worth noting for the fixed-geometry framing: the age field is no longer fixed-width — it is the last field on the row, so its width shifts nothing to its left.

PR #1419, opened 2026-07-14
fix(gateway): stop double-logging service registration

3Hren:
The gateway reported every service registration twice at startup: once under the short module name and once under the gRPC service name, using the same message text at two different log levels.

Registration is now reported exactly once per gRPC service, with the backend kind carried in a dedicated field.


3Hren:
@codex review

PR #1420, opened 2026-07-14
fix(build): install and package the mirror CLI

3Hren:
The mirror module ships a CLI crate, and the workspace build already produced its binary, but it was registered in neither the install list nor the Debian package — so the tool was built and then silently dropped on the floor.

- Add mirror to the module CLI list, so the install step picks up its binary.
- Ship the binary in the CLI Debian package alongside the other module CLIs.

Both entries have to land together: once the binary reaches the staging tree, packaging fails unless it is also claimed by the package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ CLI_MODULES := \
 	device-trafgen \
 	dscp \
 	fwstate \
+	mirror \
 	route \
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1422, opened 2026-07-14
feat(cli): colorize ready watch transitions

3Hren:
The streaming readiness log printed every transition in plain text, so the one thing worth spotting in a stream — a scope leaving or entering the ready state — carried no visual signal, even though the snapshot block above it was already colored.

Each transition line now opens with the state mark and paints the old and the new state in their own colors, keeping the timestamp, the arrow and the reason text muted.

The state-to-color mapping moves to a single place, so the streaming log and the snapshot block can no longer disagree on how a state looks.

Colors and glyphs still degrade together to plain ASCII when the output is not a terminal, and the reason text still wraps against the visible width of the line rather than its escape-laden length.


3Hren:
@codex review

PR #1424, opened 2026-07-14
fix(gateway): unify registration logging on the register path

3Hren:
Follow-up to #1419, which unified how the gateway reports the services it hosts itself.

Backends that register over the register RPC were still reported under a different field name and without the backend kind. All registration paths now share one message shape, so a single message covers built-in, in-process and out-of-process services alike.


3Hren:
@codex review

PR #1425, opened 2026-07-14
feat(lint): forbid the logger as a constructor or method argument

3Hren:
Adds a stdlib-only Go AST linter that rejects constructors and methods taking a zap logger as a parameter, so the options-pattern rule in CLAUDE.md is enforced mechanically instead of by prose.

Free functions stay out of scope by design, which keeps the WithLog options, the gRPC interceptor factories and the operator bootstrap legal.

Ships an allowlist that ledgers the 34 violations existing today. It is a one-time debt seed, not an escape hatch: a stale row fails the run, so the ledger only shrinks as the debt is paid. Migrating those constructors is deliberately left to follow-up changes.

Wires the linter into a tracked pre-commit hook (installed once via make hooks), a make lint-go target and a CI workflow.


3Hren:
@codex review

PR #1435, opened 2026-07-15
refactor(filter): drop reserved leading-underscore local names

3Hren:
- The filter query classifier used leading double-underscore local names, a leftover from when the routine was a macro.
- Leading double underscores are reserved for the implementation in C, so this renames them to plain names with no behavior change.
- It also silences the corresponding clang-tidy readability-identifier-naming findings.


3Hren:
@codex review

PR #1439, opened 2026-07-15
feat(dataplane_ut): support per-worker devices in the test harness

3Hren:
The in-process dataplane test harness pinned every worker to a single device and every injected packet to that same device, so no module could regression-test multi-device origination or any ownership rule that keys on the worker's own port.

This adds two opt-in capabilities to the harness and its Go bindings.

- Per-worker device and queue assignment, so a test can place workers on distinct devices.
- Packet injection with a caller-chosen ingress device, matching how the real dataplane stamps a packet at ingress.

Existing callers are unaffected: when the new options are left unset, the harness behaves exactly as before, and the full set of module test suites that depend on it stays green.

A new acceptance test drives a device-scoped forward rule to prove both capabilities are honoured and fail if either is ignored.

### DIFF
--- a/bindings/go/dataplane_ut/dataplane_ut.go
+++ b/bindings/go/dataplane_ut/dataplane_ut.go
@@ -89,6 +89,27 @@ free_cptr_array(const char **arr) {
 	free(arr);
 }
 
+// alloc_worker_spec_array copies n entries from src (Go-backed) into a
+// fresh C-heap array and returns it. The caller frees with
+// free_worker_spec_array.
+static struct dataplane_ut_worker_spec *
+alloc_worker_spec_array(const struct dataplane_ut_worker_spec *src, size_t n) {
+	struct dataplane_ut_worker_spec *dst =
+		(struct dataplane_ut_worker_spec *)malloc(
+			n * sizeof(struct dataplane_ut_worker_spec)
+		);
+	if (dst == NULL) {
+		return NULL;
+	}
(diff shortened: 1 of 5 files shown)


3Hren:
Addressed: HandleSegmentedPacketsOnDevice now rejects an rxDeviceID beyond the registered device count, mirroring the existing Bench guard, with a test for the rejection. @codex review

3Hren:
Addressed: the segmented on-device path now also rejects an out-of-range worker index (shared guard in handleSegmentedPackets), with a test. @codex review

3Hren:
Addressed: NewHarness now validates each WorkerSpec.DeviceID against the device count. Audited the full new surface — rxDeviceID, worker index, and WorkerSpec.DeviceID are all guarded; QueueID is a queue-selection argument, not a bounded array index, so it needs none. @codex review

3Hren:
Addressed at the C boundary: dataplane_ut_new now rejects an out-of-range worker device id and dataplane_ut_run guards its worker index, so direct C callers are protected too, with smoke-test coverage. Full C-surface audit found no other unguarded index of this class (queue_id is a PMD queue selector, not a bounded index). @codex review

3Hren:
Addressed: the invalid-worker guard in dataplane_ut_run now yields a well-formed result (drains input into result.drop, initializes both lists) so C callers never touch uninitialized memory, verified under ASan/UBSan; and the parsed Go path now validates the worker index like the segmented path, so HandlePacketsOnWorker returns an error instead of silently dropping. @codex review

PR #1441, opened 2026-07-15
refactor(pdump): reference the named mode constants in the CLI

3Hren:
- Wire the named packet-capture mode constants into the CLI conversions instead of the raw generated bindings, so they are no longer unused.
- This drops three dead-code lint suppressions with no behavior change.


3Hren:
@codex review

sakateka (approved):

PR #1446, opened 2026-07-15
feat(filter): reject non-contiguous IPv4 network masks and non-bi-contiguous IPv6 masks

3Hren:
- Validate that IPv4 network masks are a contiguous prefix and IPv6 masks are bi-contiguous, so malformed masks are rejected instead of being silently mis-classified by the packet filter.
- Reject such masks at the control-plane boundary with an invalid-argument error so clients get a clear response, while the filter engine keeps the same check as a defense-in-depth safety net.
- Cover the behaviour with Go unit tests for the validator and an ACL service test for the rejection path.

### DIFF
--- a/bindings/go/filter/types.go
+++ b/bindings/go/filter/types.go
@@ -1,6 +1,7 @@
 package filter
 
 import (
+	"encoding/binary"
 	"net"
 	"net/netip"
 )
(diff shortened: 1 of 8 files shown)


3Hren:
@codex review

PR #1447, opened 2026-07-15
chore(meson): add a meson lint CI check and fix its findings

3Hren:
- Add a CI check that lints meson files with mesonlint, triggered only when meson files change, gating on real lint diagnostics rather than the tool's formatting default.
- Remove two unused program-lookup assignments in the root build definition that the linter flags, keeping the existence and version checks intact.


3Hren:
@codex review

PR #1450, opened 2026-07-15
chore(build-deb): trigger the package build only on build-affecting changes

3Hren:
The Debian-package workflow was the only CI workflow using an ignore-list to decide when to run, and that list did not exclude documentation or agent-prompt files, so a markdown-only change would start a full package build.

Switch its pull-request trigger to an allow-list of the paths the package build actually depends on — source, build system, and packaging metadata — mirroring the pattern every other workflow already uses. A docs-only, web-only, or agent-prompt change no longer triggers it.

Release builds on a version tag remain unconditional, and the workflow now self-triggers on changes to its own file.

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -3,13 +3,6 @@ name: Build DEB Packages
 on:
   push:
     tags: ["v*"]
-    paths-ignore:
-      - "web/**"
-      - "modules/**/web/**"
-      - "devices/**/web/**"
-      - "operators/**/web/**"
-      - "package.json"
-      - "package-lock.json"
   pull_request:
     branches: ["**"]
     paths-ignore:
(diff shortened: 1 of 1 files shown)


3Hren:
Addressed: reverted to a deny-list — a package build has too many heterogeneous inputs (regenerated protos, dh_install configs, the DPDK submodule) to allow-list safely, so the trigger now excludes only provably non-build paths (web, npm lockfiles, **.md, .claude/**, docs/**) and lets every real package input through. Release-tag builds stay unconditional. @codex review

3Hren:
@codex review

3Hren:
Re: the P3 allow-list suggestion — keeping the deny-list is deliberate, and it directly resolves the earlier P2 on this same PR. A .deb build pulls from too many heterogeneous inputs to allow-list safely (regenerated protos, dh_install configs copied from the source tree, the DPDK submodule), and an allow-list that misses one ships or breaks a package with no build at all. The deny-list's failure mode is the safe one: an unlisted non-build file (another workflow YAML, a git hook, a deploy Dockerfile) triggers an unnecessary build — wasted CI, never a missed one. The list excludes the high-frequency non-build noise this PR targets (web, docs, markdown, agent prompts); the rare residual over-triggers are the accepted safe side of the tradeoff.

PR #1451, opened 2026-07-15
refactor(ready): align reason lines under the scope name

3Hren:
Pull the reason continuation lines in the readiness status block two columns left so they sit under the scope-name column. The dimmed styling already marks them as continuation text, so the deeper indent was unnecessary.


3Hren:
@codex review

3Hren:
Addressed in the latest push: the reason indent now tracks the mark's display width plus its separating space (4 in colored mode, 5 in ASCII), so continuation lines align under the scope-name column in both render modes. Colored output is unchanged.

3Hren:
@codex review

PR #1454, opened 2026-07-15
fix(acl): use atomic snapshots for acl metrics collection

semenovad:
As a temporary solution: ACL metadata is now read from an immutable snapshot without acquiring ACLService.mu, preventing long-running config updates from blocking metrics collection. Dynamic counters are still collected directly from the dataplane. A more fundamental lock-scope refactoring will follow separately


saushew:
LGTM

GeorgyKirichenko (approved):

PR #1456, opened 2026-07-16
refactor(cli): extract shared operator readiness reporting

3Hren:
The decap, forward, pipeline and route operator CLIs each carried a nearly identical copy of the readiness table rendering and summary logic. In the three smaller operator CLIs roughly 97 percent of the file was duplicated code, differing only in the operator name, the proto path and the service constant.

- Moved the shared readiness reporting into the CLI core library, so each operator CLI now keeps only its own command definitions and RPC wiring.
- Reduced the decap and forward CLIs from 265 to 117 lines each, and pipeline from 292 to 144.
- Carved the readiness parts out of the route CLI without touching its route specific commands or tests.
- Dropped dependencies that became unused in the three operator crates.

No behavior change: the rendered output, the exit code semantics and the JSON payload are all unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3102,6 +3102,7 @@ dependencies = [
  "tonic",
  "tower",
  "uuid",
+ "yanet-readinesspb",
  "yanet-ynpb",
 ]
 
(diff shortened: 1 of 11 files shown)


3Hren:
@codex review

PR #1458, opened 2026-07-16
refactor(cli): extract shared metrics rendering and fix acl percentiles

3Hren:
The acl and fwstate CLIs each carried a copy pasted metrics layer: the domain types, the number and gauge formatters, the shared gRPC call and latency sections, and the percentile helper.

- Moved the shared metrics rendering into the CLI core library, leaving each CLI only its module specific counter grouping, which genuinely differs between the two.
- Removed both local copies entirely, cutting the acl CLI from 648 to 485 lines and the fwstate CLI from 954 to 793.
- Renamed a closure binding in the percentile helper that shadowed the function's own percentile parameter, since the code now lives in the core library that every CLI links.

This is not a pure refactor. The two copies had silently diverged, and unifying them fixes the consequence: shared histogram values carry raw per bucket counts rather than cumulative ones, which is a deliberate divergence from Prometheus semantics documented at the producer. The fwstate copy accumulated them correctly while the acl copy compared a single bucket against the target as if the counts were already cumulative. The gRPC latency percentiles reported by the acl CLI were therefore wrong, and this changes what it prints for them.


3Hren:
@codex review

PR #1459, opened 2026-07-16
fix(pdump): correct the ICMPv6 target address slice

3Hren:
The neighbour solicitation and advertisement printers copied a fifteen byte slice of the payload into a sixteen byte address buffer. That combination panics rather than misprinting, so the pretty dump format aborted on ordinary neighbour discovery traffic.

- Corrected the slice in both the solicitation and the advertisement printer, since the defect was copy pasted into both.
- Added a test per printer, each asserting the rendered target address; both fail on the previous code with the reported panic.

The reachability explains why this went unnoticed: the default dump format is the concise one, which reports these packets by protocol name only and never parses the target address, and the capture formats do not print at all. Only an explicit request for the pretty format reaches the defect.

Verified against the packet definitions of the parsing library rather than the specification prose alone: the payload begins after the four byte header, and in both message types a four byte field precedes the sixteen byte target address, so the corrected offsets are right for both.


3Hren:
@codex review

PR #1460, opened 2026-07-16
fix(dataplane): correct truncating tick-to-nanosecond conversion in worker clock

3Hren:
The worker clock derived its tick-to-nanosecond multiplier with a floating-point division that was truncated when stored, so the clock ran roughly 0.15 percent slow. The error accumulated from worker start onward and reached a lag of tens of seconds over a long uptime, which was observed on a production rig. Worker time feeds firewall state and ACL timeout arithmetic as well as capture timestamps.

- Replace the truncating multiplier with a fixed-point one evaluated through a wider intermediate, cutting the relative error from roughly 1.5e-3 to 2e-10.
- Drop the tick quantisation that an earlier overflow workaround had to introduce, since the wider intermediate no longer overflows.
- Reject an uncalibrated timestamp counter instead of dividing by zero, which the new integer division would otherwise trap on.
- Add a unit test pinning the scaling across realistic clock frequencies, including ones at which the previous scheme was measurably wrong.
- Refresh the worker clock comment to describe the remaining unadjusted crystal drift.

### DIFF
--- a/lib/dataplane/config/zone.h
+++ b/lib/dataplane/config/zone.h
@@ -46,11 +46,12 @@ struct dp_worker {
 
 	// Allows to get current worker time.
 	//
-	// Currently, we init it only once
-	// and dont adjust.
-	// So, we have some drift, which is small but...
-	// (see tsc_clock docs). It is not important
-	// for now and fix should be easy, but need discuss.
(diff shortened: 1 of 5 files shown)


3Hren:
@codex review

PR #1461, opened 2026-07-16
refactor(pdump): collapse duplicated printer bodies into helpers

3Hren:
The concise printer repeated its layer four formatting verbatim in both the IPv4 and the IPv6 branch, and the ICMP printers repeated the echo detail body and the neighbour target address body between their version four and version six handlers.

- Collapsed each duplicated pair into a private helper, local to the printer since this formatting is specific to this module.
- Left the two remaining reported clones alone on purpose. One reads two different fields of two different message types and only looks alike, so a helper there would abstract over a coincidence of form. The other is four lines whose helper would need to take ownership and return a tuple for immediate destructuring, costing more than it saves.

The line count is not the point: this removes seven lines. The target address is the reason it is worth doing. The off by one corrected in the previous change lived in both copies at once and had to be fixed twice; each pair now has a single body, so a future mistake there cannot silently diverge.

Behaviour is unchanged, and the existing tests are the evidence: all thirty one pass without a single test being modified.

### DIFF
--- a/modules/pdump/cli/src/printer.rs
+++ b/modules/pdump/cli/src/printer.rs
@@ -60,6 +60,34 @@ pub fn pretty_print_metadata<W: Write>(mut writer: W, meta: &pdumppb::RecordMeta
     Ok(())
 }
 
+/// Formats an L4 protocol summary from its next-header protocol number and
+/// payload.
+///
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #1464, opened 2026-07-17
feat(lint): enforce the commit subject and PR title convention

3Hren:
Commit subjects and PR titles have drifted because the convention was documented but never enforced. Non-conforming subjects reach main through both merge strategies in use here: squash, where the pull request title becomes the subject, and rebase, where every commit lands as written.

- Add a linter for the documented subject format, checking the type against a closed set, a mandatory lowercase scope, an optional breaking-change marker, and a brief that is neither sentence-cased nor period-terminated.
- Install it as a commit-msg hook, so a bad subject is rejected before the commit lands rather than in review. The existing pre-commit hook is untouched, and `make hooks` already points git at the whole hook directory.
- Add a CI job checking the pull request title and every commit in the branch, so neither merge strategy can carry a bad subject onto main.
- Reconcile the documented convention itself, which listed one set of types for commits and a different one for pull request titles.

The linter is plain POSIX shell, so a simple text check needs no toolchain fetch and no build step before a commit hook can run.

Merge commits, autosquash markers and reverts are skipped. Existing history is never scanned: only new commits and the pull request title are checked, so the non-conforming subjects already on main are left alone.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+# Rejects a commit whose subject does not follow the "<type>(<scope>):
+# <brief>" convention from CLAUDE.md, so the rule is enforced before the
+# commit lands instead of being caught later in review or CI.
+if ! lint/commit/commitlint.sh --file "$1"; then
+	echo "commitlint failed; fix the subject above, or bypass with 'git commit --no-verify'" >&2
+	exit 1
+fi
--- a/.github/workflows/commit-lint.yml
+++ b/.github/workflows/commit-lint.yml
@@ -1,0 +1,27 @@
+name: Commit Lint
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+jobs:
+  commitlint:
+    name: Custom commitlint linter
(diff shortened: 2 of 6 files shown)


3Hren:
Good catch on the autosquash exemption — fixed.

The exemption was right for the hook and wrong for CI, so it now applies only to the hook. `git commit --fixup` stays usable locally, while the PR title and every commit in the branch reject `fixup!`/`squash!`/`amend!`, which closes the rebase-merge path you identified. Both directions are covered by tests.

`Revert "..."` and `Merge ...` stay exempt in all modes deliberately: reverts legitimately land on main (six in this history) and the commit range already excludes merges.

@codex review

3Hren:
Valid again — reproduced it on a scratch repo and fixed.

A commit with an empty message makes `git log --format=%s` emit a blank line, and the range loop treated that as noise to skip, so the commit was never validated. Since that stream is one collapsed line per commit, a blank line can only mean an empty subject, so it is now reported as one. The hook keeps its own blank-line skip, which is correct there because git strips leading blank lines from a message file.

Covered by a test using a throwaway repo, asserting both that the empty-subject range fails and that an all-conforming range still passes.

PR #1465, opened 2026-07-17
fix(devices): reclaim retired plain and vlan devices

3Hren:
A device leaving the configuration registry is parked on its creating agent's unused list instead of being freed, so the owning control plane can reclaim it once no live generation refers to it. The parking landed in #1303, which wired the mechanism but no consumer; #1306 then drained the list for the traffic generator only. Plain and vlan devices never drained at all, so every update leaked its predecessor into the agent's shared memory arena.

The arena filled after roughly three thousand updates. Since the pipeline operator reapplies configuration every thirty seconds, an affected device instance stopped accepting updates after a few hours and stayed broken until the control plane was restarted, reporting a failure to allocate counter names.

- Drain the unused device list on plain and vlan updates.
- Register the drain before the new configuration is built, so an update that fails because the arena is already full still reclaims and lets the next retry succeed, rather than leaving the instance stuck until a restart.
- Serialise device updates on both types. The reclaim is not safe against concurrent requests, and these were the only devices without that protection.
- Close the same gap in the traffic generator, where a failure to build the configuration skipped the drain.
- Cover the leak with tests that drive the real update path against a shared memory agent and require its free space to settle after the first update. They fail without the fix.

### DIFF
--- a/devices/plain/api/controlplane.c
+++ b/devices/plain/api/controlplane.c
@@ -55,6 +55,11 @@ cp_device_plain_free(struct cp_device *cp_device) {
 	cp_device_free(cp_device);
 }
 
+void
+cp_device_plain_drain_unused(struct agent *agent) {
+	cp_device_agent_drain_unused(agent, cp_device_plain_free);
+}
+
 struct cp_device_plain_config *
 cp_device_plain_config_new(
 	const char *name,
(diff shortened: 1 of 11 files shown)


3Hren:
@codex review

PR #1468, opened 2026-07-17
feat(lint): enforce type-level encapsulation in go

3Hren:
Privacy in Go is scoped to the package, so any function in a large package can reach into any struct's internals. That makes decomposition impossible to verify by review. This adds a linter restricting private field and method access to the receiver, and requiring tests to live in an external test package, where the compiler enforces the same boundary for both fields and functions.

Known violations are recorded in two self-cleaning ledgers so the debt can be paid down one entry at a time. An entry that no longer matches a live violation fails the run, so fixing the code requires removing the entry.

The linter runs from the pre-commit hook only for now, deliberately not in CI, until the ledgers are burned down. Install it with `make hooks`.

Sources that use cgo are exempt, because C struct fields cannot be told apart from private Go fields without type information. Command tests are exempt too, since a main package is unimportable and an external test could never reach it.

The two custom linters now share their source-tree walk and ledger handling. The walk skips git-ignored paths, so the tracked ledgers can never reference a private tree that a public clone does not have.


3Hren:
@codex review

PR #1469, opened 2026-07-17
fix(build): skip dot-directories when collecting protos

3Hren:
Dot-directories in this repository hold full copies of the tree: a cached source worktree and the per-agent worktrees. Every tool that walked the tree therefore saw each proto several times over.

Proto generation failed outright on duplicate definitions, the buf lint run reported the same duplicates, and the custom proto linter silently checked the stale copies instead of only the working tree.

Proto collection now skips dot-directories. The buf configuration cannot express that as a pattern, so it keeps an explicit list, which previously covered only one of the two offenders.

Protos belonging to private, git-ignored modules are still collected, so their generated code is unaffected.


3Hren:
@codex review

PR #1470, opened 2026-07-17
docs(claude): sync the module structure map with the tree

3Hren:
The module structure map had drifted on nine counts. It is a routing input for every contributor and agent, so a wrong line there produces confidently wrong work rather than an obvious error — a planning task was recently built on one of these stale claims.

Three modules listed as legacy have finished migrating, one sits between the two layouts and now says so, and only one still matches the legacy definition. Both early-stage descriptions were stale. One public module and one device were missing from the active lists entirely.

Every claim was checked against the tree using the criteria the document itself defines.

The reference list is left alone: which modules to imitate is an editorial choice, not a mechanical property of their layout.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -124,24 +124,30 @@ modules/<name>/
   internal/          # Optional: module-private Go packages (route only — discovery, rib).
 ```
 
-**Legacy** (acl, fwstate, nat64, pdump, route-mpls): no `bindings/`,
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

3Hren:
Fixed: route-mpls has neither tests/ nor fuzzing/, the text now says so. Verified the full layout matrix across all modules rather than only the directory I had originally checked.

@codex review

PR #1472, opened 2026-07-17
refactor(readiness): encapsulate scope and subscriber state

3Hren:
First entry of the encapsulation ledger burn-down. The readiness tracker reached directly into the fields of the two helper types it owns, from a free function and from half a dozen of its own methods. Both types now expose the behaviour instead, and the 23 ledger entries recording that debt are removed.

The shutdown path had grown its own copy of the state mutation, guarded by a subtly different condition for the transition timestamp. Both paths now share one method, which removes the duplication and the divergence with it. The two conditions were proven to agree on every reachable state, so behaviour is unchanged.

Both types are unexported, so the tracker's exported surface is untouched and no consumer is affected. The external test package is unmodified, which makes it an independent guard on the refactor.


3Hren:
@codex review

PR #1474, opened 2026-07-17
refactor(metrics): encapsulate bucket and entry state

3Hren:
Second entry of the encapsulation ledger burn-down. The histogram and the metric map reached directly into the fields of the small element types they own, from their own methods and from a free function. Both element types now expose the behaviour instead, and the 18 ledger entries recording that debt are removed.

The map's create path no longer reads the metric back out of the slice it just appended to, which removes one of those reaches outright rather than hiding it behind an accessor. The lookup predicate and the stored value are now answered by a single call rather than two. The guarantee that a delete predicate sees a cloned identifier now lives on the entry that owns it, and remains stated on the public method as well, since that is where consumers read it.

Tests move to the external test package, so the compiler rather than a linter forbids them from reaching into private state. The one test that verified storage is released when the last metric is deleted has no external observation point, so it now observes that through a new emptiness check. That check is the only public addition, and it is O(1) precisely because the storage is released.

Both element types stay unexported, so no consumer is affected. Behaviour was verified identical against the previous implementation across every bound set, boundary value and hash-collision case, including the construction and lookup paths.

### DIFF
--- a/common/go/metrics/histogram.go
+++ b/common/go/metrics/histogram.go
@@ -6,14 +6,40 @@ import (
 	"sync/atomic"
 )
 
+// BucketSnapshot is a point-in-time read of one histogram bucket.
 type BucketSnapshot struct {
 	UpperBound float64
 	Count      uint64
 }
 
+// bucket is one histogram bucket: an upper bound and an observation count.
 type bucket struct {
 	upperBound float64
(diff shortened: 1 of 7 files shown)


3Hren:
@codex review

PR #1475, opened 2026-07-17
refactor(xbackoff): export the options bag fields

3Hren:
Third entry of the encapsulation ledger burn-down. The constructor and the six option functions reached directly into the private fields of the options bag.

Unlike the two previous entries, there is nothing here to decompose: an options bag is a transport DTO that carries values from the option closures to the constructor and owns no behaviour, so giving it methods would be ceremony rather than encapsulation. The fix is to export its fields, which is already the convention everywhere else — of the sixty-eight non-empty option structs in the tree, sixty-seven already exported theirs, none mixed the two, and this was the last one still keeping them private.

The type's own state stays private, since it is reached only through the receiver. The exported surface is byte-identical, so neither consumer is affected: the bag itself is unexported and cannot be named from outside.

The change is a case-only rename, verified mechanically: lowercasing both revisions of the file makes them identical. Retry timing was additionally checked against the previous implementation, since the option reads inside the underlying backoff literal are the one place where a dropped field would compile cleanly and silently change behaviour.

### DIFF
--- a/common/go/xbackoff/backoff.go
+++ b/common/go/xbackoff/backoff.go
@@ -8,20 +8,20 @@ import (
 )
 
 type options struct {
-	max        time.Duration
-	multiplier float64
-	jitter     float64
-	sleeper    Sleeper
-	onRetry    func(attempt int, delay time.Duration, err error)
-	onReset    func()
+	Max        time.Duration
+	Multiplier float64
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1476, opened 2026-07-17
feat(route): report per-outcome packet counters and config gauges

3Hren:
Route previously exposed no dataplane counters at all, and its five drop reasons collapsed into one worker-global counter shared with every other module — a route drop could not be attributed even to the module, let alone to a cause.

Adds `route_drop_packets`/`_bytes` labelled by reason (no_route, ttl_expired, non_ip, empty_route_list, device_unresolved) and by address family, plus `route_forwarded_packets`/`_bytes` as the denominator.

Adds `route_nexthops` and `route_config_updated_timestamp_seconds`, and splits `route_fib_entries` by address family.

FIB size gauges are now measured once per apply instead of walking the LPM keyspace on every scrape.

Note for consumers: `route_fib_entries` now emits one series per family per config; use `sum by (config) (route_fib_entries)` for the previous single value.


3Hren:
@codex review

PR #1477, opened 2026-07-17
feat(cli): complete pipeline and function names

3Hren:
Pipeline and function name arguments now offer the same tab completion the module configs received, listing what the gateway currently knows.

These are the first tools whose listing call is shaped differently from the module ones: it is named differently, and it answers with identifiers rather than plain strings. Absorbing that took no change to the shared helper — only a different accessor at the call site, which is precisely what the helper was shaped for.

Names are offered exactly as the gateway reports them, so a completed value is always one the command accepts. This matters for functions, whose names carry a prefix that is tedious to type by hand and easy to get wrong.

Second of the staged sweep; the remaining command line tools follow.


3Hren:
@codex review

PR #1478, opened 2026-07-17
feat(cli): complete config names for acl, blackhole, dscp, mirror and nat64

3Hren:
Config-name arguments in five more module tools now offer the same tab completion the earlier ones received, listing the configs each module currently holds. All five share the simple listing shape, so they reuse the shared helper unchanged.

Two manifests are corrected along the way. One tool gains the minimum compiler version the completion helper requires. Another now declares the completion feature it had already been relying on, which until now it received only because a neighbouring crate happened to enable it — the build was correct by accident of the dependency graph rather than by declaration.

Third of the staged sweep; the remaining command line tools follow.

### DIFF
--- a/modules/acl/cli/src/args.rs
+++ b/modules/acl/cli/src/args.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 
 use clap::Parser;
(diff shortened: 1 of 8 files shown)


3Hren:
@codex review

PR #1479, opened 2026-07-17
refactor(bird-adapter): encapsulate the import holder

3Hren:
Fourth entry of the encapsulation ledger burn-down. The service reached directly into the fields of the per-import resource bundle it owns, from four of its methods, and its log-filter core used a non-standard receiver name.

The bundle now owns its behaviour. It is built by a single constructor rather than being allocated empty and filled a hundred lines later, and it exposes its teardown, its connection state, its reader run and its stream lifecycle as methods. Three cleanups fell out of that:

- Collapsing the two divergent teardown paths into one method removed a copy that carried nil guards which could never fire, since the map only ever holds a fully-built bundle.
- Single-phase construction let a write-only field be dropped. That field held a value copy of the live route store, aliased to the real one only by accident of the underlying type's layout, and a future change to that type would have silently desynced it.
- The session-info conversion moved off the list handler onto the bundle that actually has the data.

Both affected types are unexported, so the exported surface is unchanged. No work moved across the mutex that guards the import map, and the point at which a bundle becomes visible to the list handler is unchanged. One display-only timestamp is now sampled a few milliseconds earlier, before a replaced import's connection is torn down rather than after; it feeds only the printed session age.


3Hren:
@codex review

PR #1480, opened 2026-07-17
feat(cli): complete config names for fwstate, pdump, route, route-mpls and balancer2

3Hren:
The remaining config-holding module tools now offer tab completion for their config-name arguments, listing what each module currently holds. One of them reports its names under a differently-named field, absorbed by the shared helper with only a different accessor at the call site.

A few manifests are tidied to match the others: two tools gain the minimum compiler version the completion helper requires, and one now declares the completion feature it had been relying on only because a neighbouring crate enabled it.

Fourth of the staged sweep. What remains is the non-config listable names — the traffic generator, the route operator tables, and balancer session states — plus the private tools, which land separately.

### DIFF
--- a/modules/balancer2/cli/src/main.rs
+++ b/modules/balancer2/cli/src/main.rs
@@ -11,10 +11,15 @@ use std::{
 };
 
 use clap::{ArgAction, CommandFactory, Parser};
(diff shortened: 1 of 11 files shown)


3Hren:
@codex review

PR #1481, opened 2026-07-17
feat(cli): complete config names for the traffic generator and route operator

3Hren:
The traffic generator and the route operator tools now offer tab completion for their config-name arguments, listing what each currently holds. Both share the simple listing shape and reuse the shared helper unchanged.

Fifth of the staged sweep. What remains is the listable names that are not module configs — the neighbour tables and the balancer session states — plus the private tools, which land separately.


3Hren:
@codex review

PR #1482, opened 2026-07-17
feat(cli): complete neighbour table and balancer session-state names

3Hren:
Two more listable-name arguments now offer tab completion. The neighbour operator completes its table names from the table listing, and the balancer completes its session-state names from their own listing — distinct from the config names it already completes.

The neighbour listing answers with structured table entries rather than plain strings, so the shared helper takes a different accessor at the call site to read each table's name — the same way the pipeline and function tools already do.

Final public tool in the staged sweep.


3Hren:
@codex review

PR #1483, opened 2026-07-17
fix(cli): honour the typed endpoint when completing service names

3Hren:
The readiness and metrics tools complete a service name from the gateway registry. They recovered the connection endpoint from an empty synthetic command line, so an endpoint the user had already typed was ignored — completion always queried the default gateway and could offer services from the wrong host.

They now recover the connection flags from what the user has actually typed, reusing the same recovery the config-name completion already relies on. Completion reaches the gateway being targeted; a gateway that is down or unreachable still yields no candidates, silently.


3Hren:
@codex review

PR #1484, opened 2026-07-17
refactor(bird-adapter): encapsulate the wire decoder fields

3Hren:
Fifth entry of the encapsulation ledger burn-down, and the sibling of the import-holder change. The BIRD export decoder reached directly into the fields of the wire-format structs it casts from raw socket bytes, from several of its methods, and named two predicate receivers with a non-standard letter.

These are wire structs: their fields represent protocol bytes, are taken by address for unsafe pointer arithmetic, and appear in a compile-time offset, so they cannot be hidden behind accessors and cannot stay private. They are exported instead, which is the right shape for a wire type rather than a decomposition. The two predicates are exported because they are called from another type. No struct field is added, removed, or reordered, so the memory layout the casts depend on is unchanged; the size-assertion and the byte-fixture decode tests guard exactly that.

The behavioural decoder tests move to the external test package, reaching the package only through its public decoder entry point. The one assertion that checks the byte size of the private address structs stays internal, since those union-decomposition types have no meaning outside the package; its ledger entry carries that reason.

One same-package call site was updated to the renamed constructor, which is unavoidable within the package.


3Hren:
@codex review

3Hren:
Fixed: renaming newUpdateDecoder to NewUpdateDecoder made its loglint allowlist row stale; updated the row to the new name. `make lint-go` is now green.

@codex review

PR #1485, opened 2026-07-17
chore(cli): align shared CLI crate MSRV with the core library

3Hren:
Four shared CLI crates declared a minimum supported compiler version one release behind the core library they all depend on. Since that dependency already requires the newer version, the lower floor was unattainable in practice. Raising these four to match makes the declared minimum reflect what actually builds.


3Hren:
@codex review

PR #1486, opened 2026-07-17
test(route): cover per-outcome counters end to end

3Hren:
Verifies that each route packet outcome increments exactly the expected dataplane counter, read back through the same path the metrics service uses.

Covers nine of the eleven counters (v4/v6 forwarded, v4/v6 TTL-expired, v4/v6 no-route, non-IP, v4/v6 device-unresolved) and asserts both the packet count and the exact byte length, with sibling counters held at zero.

Closes the gap where the counters were validated only against a fake backend and by packet-list assertions, neither of which exercised the real counter storage.


3Hren:
@codex review

PR #1489, opened 2026-07-17
perf(pipeline): execution-context process helpers internal linkage

GeorgyKirichenko:
- Give the per-level execution-context process helpers internal linkage so
the device-entry dispatch path folds into a single inlinable region,
removing four consecutive external call boundaries.
- Drop the now-unused pipeline-level prototype and its forward declaration
from the public header, leaving the device input and output entry points as
the only external surface.


GeorgyKirichenko:
@codex review

PR #1496, opened 2026-07-20
build(deb): ship bash and zsh completions for yanet-cli

3Hren:
- Ships clap dynamic shell completions for all yanet-cli binaries in the yanet2-cli package, so tab-completion works right after install without editing shell startup files.
- Generates the bash and zsh completion loaders from the shipped binary list during the package build, and installs them into the standard bash and zsh completion directories.
- Recommends bash-completion so the completion directory is active out of the box.


3Hren:
@codex review

PR #1499, opened 2026-07-21
ci(lint): run the encapsulation linter in CI

3Hren:
The encapsulation linter was only ever invoked by the local pre-commit hook and the Go lint make target, never by CI. A contributor without the hooks installed could therefore land violations on the main branch, which is exactly what happened with the recent ACL metrics snapshot work — every unrelated commit made locally now trips the hook on someone else's debt.

- Add a workflow job that runs the encapsulation linter and its tests, mirroring the existing logger linter job.
- Extend the workflow path filters so a change to the linter or its allowlists retriggers the checks.
- Record the ACL controlplane accesses that currently fail the linter in the debt ledger, and drop the two rows that went stale when those accesses moved to another function.

The ledger rows are deliberate, temporary debt with a distinct reason string so the burn-down batch is easy to find. Encapsulating the ACL metrics snapshot state properly is tracked separately.

Note that making this job block merges still requires registering it as a required status check in the branch protection settings.

### DIFF
--- a/.github/workflows/go-lint.yml
+++ b/.github/workflows/go-lint.yml
@@ -5,12 +5,14 @@ on:
     paths:
       - '**/*.go'
       - 'lint/logger/**'
+      - 'lint/encapsulation/**'
       - '.golangci.yml'
       - '.github/workflows/go-lint.yml'
   pull_request:
     paths:
       - '**/*.go'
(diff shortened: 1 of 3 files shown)


3Hren:
@codex review

PR #1500, opened 2026-07-21
refactor(lint): clear the trivial rows from the encapsulation ledger

3Hren:
First batch of the encapsulation debt burndown. Sixteen of the ledger's rows shared a single shape: a control plane service reaching into a private field of its own package-private config struct, or a metrics aggregator calling a sibling object's private method. The convention asks for an exported field or method on the owning type even when that type is unexported, so that is what these types now have.

- Export the module handle and rule fields on the blackhole, forward and mirror config structs.
- Export the flag and mark fields on the dscp config struct.
- Export the per-gateway collect method in the pipeline operator.
- Rename the two fields of a function-local struct in the fwstate proto helpers, which is the smallest change that satisfies the rule there.
- Drop the sixteen corresponding ledger rows, which the linter would otherwise report as stale.

The renames are mechanical and behaviour preserving: no collisions with existing exported members, and the handle release paths keep their nil guards and ordering.

One pre-existing hazard surfaced during review and is deliberately left alone here: forward and mirror release the module handle without a nil guard on the update path but with one on the delete path. Giving those config types a method that owns the guard is tracked separately.


3Hren:
@codex review

PR #1501, opened 2026-07-21
refactor(lint): exempt same-type parameters from the encapsulation rule

3Hren:
The encapsulation rule requires a private field or method to be reached through the receiver. It misfires on a method that operates on another value of its own type, such as a bitwise operation on a 128-bit integer helper: there is no other spelling available, no encapsulation boundary is crossed, and no exported surface would make the code more reviewable. That is a rule wrong on a whole class rather than inconvenient on one instance, so the fix belongs in the linter rather than in the debt ledger.

- Accept a selector whose base names a method parameter declared with the same type as the receiver, ignoring pointer and generic wrapping.
- Keep everything else flagged: local variables, results, receiverless functions, parameters of a different or qualified type, and chained bases.
- Retire the two ledger rows that only existed because of the misfire.
- Document the exemption alongside the existing carve-out for files that use cgo.

The exemption stays a name-based approximation, like the rest of this linter's syntax-only analysis, and the doc comments say so. Its blast radius was measured against the previous ledger: exactly the two retired rows and nothing else. The boundary tests were checked by mutation — changing the type-name helper to accept qualified identifiers turns the two new tests red and no others.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -270,6 +270,10 @@ Meson orchestrates C/DPDK builds and Go binary compilation (via `custom_target` 
   `lint/encapsulation/allowlist-private.txt` — do not add new rows.
   Files that `import "C"` are exempt, because C struct fields cannot
   be distinguished from Go private fields without type information.
+  A method may also reach into the private fields or methods of a
+  parameter declared with the same type as its receiver, because an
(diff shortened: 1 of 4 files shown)


3Hren:
@codex review

PR #1502, opened 2026-07-21
refactor(operator): drop the redundant function applier wrapper

3Hren:
The decap and forward gateway actuators each kept a small private struct pairing a function applier with the function name, purely so the apply loop could name the function in errors and logs. The applier already carries that name in the chain spec it publishes, so the pairing stored the same string twice and existed only because the applier did not expose it.

- Expose the function name from the applier itself.
- Hold the appliers directly in both actuators and delete the wrapper struct.
- Retire the four encapsulation ledger rows that the wrapper's private fields required.

Behaviour is unchanged: identical error text, log messages and levels, branching and iteration order. The two actuators are maintained as near-verbatim copies of each other and stay parallel — their per-file diffs for this change are identical.

### DIFF
--- a/common/go/operator/function.go
+++ b/common/go/operator/function.go
@@ -79,6 +79,11 @@ func NewFunctionApplier(
 	}
 }
 
+// Name returns the function identifier the applier publishes.
+func (m *FunctionApplier) Name() string {
+	return m.spec.Name
+}
+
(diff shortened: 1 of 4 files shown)


3Hren:
@codex review

PR #1503, opened 2026-07-21
refactor(nat64): export the fields of the cached module config

3Hren:
The control plane keeps a small per-instance record pairing the NAT64 configuration with its module handle, and every request handler reached into that record's private fields — ten rows of the encapsulation debt ledger, all in one file.

- Export both fields of the cached record and update every handler, the clone helper and the default construction.
- Retire the ten corresponding ledger rows.

The exported-field payment rather than a method is deliberate: the module handle is released at exactly one site, so there is no duplicated nil guard for a method to own. The resulting shape matches the dscp module field for field. Nothing on the wire moves — the proto messages, the JSON rendering and the gRPC responses are untouched.


3Hren:
@codex review

PR #1504, opened 2026-07-21
refactor(route-mpls): own the module handle release in one method

3Hren:
Eleven ledger rows in one file. The cached per-config record kept its state private while every request handler reached inside it, and three separate places hand-rolled the same nil check before releasing the module handle.

- Export the record's prefix trie and module handle, its rule builder, and the nexthop conversion helper.
- Give the record a method that owns the handle release, and call it from all three sites.
- Retire the eleven corresponding ledger rows.

The method rather than a bare exported field is deliberate here: three copies of one guard is where a fourth eventually forgets the nil check, and two sibling modules already show that drift, guarding on one path and not the other. Behaviour is unchanged on every path — same ordering relative to the backend call, same error paths, same lock scope, no handle freed twice and none left reachable after release.


3Hren:
@codex review

3Hren:
@codex review

PR #1505, opened 2026-07-21
refactor(route-operator): give the FIB gauges and route identity a real surface

3Hren:
Fourteen ledger rows in the route operator. The per-module FIB gauges were written field by field from one method and read back field by field from another, the per-gateway metrics renderer was private while the aggregate renderer called it, and the route-identity predicate was private while the RIB and the route list both used it.

- Give the FIB gauges a method that owns the mapping from build statistics to gauge values, and export the gauges the renderer reads.
- Export the per-gateway metrics renderer, matching the shape already shipped in the pipeline operator.
- Export the route-identity predicate used by the RIB and the route list.
- Retire the fourteen corresponding ledger rows.

The predicate's logic is untouched — only its name changes — which matters because route identity is what the operator groups equal-cost paths by. The gauge lock boundaries are also unchanged: the mutex still guards only the map lookup and insert, and the gauge stores stay outside it.

The remaining row in this file is left as acknowledged debt. It covers a free function that seeds static configuration reaching into the route service, and exporting the accessor for it would be the wrong shape — the seeding belongs inside the service, which is a larger change than this batch should carry.


3Hren:
@codex review

3Hren:
@codex review

PR #1508, opened 2026-07-21
fix(gateway): keep a shared backend open while other services still use it

3Hren:
The gateway dials a single connection at startup and registers it under several service names at once — its own registry, auth, readiness and metrics services, plus every service hosted in the gateway process. Replacing any one of those names closed that shared connection, so a registration aimed at a single service silently took down every other service reachable through it, until the process was restarted. A displaced backend is now closed only once no registered name still resolves to it, so losing one name affects only that name.

A registration refresh also no longer changes how an existing service is recorded as being hosted. That mattered once stale backends became evictable: only self-registering backends are eligible for eviction, so accepting the caller-supplied hosting claim on a refresh let a registration mark the gateway's own shared connection as evictable, after which it was reaped on the next sweep with the same consequences. Registrations that genuinely move a self-registering service to a new address are unaffected, and a name lost to a takeover recovers exactly as it did before.

Also retires one row from the encapsulation ledger, which the removed assignment made stale.

### DIFF
--- a/controlplane/gateway/registry.go
+++ b/controlplane/gateway/registry.go
@@ -118,12 +118,12 @@ func (m *BackendRegistry) GetBackend(service string) (proxy.Backend, bool) {
 // the registry changed.
 //
 // The displaced backend (the previous one on an endpoint change, or the
-// redundant new one on an unchanged-endpoint re-registration) is closed after
-// the lock is released.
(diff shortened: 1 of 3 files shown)


3Hren:
@codex review

PR #1512, opened 2026-07-23
refactor(cli): drop redundant per-operator ready subcommands

3Hren:
The top-level ready command already discovers every operator readiness service through the gateway registry and covers scope filtering, so the per-operator ready subcommands only duplicated it.

- Remove the ready subcommand from the pipeline and route operator CLIs, keeping their other commands intact.
- Delete the decap and forward operator CLIs, which exposed nothing besides readiness.
- Drop the now-unused shared readiness helper and prune the corresponding dependency, workspace member, Makefile, and Debian packaging entries.
- Leave the operators' server-side readiness services untouched, since the top-level command still probes them.


3Hren:
@codex review

PR #1513, opened 2026-07-24
test(functional): make the qemu vm-ready timeout configurable

3Hren:
- Adds an environment override (YANET_VM_READY_TIMEOUT) for how long the functional-test harness waits for a VM to reach the shell prompt, defaulting to the previous 120 seconds.
- Lets slow emulated (no-KVM/TCG) runners raise the boot budget without a code change, which is what the self-hosted QEMU tier needs.


3Hren:
Addressed: threaded `VMReadyTimeout()` into the validation VM's initial readiness wait (pool.go), so a slow TCG/no-KVM run no longer discards the freshly-built booted template on a 60s timeout and falls back to cold-booting every slot. The in-loop restore readiness already routes through the same override.

@codex review

PR #1517, opened 2026-07-24
ci(workflows): install apt packages directly instead of caching them

3Hren:
The functional tests, and intermittently other jobs, have been failing with a missing command. The cause is a self-sustaining loop in the package caching action.

- The action never refreshes the package index, so once a mirror retires a pinned version the install fails with a not-found error. This is what broke the QEMU tier, through a transitive dependency of the emulator package.
- Its post step still saves a cache entry describing an empty package set. Because the cache key is derived from the package list, every later run restores that empty entry, installs nothing, and fails reporting a missing command.
- Purging the entry only exposes the original install failure, and the run that then fails immediately recreates the empty entry. The jobs cannot recover on their own, and the failure reproduces identically on the default branch.

Installing the packages directly removes both halves of the loop: the index is refreshed before every install, and no cache remains that can be poisoned. The package list of each job is unchanged.

Also give the fuzz workflow triggers scoped to its own file. It previously ran only on a schedule, so a change to it could not be verified by the pull request making that change, which is how this defect survived unnoticed. The scoping is a single literal path, so ordinary code changes do not start it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,14 +37,10 @@ jobs:
       cache_name: build-asan-test
 
     steps:
-      - name: update apt (act hack)
-        if: ${{ env.ACT }}
-        run: apt-get update
-
-      - uses: awalsh128/cache-apt-pkgs-action@latest
-        with:
-          packages: meson clang cmake python3-pyelftools libnuma-dev libpcap-dev git protobuf-compiler
-          version: 1.1
+      - name: Install build dependencies
+        run: |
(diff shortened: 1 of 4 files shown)


3Hren:
Closing in favour of a different approach. Removing the cache treats the symptom at the cost of CI time; the actual root cause is that the caching action reports success even when the package install failed, then stores that empty result as a valid cache entry. The replacement keeps caching and removes the cause.

PR #1532, opened 2026-07-27
fix(controlplane): log counter collection failures instead of dropping them silently

3Hren:
- The counters service discarded an entire metric family from its snapshot whenever collection failed, with no log line, no error to the caller and no counter of failures. A family disappearing from a scrape left nothing anywhere to explain it.
- The two families were also covered by a single guard each, so a failure in one was indistinguishable from a failure in the other.
- Both failures are now reported and the families stay independent, so one of them failing no longer hides the other.
- A failure still omits that family from the snapshot rather than serving stale values, which is the behaviour a consumer should see.

### DIFF
--- a/controlplane/builtin/counters.go
+++ b/controlplane/builtin/counters.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"strconv"
 
+	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1540, opened 2026-07-27
fix(controlplane): hold the config lock while draining the unused list

### LINKED ISSUE #1537: controlplane: cp_device/cp_object unused-list drain detaches without the configuration lock
## Concern

The unused-list drain helper detaches the per-agent `unused_*` list with a plain store and no synchronization. Meanwhile the registry retirement callback (`*_registry_item_free_cb`) parks items onto that same list while running under the configuration lock (it is reached via `cp_config_gen_free`, which every update path calls while holding `cp_config_lock`). An interleaving where the retirement callback publishes a new head just before the drain's detach store clears it loses the newly parked item permanently; for `cp_object` this leaves `loaded_object_count` nonzero and prevents agent reclamation. The fix direction (per the review comment): detaching the list must use the same synchronization as the parking callback.

- `cp_object_agent_drain_unused` (lines 115-128) detaches at 119-120 (`object = ADDR_OF(&agent->unused_object); SET_OFFSET_OF(&agent->unused_object, NULL);`) with no lock.
- The parking callback `cp_object_registry_item_free_cb` (lines 110-112) runs under the config lock via `cp_object_registry_fini` ← `cp_config_gen_free`.

The identical unsynchronized detach already exists for devices:

- `cp_device_agent_drain_unused` in `lib/controlplane/config/cp_device.c:454-467` detaches at 458-459 with no lock. Its parking callback `cp_device_registry_item_free_cb` (`cp_device.c:449-451`) runs under the config lock via `cp_device_registry_fini` ← `cp_config_gen_free`. Byte-for-byte the same lock-discipline gap.

`cp_module` has no drain helper at all, so it does not exhibit this specific race, but its parking side (`cp_module_registry_item_free_cb`, `lib/controlplane/config/cp_module.c:274-276`) shares the same assumption that the unused list is only ever touched under the config lock.

Surfaced by #1526.

GeorgyKirichenko:
The unused-list drain detached the per-agent list with a plain store while the parking callback runs under the configuration lock, so a concurrent park could publish a new head just before the detach cleared it and the newly parked item was lost.

- Take the configuration lock around the device drain so the detach matches the parking callback discipline.
- Add the missing module drain helper with the same lock discipline, mirroring the device path.
- The race is live in production: the device drain runs from the device update handler after the lock is released, while parking holds it.

The `cp_object` entity in #1526 has the identical shape and adopts the same lock on rebase.

Closes #1537.

### DIFF
--- a/lib/controlplane/config/cp_device.c
+++ b/lib/controlplane/config/cp_device.c
@@ -453,6 +453,14 @@ cp_device_registry_item_free_cb(struct registry_item *item, void *data) {
 
 void
 cp_device_agent_drain_unused(struct agent *agent, cp_device_free_fn free_fn) {
+	struct cp_config *cp_config = ADDR_OF(&agent->cp_config);
+
+	// Parking (cp_device_registry_item_free_cb) runs under cp_config_lock
+	// via cp_config_gen_free on every update path, so the detach that
+	// steals the list must take the same lock to avoid racing a concurrent
+	// park. Must not be called with cp_config_lock already held.
+	cp_config_lock(cp_config);
+
 	// Detach the whole list first so a device's own free cannot observe a
 	// half-walked list.
 	struct cp_device *device = ADDR_OF(&agent->unused_device);
@@ -464,6 +472,8 @@ cp_device_agent_drain_unused(struct agent *agent, cp_device_free_fn free_fn) {
 		free_fn(device);
 		device = prev;
 	}
+
+	cp_config_unlock(cp_config);
 }
 
 void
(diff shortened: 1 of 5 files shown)


GeorgyKirichenko:
Re: the module-drain mixed-destructor finding — not a defect: each module type attaches its own named agent (`modules/<name>/controlplane/mod.go` calls `AgentAttach(<type>)`, e.g. "forward", "nat64"), so an agent's `unused_module` list and its same-name `prev` chain are single-type. The single type-specific `free_handler` is therefore correct; a mixed list cannot occur.

GeorgyKirichenko:

PR #1541, opened 2026-07-27
fix(controlplane): unpark a reused device or module on registry upsert

### LINKED ISSUE #1538: controlplane: re-inserting a parked cp_device/cp_module/cp_object via registry upsert leaves it on the unused list (use-after-free)
## Concern

`*_registry_upsert` calls `registry_replace` with the retirement `*_registry_item_free_cb`. If a caller deletes or replaces an entity and then re-adds the **same** instance before the unused list is drained, the re-insert gives the parked entity's `config_item` a live registry reference again while it remains linked on `unused_*`. The next normal drain then finalizes and frees an entity still referenced by the active generation — a use-after-free. The fix direction (per the review comment): insertion must unpark such an entity, or reject reuse until retirement has completed.

- `cp_object_registry_upsert` (lines 181-217) calls `registry_replace(... cp_object_registry_item_free_cb ...)` at 204-211, with no check that `new_object` is itself currently parked on `unused_object`.

The same upsert-without-parked-check shape already exists for devices and modules:

- `cp_device_registry_upsert` in `lib/controlplane/config/cp_device.c:521-557` calls `registry_replace(... cp_device_registry_item_free_cb ...)` at 544-551, no parked-check on `new_device`. Identical shape.
- `cp_module_registry_upsert` in `lib/controlplane/config/cp_module.c:365-395` calls `registry_replace(... cp_module_registry_item_free_cb ...)` at 387-394, no parked-check on `new_module`. Identical shape.

Surfaced by #1526.

GeorgyKirichenko:
Re-inserting a parked entity through the registry upsert gave its config item a live registry reference again while it stayed linked on the creating agent unused list, so the next drain finalized an entity still referenced by the active generation, a use-after-free.

- Unlink the entity from the unused list at the top of each device and module upsert before the replace.
- Upsert already runs under the configuration lock on every production path, so the walk races nothing.

The `cp_object` entity in #1526 has the identical upsert shape and adopts the same unpark on rebase.

Closes #1538.

### DIFF
--- a/lib/controlplane/config/cp_device.c
+++ b/lib/controlplane/config/cp_device.c
@@ -518,6 +518,40 @@ cp_device_registry_lookup(
 	);
 }
 
+// Remove new_device from its creating agent's unused_device list if it is
+// parked there.
+//
+// Re-inserting a parked device via registry upsert would otherwise leave it
+// on the unused list while the registry also owns it, so a later drain would
+// free memory the registry still references. Called only after the replace
+// succeeds, so a failed upsert leaves the device parked and reachable by the
+// drain. Every production upsert path holds cp_config_lock, the same lock
+// parking uses.
+static void
+cp_device_unpark(struct cp_device *new_device) {
+	struct agent *agent = ADDR_OF(&new_device->agent);
+	if (agent == NULL) {
+		return;
+	}
+
+	struct cp_device *cursor = ADDR_OF(&agent->unused_device);
+	struct cp_device *above = NULL;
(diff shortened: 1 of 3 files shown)


GeorgyKirichenko:
Addressed both unpark findings (force-pushed): `cp_device_unpark` / `cp_module_unpark` now run only after `registry_replace` succeeds, so a failed upsert leaves the entity parked and reachable by the drain instead of dropping it.

@codex review

GeorgyKirichenko:

PR #1542, opened 2026-07-27
fix(controlplane): roll back a failed batched upsert without parking entities

### LINKED ISSUE #1536: controlplane: batched device/module/object upserts park earlier items on the unused list when a later entry fails
## Concern

The batched update routines upsert entities one at a time into the freshly copied `new_config_gen`. If a later entry fails after earlier new entities were already inserted, the error path finalizes the whole partial generation via `cp_config_gen_free`, which calls each registry's `*_fini` → `*_registry_item_free_cb` and parks the already-inserted entities on their creating agent's `unused_*` list. A caller performing ordinary error cleanup can then finalize/free those same entities while the parked list still points at them, so a later drain reaches freed storage; if the caller leaves them alone instead, their loaded counts and allocations leak. The fix direction (per the review comment): the failure path should roll back without retiring submitted entities, or otherwise provide unambiguous ownership handling.

`lib/controlplane/config/zone.c`:

- `cp_config_update_objects` (lines 680-726): upsert loop at 698-712, `goto error_free` at 710.
- `error_free:` at 721-722 calls `cp_config_gen_free` (`zone.c:157-196`), which calls `cp_object_registry_fini` at `zone.c:189` → `cp_object_registry_item_free_cb` (`lib/controlplane/config/cp_object.c:97-113`) parks the already-submitted objects on `unused_object`.

The same loop-upsert-then-`cp_config_gen_free`-on-error shape already exists for devices and modules:

- `cp_config_update_devices` in `lib/controlplane/config/zone.c:631-678`: upsert loop at 650-664, `goto error_free` at 662, error path `cp_config_gen_free` at 673-674 → `cp_device_registry_fini` (`zone.c:188`) → `cp_device_registry_item_free_cb` (`lib/controlplane/config/cp_device.c:437-452`). Identical shape.
- `cp_config_update_modules` in `lib/controlplane/config/zone.c:274-324`: upsert loop at 292-310, `goto error_free` at 308, error path `cp_config_gen_free` at 319-320 → `cp_module_registry_fini` (`zone.c:185`) → `cp_module_registry_item_free_cb` (`lib/controlplane/config/cp_module.c:267-277`). Identical shape.

Surfaced by #1526.

GeorgyKirichenko:
When a later entry of a batched device or module update failed, the error path finalized the whole partial generation and parked the already-submitted entities on the creating agent unused list, clashing with the caller own cleanup of those same entities.

- Drain the partial device and module registries without invoking the parking callback before freeing the failed generation, leaving the submitted entities at a zero refcount for the caller to reclaim directly.
- Treat a NULL free callback in the registry unref as a clean no-op, so the rollback drops references without side effects.
- Scoped to devices and modules, the entities whose updates the caller owns; functions and pipelines create their items inline and are out of scope.

The `cp_object` entity in #1526 has the identical update shape and adopts the same rollback on rebase.

Closes #1536.

### DIFF
--- a/lib/controlplane/config/cp_module.c
+++ b/lib/controlplane/config/cp_module.c
@@ -287,6 +287,14 @@ cp_module_registry_fini(struct cp_module_registry *module_registry) {
 	);
 }
 
+void
+cp_module_registry_fini_rollback(struct cp_module_registry *module_registry) {
+	// NULL free_func: items that reach refcnt zero are left in place for
+	// the caller to reclaim instead of being parked on an agent's unused
+	// list.
+	registry_fini(&module_registry->registry, NULL, NULL);
+}
+
 struct cp_module *
 cp_module_registry_get(
 	struct cp_module_registry *module_registry, uint64_t index
--- a/lib/controlplane/config/cp_module.h
+++ b/lib/controlplane/config/cp_module.h
@@ -181,6 +181,15 @@ cp_module_registry_copy(
 void
 cp_module_registry_fini(struct cp_module_registry *module_registry);
 
+// Drop every reference the registry holds without parking modules on an
+// agent's unused list.
+//
+// Used on the failure path of a batched upsert: the new generation was never
+// published, so its modules must not be retired (parked); the caller still
+// owns them and reclaims them directly. Idempotent on zero-init.
+void
+cp_module_registry_fini_rollback(struct cp_module_registry *module_registry);
+
 /**
  * Get a module from the registry by index.
  *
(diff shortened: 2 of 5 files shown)


GeorgyKirichenko:
Addressed the device-cleanup finding (force-pushed): the device facet is reverted entirely. Device services do not free a rejected device themselves — they rely on `defer DrainUnusedDevices` (`devices/*/controlplane/service.go`) reclaiming the parked device — so parking on the error path is the correct ownership for devices. The rollback is now scoped to modules only, where the caller frees directly via `module.Free()` and the parked entry would otherwise dangle.

@codex review

GeorgyKirichenko:

PR #1543, opened 2026-07-27
fix(controlplane): drain retired devices and modules across agents

### LINKED ISSUE #1535: controlplane: retired cp_device/cp_module/cp_object items are parked on the creating agent and never drained cross-agent
## Concern

When a replacement (same-name) agent instance updates or deletes an entity that a previous agent instance created, the registry retirement callback parks the entity on the **creating** (old) agent's `unused_*` list. The drain helper only walks the currently attached agent's list, so an entity parked on a prior agent is never reclaimed through the normal drain path. The fix direction (per the review comment): the drain path needs to cover prior agents, or retirement must make these items reachable from the current owner.

- `cp_object_registry_item_free_cb` (lines 97-113) parks on `ADDR_OF(&object->agent)` — the creating agent, not the agent performing the replacement (lines 110-112).
- `cp_object_agent_drain_unused` (lines 115-128) drains only the passed-in agent.

The same parking-on-creating-agent shape, with the same drain-only-current-agent gap, already exists for devices and modules:

- `cp_device_registry_item_free_cb` in `lib/controlplane/config/cp_device.c:437-452` parks on `ADDR_OF(&device->agent)` (lines 449-451); `cp_device_agent_drain_unused` in `lib/controlplane/config/cp_device.c:454-467` drains only the passed-in agent (detach at 458-459). Identical shape.
- `cp_module_registry_item_free_cb` in `lib/controlplane/config/cp_module.c:267-277` parks on `ADDR_OF(&module->agent)` (lines 274-276); there is no `cp_module_agent_drain_unused` at all, so parked modules have no drain path.

## Note on severity

PR #1526 adds a live `loaded_object_count` (incremented in `cp_object_init`, decremented in `cp_object_fini`) and gates prior-agent reclamation on it in `agent_attach` / `agent_free_unused_agents` (`lib/controlplane/agent/agent.c:275-276` and `:2028-2029`). For `cp_object` this turns the cross-agent drain gap into permanent retention of the old agent together with its entire arena: the parked objects keep `loaded_object_count` nonzero, which blocks reclamation, and the drain never runs on that agent. The `cp_device` / `cp_module` analogues lack a live counter (`loaded_module_count` is declared in `agent.h:40` but never incremented anywhere), so their symptom differs — abandoned items in a reclaimed agent's arena — but the underlying design gap is identical.

Surfaced by #1526.

GeorgyKirichenko:
A device or module replaced by a later same-name agent was parked on the creating agent unused list, but the drain only walked the currently attached agent, so items parked on a prior agent were never reclaimed.

- Walk the agent prev chain in both the device and module drains so prior-agent parking is reached, and take the configuration lock around the drain.
- Add the missing module drain helper mirroring the device path.
- Gate prior-agent reclamation on a live device count so a creating agent is not cleaned up before its parked devices are drained; the device update handler already drains in production, so the count settles.

The module count is intentionally left unmaintained here: modules have no production drain today, so maintaining the count would retain module-creating agents across reattach. That gating lands together with the production module drain wiring in #1526, which also extends all four fixes to `cp_object`.

Closes #1535.

### DIFF
--- a/devices/plain/api/controlplane.c
+++ b/devices/plain/api/controlplane.c
@@ -58,6 +58,11 @@ cp_device_plain_free(struct cp_device *cp_device) {
 void
 cp_device_plain_drain_unused(struct agent *agent) {
 	cp_device_agent_drain_unused(agent, cp_device_plain_free);
+	// The drain just freed devices parked on prior agents, dropping their
+	// loaded_device_count; reclaim any prior agent that reached zero so the
+	// count gate does not retain empty predecessors. Runs after the drain,
+	// which releases cp_config_lock.
+	agent_free_unused_agents(agent);
 }
 
 struct cp_device_plain_config *
--- a/devices/trafgen/api/controlplane.c
+++ b/devices/trafgen/api/controlplane.c
@@ -126,6 +126,8 @@ cp_device_trafgen_free(struct cp_device *cp_device) {
 void
 cp_device_trafgen_drain_unused(struct agent *agent) {
 	cp_device_agent_drain_unused(agent, cp_device_trafgen_free);
+	// Reclaim prior agents whose parked devices the drain just freed.
+	agent_free_unused_agents(agent);
 }
 
 struct cp_device_trafgen_config *
--- a/devices/vlan/api/controlplane.c
+++ b/devices/vlan/api/controlplane.c
@@ -60,6 +60,8 @@ cp_device_vlan_free(struct cp_device *cp_device) {
 void
 cp_device_vlan_drain_unused(struct agent *agent) {
 	cp_device_agent_drain_unused(agent, cp_device_vlan_free);
+	// Reclaim prior agents whose parked devices the drain just freed.
+	agent_free_unused_agents(agent);
 }
 
 struct cp_device_vlan_config *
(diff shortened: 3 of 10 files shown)


GeorgyKirichenko:
Addressed two of the three findings; the third is not a defect (force-pushed):

- Mixed-destructor across the prev chain: not a defect. Each device type attaches its own named agent (`devices/{plain,vlan,trafgen}/controlplane/mod.go` calls `AgentAttach(<type>)`), so an agent's `unused_device` list and its same-name `prev` chain are single-type. The type-specific `free_fn` is correct; a mixed list cannot occur.
- Reclaim after drain: addressed. The device drain wrappers (`cp_device_{plain,vlan,trafgen}_drain_unused`) now call `agent_free_unused_agents` after the drain, so a prior agent whose `loaded_device_count` reached zero is reclaimed immediately instead of waiting for an unrelated update.
- Count on early failure: addressed. `loaded_device_count` is now incremented when a device is parked (`cp_device_registry_item_free_cb`) and decremented when the drain frees it, rather than at init/fini. A device that is built but never parked (early update failure) no longer affects the count.

@codex review

GeorgyKirichenko:

PR #1552, opened 2026-07-28
refactor(cli): defer output payload construction to the active backend

3Hren:
The CLI output facade took a fully built serializable payload alongside a human render callback, but only one of the two is ever used: a serializing format ignores the callback, a human format ignores the payload. Every read subcommand therefore built a structure that its selected format immediately discarded.

The payload now crosses the backend boundary as a closure, so a serializing backend builds it and a rendering one never does. The trait documents the exclusivity as an obligation on implementations, since that is what the saving rests on.

The empty-result flag and its message leave the facade as well. The render callback already knows whether its own data is empty, so the flag was a second source of truth; a dedicated helper now reports the empty case and keeps it on the error stream, so a piped run still reads as no rows.

All call sites are converted, and three of them now build their display or payload structures inside the callback that needs them, which is where the deferral actually pays off. Observable output is unchanged in both formats.

### DIFF
--- a/cli/core/src/output.rs
+++ b/cli/core/src/output.rs
@@ -46,22 +46,16 @@ pub trait Output: Send + Sync {
 
     /// Output result data.
     ///
-    /// Backends receive both a structured `payload` (for serialization-
-    /// oriented formats) and a `render` callback (for free-form rendering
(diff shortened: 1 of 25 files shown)


3Hren:
@codex review

PR #1554, opened 2026-07-28
fix(build): derive the aarch64 instruction set from the selected platform

3Hren:
The project hardcoded a single aarch64 instruction-set baseline while DPDK derived its own from the selected platform. Both arrived on the same compile command as conflicting switches, and with warnings treated as errors the compiler refused them outright.

The mismatch stayed hidden under the default platform, where the two sides happen to resolve to the same string. It appeared only when tuning for real Ampere silicon, which is exactly the configuration that matters on that hardware.

- The baseline now comes from the same platform selector DPDK uses, mirroring how a single option already drives both sides on x86.
- Configuration fails loudly if the two ever disagree, because DPDK probes the compiler at configure time and can resolve a different switch than expected on an older toolchain.
- A platform with no mapping is rejected rather than guessed, since such a platform would otherwise build the project and DPDK at different baselines inside one binary.
- Verified on x86 unchanged, and on aarch64 across all three supported platforms, where the conflict is now absent.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -17,15 +17,9 @@ project(
 yanet_project_args = ['-g', '-D_GNU_SOURCE']
 if host_machine.cpu_family() in ['x86', 'x86_64']
   yanet_project_args += '-march=' + get_option('cpu_arch')
-elif host_machine.cpu_family() == 'aarch64'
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #1556, opened 2026-07-28
style(build): rephrase two comments to start with a capital word

3Hren:
One comment opened with an architecture name and the other with a function name, both lowercase.

Project style requires a comment to start with a capital word, and an opener that is an identifier or an abbreviation to be rephrased rather than simply capitalised.

- Comments only, no build logic touched.


3Hren:
@codex review

PR #1557, opened 2026-07-28
fix(common): add a portable fallback for the B-tree block search

3Hren:
The B-tree headers required a vector instruction set that exists on only one architecture, so they could not be compiled anywhere else. A filter header includes them unconditionally, which took four packet-processing modules down with them. This was the last remaining blocker for building the dataplane on aarch64.

The vector path is now selected only when the compiler reports that instruction set available, with a plain loop everywhere else.

- The stored block layout and its biased representation are unchanged, so both paths read and write the same shared memory. Verified byte for byte across seventy-two generated trees.
- The two paths were shown to agree by differential testing over roughly four million block searches and eight hundred thousand whole-API queries, covering the value extremes where the bias wraps, duplicates, and every match position including none.
- Verified on x86 with and without the vector extension, cross-compiled for aarch64, and executed on aarch64 under emulation.

### DIFF
--- a/common/btree/u16.h
+++ b/common/btree/u16.h
@@ -2,23 +2,26 @@
 
 #include "../big_array.h"
 #include <assert.h>
-#include <immintrin.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+
+#if defined(__AVX2__)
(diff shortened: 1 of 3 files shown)


3Hren:
@codex review

PR #1558, opened 2026-07-28
fix(nat64): guard byte literals passed to the copy helper in tests

3Hren:
On several target architectures the packet copy helper is a preprocessor macro rather than a function. A brace-enclosed byte literal passed to it is split on its own commas, because only parentheses shield an argument from the preprocessor, so the build fails on argument count.

- The literals are now parenthesized, which hides the commas from the preprocessor and leaves the generated code identical on every architecture.
- A structural scan for arguments carrying a comma the preprocessor would count but the language would not confirms these were the only affected call sites in the tree.
- Verified on x86 with a clean build and the full test suite, and by cross-compiling the affected object for aarch64, where it previously failed.


3Hren:
@codex review

PR #1559, opened 2026-07-28
feat(build): accept the auto-detecting platform value on aarch64

3Hren:
The instruction-set baseline is chosen from a mapping keyed by the selected platform, which covered only the portable value and the two named server chips. It rejected the value that asks the build system to detect the host CPU itself — which is both the upstream default and the natural choice when building on the target machine.

That value now adopts whatever baseline the detection resolved, since there is no fixed flag that could be tabulated for it. Every other platform keeps the mapping and the consistency check unchanged, verified byte for byte against the previous behaviour.

- The option's description now documents the detecting value, which was previously absent from the only place the vocabulary is discoverable.
- A cross build can never reach this path: the build system takes the chip from the cross file and requires it to name a supported one, failing first in both of the two ways the situation can arise. Both were reproduced.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -173,47 +173,60 @@ if host_machine.cpu_family() == 'aarch64'
   # config has no mcpu switch would otherwise configure and compile
   # silently with yanet and DPDK at different ISA baselines, so an
   # unlisted platform is rejected here rather than guessed.
+  #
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1562, opened 2026-07-29
feat(nat64,fwstate,route): add -4/-6 short aliases for address-family flags

3Hren:
Every address-family option across the module and operator CLIs now accepts a single-character alias alongside its long form, matching what users expect from common networking tools.

The long forms are untouched, so existing scripts, functional tests and documentation stay valid. This is a strict widening of accepted input: the digit shorts were previously a hard parse error, so no command line that worked before changes meaning.

Each affected CLI also gains a parser-validity test. A short-flag collision is reported only under debug assertions and is silent in a release binary, so until now nothing guarded against one. The tests were confirmed to fail when a real collision is injected.

The MTU options in the address-translation CLI keep long forms only, since they are distinct options rather than address-family selectors.

### DIFF
--- a/modules/fwstate/cli/src/args.rs
+++ b/modules/fwstate/cli/src/args.rs
@@ -138,7 +138,7 @@ pub struct EntriesCmd {
     pub config_name: String,
 
     /// Use IPv6 map instead of IPv4
-    #[arg(long)]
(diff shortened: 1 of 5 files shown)


3Hren:
@codex review

PR #1563, opened 2026-07-29
docs(common): correct the coverage claim in the crc32 test comments

3Hren:
Both comment blocks in the crc32 test are reshaped into the project's brief, blank line, detail form.

The test's brief also claimed coverage of every tail length from one to eight bytes. That overstated the vector set: an eight-byte input is consumed entirely by the main loop and leaves no tail at all, and there is no eight-byte tail branch to fire. The neighbouring clause about an input past eight bytes became misleading for the same reason, so the description now states per vector class what the golden values actually exercise.

Comment-only. The emitted code is unchanged apart from line numbers, and the golden-vector test still passes.

### DIFF
--- a/tests/common/crc32_test.c
+++ b/tests/common/crc32_test.c
@@ -14,9 +14,11 @@ struct crc32_vector {
 };
 
 // These values were captured by running crc32() on the x86 build, which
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #1564, opened 2026-07-29
refactor(metrics): hoist duplicated proto metric constructors into commonpb

3Hren:
Seven control-plane and operator packages each carried a byte-identical unexported constructor for a protobuf metric or label. The helpers now live once in the shared commonpb package, alongside the existing metric conversion helpers, and every call site uses them.

- Adds three constructors to the shared protobuf package, following its established naming pattern for message constructors.
- Removes the fourteen local copies across the ACL, forward, fwstate and route modules, the pipeline and route operators, and the built-in counters service.
- Migrates all one hundred twenty four call sites.

The change is purely mechanical: emitted metric names, values and labels are unchanged. The two remaining histogram helpers are deliberately left in place, since their signatures genuinely differ.

### DIFF
--- a/common/commonpb/v1/metrics.go
+++ b/common/commonpb/v1/metrics.go
@@ -81,3 +81,29 @@ func MetricRefsToProto[T metrics.Counter | metrics.Gauge | metrics.Histogram](
 	}
 	return res
 }
+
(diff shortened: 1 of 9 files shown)


3Hren:
@codex review

PR #1565, opened 2026-07-29
fix(metrics): ignore invalid NaN observations instead of panicking

3Hren:
Observing a NaN value panicked the control plane with an out-of-range index. Every ordered comparison against NaN is false, so no bucket accepts it and the bucket search runs past the end of the slice.

A NaN is now treated as invalid input and dropped. It is deliberately not folded into the catch-all bucket instead: that bucket feeds percentile estimation, so a NaN parked there would inflate the reported tail latency.

The defect was latent rather than live. Both call sites observe a duration in seconds, which can never be NaN, so no production path could reach the panic — this closes the gap for future callers of a shared library.

Adds a regression test covering the dropped NaN, plus assertions that positive and negative infinity are still recorded, which fence against a future guard that over-rejects.

### DIFF
--- a/common/go/metrics/histogram.go
+++ b/common/go/metrics/histogram.go
@@ -64,8 +64,19 @@ func NewHistogram(bounds []float64) *Histogram {
 }
 
 // Observe records a new value.
+//
+// NaN is treated as invalid input and dropped. No bucket accepts NaN, so an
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1566, opened 2026-07-29
chore(agents): let the planner delete the items it closes

3Hren:
The planner's charter instructed it to delete an item file whenever an item reached a closed state, while its hard constraints forbade every file-mutating command without exception. Being ordered to do something it was simultaneously denied the means to do, it repeatedly attempted the deletion anyway, and closed work accumulated in the tracker until someone swept it by hand.

It now has a single narrow carve-out: it may remove its own tracker item files and nothing else, one named file at a time.

Closing an item now removes both the item file and its index row in the same pass, on the close path and on the reconciliation path alike. The outcome survives in the git history and the linked pull request, so nothing is lost by removing both, and a finished item can no longer linger in the index looking like live work.


3Hren:
@codex review

PR #1571, opened 2026-07-29
refactor(cli): unify empty-result output across every CLI

3Hren:
Every CLI reported an empty result differently: 27 sites used a lowercase fragment, 4 a capitalized sentence, and 4 more hand-rolled a `println!` straight to stdout — one of them coloured yellow, two preceded by a column header that made a zero-row result indistinguishable from a table whose body never arrived.

- All of them now report through `output::empty`/`output::empty_with_hint`, which renders a `[–]`/`[-]` mark on stderr, greyed when colour is on.
- The wording adopts the register the Web UI already uses for the same data — `No decap configurations found.` — copied byte-for-byte wherever a Web string exists, and extended on the same template elsewhere.
- A command given a filter argument now names it, so a typo is visible: `No FIB entries found for 'route0'.` Previously only 4 of 31 sites did this.
- The note is suppressed entirely when stdout is not a terminal, or when the selected format serializes, so a pipe or a redirect gets zero bytes on both channels. That suppression lives in the primitive rather than in a format check each caller has to remember.
- Every list command additionally suggests the command that would create the missing thing; each suggestion was checked to actually succeed in the state that prints it, which is always a first create.
- `CLAUDE.md` records the convention.

The convention was chosen against the behaviour of `ip`, `git`, `podman`, `systemctl`, `kubectl`, `gh` and FRR, none of which colour an empty-result message, and against this project's own success-message and Web UI registers.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -394,6 +394,11 @@ Meson orchestrates C/DPDK builds and Go binary compilation (via `custom_target` 
   too — e.g. if the message declares `nexthop_addrs` (tag 3) before
   `do_flush` (tag 4), the literal must place `nexthop_addrs` before
   `do_flush`, not append it last. `rustfmt`/`clippy` do not catch this.
+- **Empty CLI results**: report through `output::empty`/`empty_with_hint`,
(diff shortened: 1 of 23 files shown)


3Hren:
@codex review

PR #1572, opened 2026-07-29
perf(filter,acl): share the IPv6 half-classification across both ACL filters

3Hren:
The ACL module issues two filter queries per IPv6 packet class whose signatures both carry the source and destination address attributes, so the same address tries were walked twice per pass and four times per packet across the inbound and outbound configurations. This builds one shared per-direction classification over the union of both signatures' rule sets, with remap arrays translating union classes back to each local classifier, so a single walk answers both.

The implementation is migrated from a stranded WIP branch that had never been measured on real hardware. Only the sharing is taken; the packed union tries bundled in that commit are left for a separate change with its own evidence.

### Measured on the lab firewall stand

- Delivered throughput at saturation rises from 16.43 to 17.83 million packets per second, **+8.5%**, against a noise floor below 1%.
- Baseline was measured three times, once *after* the candidate and across two full binary swaps, and reproduced itself to 0.6% — so stand drift does not explain the gain.
- Inbound ACL falls 223 to 177 ns per packet, outbound 212 to 170 ns.
- Instructions per packet fall 4.8% and instructions per cycle rise 3.3%. Both moving is the signature this latency-bound path requires: only removing dependent trie hops can lift IPC here.
- Both sides were built locally from identical sources with byte-identical build configuration, so the toolchain is not a confound.

### Verdict equivalence is the deliverable

A classification bug on a firewall does not slow anything down, it lets a packet through, so equivalence carries more weight here than the speedup.

- The differential test compiles the same ruleset both ways in one process and asserts identical per-packet verdicts **and** identical per-rule counter vectors — a verdict-only assertion passes when two rules swap outcomes symmetrically.
- It runs at twelve thousand rules by default, sized to about 1.3 GiB so it survives CI, with a forty thousand rule tier behind an environment variable.
- The corpus deliberately includes addresses landing in one local partition but not the other, masks that are non-contiguous across the address yet contiguous within each half, and the fragment path that reaches only one of the two signatures.
- A positive control asserts the kill switch is unset and both filter rule counts are non-zero, because there are three separate ways the shared classification is legitimately not built and any of them would otherwise let the test compare the unshared path against itself.
- Equivalence was additionally confirmed on the stand against the live production ruleset under real traffic, with an oracle that recomputed the local classification per packet: 4.75 million packets classified, zero divergences. That instrument has served its purpose and is not part of this change.

### Deliberate choices worth a reviewer's attention

- A failure to build the shared classification rejects the configuration apply, exactly as a failure of any other filter does. An earlier revision fell back to the unshared path instead; that bought a flag in shared memory, a metric and a second branch in the hot loop for a case a correctly sized arena never reaches.
- The unshared path remains, keyed on the absence of the shared structure rather than on stored state, because it is what the differential test compares against.
- The leaf vertex indices are derived from the signature rather than written as literals; getting them wrong would reinterpret a different attribute's payload with no build error.
- Rule compilation costs about 15% more time and 1% more arena.

### DIFF
--- a/lib/filter/classifiers/net6.h
+++ b/lib/filter/classifiers/net6.h
@@ -1,6 +1,9 @@
 #pragma once
 
+#include <stdbool.h>
+
 #include "common/lpm.h"
+#include "common/memory_address.h"
 #include "common/value.h"
 
 struct net6_classifier {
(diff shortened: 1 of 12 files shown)


3Hren:
@codex review

3Hren:
@codex review

GeorgyKirichenko (approved):

PR #1585, opened 2026-07-30
feat(lint): consolidate the Go linters and gate eleven more conventions

3Hren:
Merges the logger and encapsulation linters into a single style gate and extends it with eleven further checks drawn from the coding conventions, so rules previously enforced only by review are now mechanical.

- Consolidates three Go linters into one that walks the source tree a single time, and collapses the wiring to one Make target, one pre-commit block and one CI job.
- Adds checks for receiver naming, loop index naming, map construction, gRPC client creation, sugared-logger usage, log message and field-key casing, test context usage, handler parameter naming, constructor naming and logger field placement.
- Preserves the three existing checks and every one of their allowlist reasons unchanged, so no tracked debt loses the note explaining how to pay it off.
- Gives each check its own file scope, so production, test and cgo code are each treated according to what the rule can meaningfully mean there.
- Ledgers all known violations in one allowlist keyed by check, path and site, with a mandatory reason per row. A row that no longer matches a live violation fails the run, so a fix and the removal of its row land together.
- Leaves two of the new checks with an empty ledger; they exist purely to prevent a regression.
- Excludes test-framework failure messages from the log message casing check, because the testing helpers expose some of the same method names as the logger API and their messages are read by a human running the test.


3Hren:
Thanks — four of the five were real and are now fixed; the fifth was already addressed in a force-push that landed after the reviewed commit. Details per finding.

**Detect unaliased v1 protobuf imports** — real, fixed. Import local names were derived from the last path segment, so `.../decappb/v1` resolved to `v1` and the repository's ordinary unaliased shape was never matched. A versioned trailing segment now resolves to the segment before it, and the derivation is shared with the general import-alias resolver rather than duplicated. Verified on a probe package: an unaliased `.../decappb/v1` import with a blank `_ *decappb.UpdateConfigRequest` method parameter is now reported.

**Catch assigned loop indices too** — real, fixed. Only a short declaration in the loop's init clause was inspected, so a loop reusing an already-declared variable was accepted. An assignment in the init clause is now covered too. Verified: `var i int; for i = 0; i < 3; i++ {}` is now reported.

**Flag anonymous zap logger fields** — real, fixed. An embedded field has no name, so the emitting loop produced nothing and a non-last logger could slip through by being embedded. The embedded case is now reported, keyed by its type name. Verified: a struct embedding the logger ahead of another field is now reported.

**Run zap call checks without same-file imports** — real, fixed. Eligibility for the logger-related checks was file-scoped, so a method using a logger field declared in a sibling file bypassed them entirely. Eligibility is now package-scoped: if any file in a directory imports the logging package, every file in that directory is eligible. Alias resolution stays per file. Verified on a two-file probe package where only the struct's file imports the package: both the uppercase-message and the sugared-logger calls in the other file are now reported.

**Make loopindex allowlist keys unique per loop** — correct, and already fixed before this review was posted. Keys now carry a per-occurrence ordinal, so each occurrence gets its own row. Evidence on your own example: the two loops in `TestRingBufferRunReaders` occupy two independent rows, and deleting only the second row makes exactly the second loop report while the first stays suppressed. Adding a fourth loop to an already-ledgered function with three now exits non-zero instead of being absorbed. Repository-wide, live violations and ledger rows are now equal at 346, where before they were 346 against 299.

Each fix carries a fixture that fails without it, and each predicate was mutation-tested by corrupting it and confirming a red test. The four fixes surfaced no new violations in the tree, so the ledger is unchanged at 346 rows with none stale.

@codex review

PR #1586, opened 2026-07-30
refactor(filter): build LPM from range_index instead of dual-writing in collector

GeorgyKirichenko:
The range collector previously inserted every range into both the LPM and the range_index simultaneously. This coupled the collector to lpm.h and created a latent consistency hazard: the two structures stayed in sync only because the emit loop was a single chokepoint.

Now the collector produces only a range_index. A new range_index_build_lpm() derives the LPM from the range_index in a dedicated pass — walking the radix in ascending order and inserting [from..next_from-1] -> value. The collector no longer includes lpm.h.

- Runtime query path (lpm4_lookup / lpm8_lookup) and in-memory LPM layout are unchanged.
- Zero dataplane/ABI impact — purely a compile-time construction-ordering refactor.
- Only the classic net4 and net6 filter compilers are affected; the _fast variants are untouched.

### DIFF
--- a/common/range_collector.h
+++ b/common/range_collector.h
@@ -6,7 +6,6 @@
 #include "memory.h"
 
 #include "key.h"
-#include "lpm.h"
 #include "radix.h"
 #include "range_index.h"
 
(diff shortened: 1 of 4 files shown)


GeorgyKirichenko:
@codex review

PR #1589, opened 2026-07-30
feat(cp): spawn per-worker counter storage for cp_object in the generation ectx

GeorgyKirichenko:
- Give each cp_object a per-worker execution context (object_ectx) bound in config_gen_ectx, mirroring device_ectx.
- A counter_storage is spawned from the object's counter_registry for each worker and registered in the worker's counter storage registry, realizing the deferred expansion of object counters into per-worker storages.
- Add counter-storage-registry lookup and insert for objects, and build and free the objects ectx array alongside devices.
- Objects are inert (no packet handler), so object_ectx carries only the cp_object reference and its counter storage.

### DIFF
--- a/lib/controlplane/config/cp_counter.c
+++ b/lib/controlplane/config/cp_counter.c
@@ -346,6 +346,28 @@ cp_config_counter_storage_registry_insert_device(
 }
 
 struct counter_storage *
+cp_config_counter_storage_registry_lookup_object(
+	struct cp_config_counter_storage_registry *registry,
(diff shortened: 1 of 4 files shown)


GeorgyKirichenko:
Addressed: object_ectx_create now sets its counter_storage field before the registry insert (same fix as the chain/function/pipeline/device sites in #1592), so object_ectx_free reaches it on insert failure. This branch is rebased on top of #1592.

@codex review

PR #1595, opened 2026-07-30
fix(cp): key the object registry and API by type and name

GeorgyKirichenko:
- Keys the control-plane object registry and its lookup, delete, and generation-lookup APIs by both type and name instead of name alone.
- Threads type and name through the config-generation and agent object APIs, matching how the module registry is already keyed.
- Makes a same-name, different-type object a distinct slot, so object identity is consistent end to end.

### DIFF
--- a/api/agent.h
+++ b/api/agent.h
@@ -228,11 +228,16 @@ agent_update_objects(
 	yanet_error **err
 );
 
-// Delete the shared object with the specified name.
+// Delete the shared object with the specified type and name.
 //
 // @return -1 if the object does not exist, 0 on success.
 int
-agent_delete_object(struct agent *agent, const char *name, yanet_error **err);
+agent_delete_object(
+	struct agent *agent,
+	const char *object_type,
+	const char *object_name,
+	yanet_error **err
+);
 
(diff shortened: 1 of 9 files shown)


GeorgyKirichenko:
@codex review

GeorgyKirichenko:
Addressed: the object registry is now keyed by type and name throughout (registry comparator, lookup, upsert, delete, and the generation lookups/delete), mirroring cp_module_registry. A same-name different-type object is a distinct slot, so the counter-continuity lookup is consistent end to end and values survive a type-changing replacement. See commit 928f4632.

The P1 (module ABI bump for the config_gen_ectx layout change) is still open — will follow up.

@codex review

GeorgyKirichenko:
Both P1s addressed:
- ABI bump: YANET_MODULE_ABI_VERSION 9 to 10 (dc3d4372).
- Private agent_delete_object prototype: updated to the four-argument type+name form (25e64117), matching the public api/agent.h declaration and the definition.

@codex review

GeorgyKirichenko:
@codex review

PR #1601, opened 2026-07-31
fix(commentlint): isolate git-backed tests

3Hren:
- Run commentlint directly from the pre-commit hook instead of executing its Git-backed unit tests.
- Isolate test fixture Git commands from linked-worktree repository variables.
- Preserve Git-selected alternate indexes in production scans and cover both regression paths.


3Hren:
@codex review

PR #1602, opened 2026-08-01
fix(cli): anchor the unregistered-service marker at the message start

3Hren:
- The CLI classified a not-found status as an unregistered service by searching for a marker anywhere in the server's message, so a config whose name quotes that marker was misclassified and the CLI printed an unrelated hint about the operator being down instead of the real message.
- The marker now has to open the message, which no resource message that merely quotes it inside a config name can do. Both classification sites share one predicate.
- Added regression coverage for both sites, verified to fail if the anchored match is loosened again.

### DIFF
--- a/cli/core/src/errors.rs
+++ b/cli/core/src/errors.rs
@@ -217,10 +217,29 @@ impl Error {
     }
 }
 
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #1604, opened 2026-08-01
fix(web): keep loading the fwstate page when one configuration disappears

3Hren:
The fwstate page fanned out over its configuration list as a single all-or-nothing batch, so a configuration deleted between listing it and reading it failed the entire load and left the page blank behind a generic error.

- Reuse the shared loader that treats a configuration the control plane no longer knows as gone rather than as a failure, so the remaining configurations still render.
- Warn once when the control plane knows none of the configurations that were listed, so an empty page is never silently indistinguishable from a genuinely empty instance.
- Drop a linked ACL configuration from the table once it no longer exists, instead of leaving it on screen permanently marked unavailable.

A genuine backend failure still fails the load and still marks a row unavailable, so real outages stay as visible as before. This completes the fan-out fix started for the decap, forward, route and acl pages.


3Hren:
@codex review

PR #1605, opened 2026-08-01
fix(web): cross-check the shared-memory inventory on the acl and route pages

3Hren:
A module control plane rebuilds its configuration map empty on startup, while the dataplane keeps its configurations in shared memory across that restart. The acl and route pages listed configurations only from the control plane, so after a restart they rendered a plain empty page and offered to create a first configuration while the dataplane was still forwarding under the old one.

- Combine the control plane's own configuration list with the shared-memory inventory the inspect API exposes, which is the source the decap and forward pages already trusted.
- Warn when the control plane knows none of the configurations that shared memory still holds, instead of rendering an empty page indistinguishable from a genuinely empty instance.
- Move the inventory lookup the decap and forward pages carried inline into a shared helper, so all four pages read it the same way.

A configuration the control plane no longer knows is still dropped from the view, and a module outage still fails the load outright rather than being misread as an empty instance. Creating a configuration stays available in the warned state, since re-applying one is how an operator recovers.

### DIFF
--- a/modules/acl/web/useAclDraft.ts
+++ b/modules/acl/web/useAclDraft.ts
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useReducer, useState } from 'react';
-import { API, loadKnownConfigs } from '@yanet/core/api';
+import { API, inventoryConfigNames, loadKnownConfigs, unionConfigNames } from '@yanet/core/api';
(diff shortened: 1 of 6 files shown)


3Hren:
@codex review

PR #1606, opened 2026-08-01
build(deb): narrow the binary package architectures to the supported hosts

3Hren:
- Nine binary packages declared an architecture wildcard that claimed support for every architecture Debian knows about, or, for the meta-package, for none in particular.
- The project in fact builds only on 64-bit x86 and little-endian ARM: the CRC-32C helper stops the compile with an explicit error on anything else, and the build system only knows how to derive ISA flags for those two families.
- Narrow every package to amd64 and arm64 so a build daemon no longer attempts an architecture the source cannot compile on, and so the meta-package is not advertised on hosts where the components it pulls in are never built.
- Of the twenty valid dpkg architectures whose CPU is x86-64 or aarch64, only these two are glibc/Linux; the rest are non-Linux ports, alternative libc ports, or the 32-bit-pointer x86-64 ABI, none of which can satisfy the build dependencies.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,7 +19,7 @@ Build-Depends: debhelper (>= 12)
  , python3-pyelftools
 
 Package: yanet2-dataplane
-Architecture: any
+Architecture: amd64 arm64
 Depends: ${shlibs:Depends}
  , ${misc:Depends}
  , libnuma1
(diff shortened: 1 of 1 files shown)


3Hren:
Addressed: the meta-package is now restricted to the same architecture list as the components it depends on, so it is no longer advertised on hosts where those are never built.

@codex review

PR #1607, opened 2026-08-01
docs(ai): route GitHub access through the MCP server

3Hren:
The agent instructions told every agent to shell out to the GitHub CLI, so results arrived as text to parse rather than as structured data. They now reach GitHub through the MCP server.

The CLI is kept only for the operations the server does not cover, and each one is named with its reason: waiting on checks, which has no continuous watch; the administrative merge bypass; endpoints with no tool; and refreshing authentication scopes.

Agents whose charter forbids GitHub writes are granted a read-only server, so those tools are absent rather than merely forbidden in prose. The grant topology differs between clients, so the shared instructions now state which client they describe rather than asserting one arrangement for both.

The publish runbook records two traps the change would otherwise have introduced. The server's file-write tools commit server-side and so never substitute for the local git flow, which owns staging discipline, the commit hook, and the verification gates. And the legacy commit-status view reports a permanently pending state for this repository, because every check reports as a check run instead.


3Hren:
@codex review

3Hren:
@codex review

PR #1608, opened 2026-08-01
fix(route): reject neighbours whose link has no usable source address

3Hren:
- A link without an Ethernet address of its own was given a fabricated all-zero hardware address, which then became the source MAC of every frame forwarded over it.
- Such a frame is not learnable and many switches drop it outright, so traffic left the box and disappeared with nothing logged and no counter moved.
- Neighbours discovered over a link that cannot supply a usable source address are now rejected and reported instead, leaving the nexthop unresolved where the existing metric already makes it visible.
- The loopback device reports six genuine zero bytes and falls into the same class, so it is rejected on the same grounds.
- The kernel table is now read through an injected source, so the resolution path is covered by a regression test.

### DIFF
--- a/operators/route/internal/discovery/neigh/neigh.go
+++ b/operators/route/internal/discovery/neigh/neigh.go
@@ -125,6 +125,16 @@ func WithOnHealthy(fn func()) Option {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1611, opened 2026-08-01
feat(dataplane): configurable numa-local worker thread stack

GeorgyKirichenko:
Add a per-worker stack size option to the dataplane configuration. When set, the worker thread stack is allocated from the worker instance's NUMA-local dataplane memory zone and backed by hugepages, keeping stack accesses on the same node as the pinned core. Leaving the size unset preserves the existing operating-system default stack.


GeorgyKirichenko:
Do not merge yet until I check how worth this is

3Hren:
Guards.

PR #1612, opened 2026-08-01
perf(filter): route net6 attributes through lpm_hash

GeorgyKirichenko:
Route the net6 half-classification through a byte-stride trie with a hashed top level. An open-addressing hash maps the leading hops bytes of an 8-byte half to the trie page those bytes lead to, so a hit walks only the remaining bytes and a miss falls back to the root walk, which by construction terminates within hops loads. The container is built once from the range partition the compiler already produces, entries are derived from the finished trie (one per page at the cut level), and lookup results are bit-identical to the plain walk.

- Generalize range_index_build_lpm into a range_index_build driving a per-range callback; the lpm and lpm_hash builders are thin wrappers over it.
- The hash side is a value_slot_index modelled on str_index: a self-contained open-addressing slot table with auto-grow, with all storage flowing through the memory_context so the structure is usable from shared memory.
- A hit skips the first hops dependent page loads; hops six cuts the dependent chain per packet roughly in half on the production acl ruleset across all four half-tries (a few hundred to a few thousand entries per trie).
- The shared net6 classification takes the same fast path: its remap arrays are built by walking the union range_index directly, since the container exposes no trie walk.
- Lookups are plain scalar lpm_hash_lookup calls in the net6 attribute queries and the acl dataplane: a hop-major batched walk measured 1.9..2.3x slower than the scalar loop on production traffic at -O2, so batching is not used.
- A self-contained benchmark under tests/common asserts equality with a full LPM control over biased and whole-space probe sets, then A/B-times the scalar lookup paths.

All filter suites, the full meson test set, and the acl Go suites pass, including the shared-classification functional tests.

### DIFF
--- a/common/lpm_hash.h
+++ b/common/lpm_hash.h
@@ -1,0 +1,484 @@
+#pragma once
+
+/*
+ * Hybrid byte-stride LPM with a hashed top level for 8-byte range maps.
+ *
+ * Lookup probes a hash on the first hops key bytes for the page number of
+ * the trie page those bytes lead to, then walks the remaining bytes from
+ * that page. A miss falls back to the root walk, which is short by
+ * construction. Values are bit-identical to lpm8_lookup on the same trie.
+ *
+ * Correctness of the shortcut: an entry exists for a prefix of hops bytes
+ * exactly when every slot on its path in hops 0..hops-1 holds a page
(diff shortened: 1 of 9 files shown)


GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

PR #1616, opened 2026-08-01
docs(common): fix the power-of-two alignment doc comment

3Hren:
- Removed stray keyboard input that had been left in the brief line of the power-of-two alignment helper's doc comment.
- Corrected the documented parameter name, which did not match the function signature.


3Hren:
@codex review

PR #1618, opened 2026-08-01
refactor(auth): pass the logger to authenticator factories through options

3Hren:
- Replace the positional logger argument of the three authenticator factories with the options pattern already used across the control plane.
- Drop the argument outright from the basic authenticator factory, which never logged anything.
- Let the two SSH factories accept their package option family, so a caller-supplied logger reaches the authenticator while configuration file settings keep taking precedence.
- Bind the factory registry to the manager logger where it is already in scope, which also removes the logger from the factory contract itself.
- Clear the three corresponding rows from the style ledger.

### DIFF
--- a/controlplane/internal/auth/basic/factory.go
+++ b/controlplane/internal/auth/basic/factory.go
@@ -3,17 +3,13 @@ package basic
 import (
 	"fmt"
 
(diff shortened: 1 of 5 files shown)


3Hren:
@codex review

PR #1619, opened 2026-08-01
fix(pdump,web): preserve sub-millisecond precision in the web pcap export

3Hren:
The browser built its exported pcap timestamps from the JavaScript date it renders in the packet list. That date carries whole milliseconds, so the export discarded everything below a millisecond and its fractional field was always a round number — the same capture exported from the command-line tool disagreed by up to a millisecond per packet.

The export now reads the dataplane timestamp directly instead of going through that date, and writes the nanosecond-resolution pcap variant the command-line tool already writes, so both exports of one capture carry the same format and agree.

The declared type of the timestamp field claimed a shape the gateway does not send; it now matches the other 64-bit wire fields in the API layer. The date shown in the packet list is unchanged and remains at millisecond resolution, which is what the interface renders.

The pcap encoding moved out of the page component into its own module so it could be covered by a unit test, which pins the timestamp arithmetic, the file header, the fallback for records without a dataplane timestamp, and the payload layout.

### DIFF
--- a/modules/pdump/web/PdumpPage.tsx
+++ b/modules/pdump/web/PdumpPage.tsx
@@ -16,6 +16,7 @@ import FilterRow from './FilterRow';
 import ConfigStrip from './ConfigStrip';
 import PacketDrawer from './PacketDrawer';
 import DeleteConfigDialog from './DeleteConfigDialog';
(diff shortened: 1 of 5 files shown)


3Hren:
@codex review

PR #1624, opened 2026-08-02
docs(gateway): describe the readiness marker as an external contract

3Hren:
- Restate the comment above the gateway's readiness block so it explains what the marker means and why the log message wording is a contract, without referring to tests or naming an external component.
- State that a run publishes at most one marker, since the waiting branch publishes none when the context is canceled before every runner has registered.
- Drop the same reference to tests from the documentation of the gRPC metrics factory option, which is a sibling of the same class in the same file.
- Comment-only change: no behaviour, no code, no renames.


3Hren:
@codex review

PR #1627, opened 2026-08-02
docs(lint): document the missing stylelint key forms and per-check rationale

3Hren:
The allowlist key format described in the style linter's package comment, and paraphrased in the allowlist header, enumerated four forms the name segment can take. Two real ones were missing.

- A file's own package clause name, which the internal test-package check emits. This is not hypothetical: dozens of live ledger rows are already in that shape, and a reader could not match any of them to the documented forms.
- A bare type name, emitted when a violation is found while walking a type declaration. No live rows today, so latent rather than urgent, but reachable.

Both are now documented in the package comment and in the allowlist header, which remain paraphrases of each other. The enumeration also glossed an interface method as a receiver-and-method pair, which it is not.

The package comment further claimed to carry each check's rationale, inlining it for a few and deferring the rest to their declarations, where several carried only a note on scope. Every one of the fourteen checks now states why the convention it enforces exists, so that deferral is honest. Where one check has two implementing functions, the rationale lives on one and the sibling points at it.

One check's scope note named a caller that only reaches it transitively; corrected to the real one.

No behaviour change: no check logic was touched and no allowlist row was added, removed, or reordered.

### DIFF
--- a/lint/style/allowlist.txt
+++ b/lint/style/allowlist.txt
@@ -7,15 +7,18 @@
 # where <check> is one of logger, private, testpkg, receiver, loopindex,
 # maplit, grpcdial, sugar, zapmsg, zapkey, testctx, handlerblank, barenew,
 # or loggerlast. <name> is the enclosing function or method
-# ("RecvType.MethodName" or "FuncName"), a struct field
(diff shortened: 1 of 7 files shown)


3Hren:
@codex review

PR #1628, opened 2026-08-02
chore(agents): classify planner items by kind and origin

3Hren:
Follow-up tasks and technical debt were only distinguishable by reading the prose provenance line on each tracker item, so there was no way to ask for the debt that appeared over some period and act on it.

- Give every task two orthogonal marks: `kind`, what the work is, and `origin`, how it surfaced. Keeping them separate is what makes both "fix the debt from yesterday" and "burn down what review left behind" answerable, which one combined field could not do.
- Put both marks, the date the item was raised, its effort and its repository in the planner's index, so the question is answerable from a single read rather than by opening every item file.
- Add a mode that returns the matching items ranked and grouped into batches that could each ship as one pull request, with a default window so the answer is a shortlist rather than a backlog dump.
- Order the two axes explicitly where several values fit, so the same evidence classifies the same way every time. A defect is never folded into a cleanup batch, whatever its confirmation state, since it belongs with the bug hunter instead.
- Deciding what actually starts stays with the architect: the new mode hands back a proposal and never moves an item into progress by itself, which would otherwise leave items looking active that nobody picked up.
- Correct the planner's stale claim that it tracks only the public repository. One tracker has covered both since late July, and the new mode acts on tracker contents in bulk, so an unmarked private item would be worked in the wrong checkout.

### DIFF
--- a/.claude/agents/architect.md
+++ b/.claude/agents/architect.md
@@ -140,14 +140,17 @@ Provides technical guidance on: RFC compliance, DPDK APIs, packet formats, proto
 Reviews code quality and verifies task completeness: conventions, safety, builds, tests.
 **Use after**: implementation is done, to verify everything works and meets standards.
 
-### `planner` — Multi-Horizon Planning Partner (read-only on code, public repo only)
+### `planner` — Multi-Horizon Planning Partner (read-only on code, both repos, public-first)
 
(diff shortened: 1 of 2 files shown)


3Hren:
@codex review

PR #1632, opened 2026-08-02
perf(counters): refcount counter_storage to skip rebuild for unchanged modules

### LINKED ISSUE #1631: perf(counters): refcount counter_storage to skip wrapper rebuild for unchanged modules
Every configuration generation install rebuilds the entire ectx tree, including spawning fresh counter_storage wrappers for every node — even though counter value pages are already COW-shared across generations via block refcounting (counters.c:389).

For a typical single-module config update, 99%+ of modules are unchanged (same cp_module pointer shared via registry refcounting). Their counter_registry pointer is identical, so the counter layout (names, sizes, offsets, handle resolution) is also identical. The wrapper rebuild allocates a counter_storage struct, 7 pool block-arrays, and a counter_value_handles array, then walks the registry to resolve every handle — all producing the same result as the old generation.

Add a refcnt field to counter_storage. In counter_storage_spawn, if the old counter_storage is non-NULL and its registry pointer matches the requested registry, bump refcnt and return the old storage directly. In counter_storage_free, decrement refcnt and skip reclamation until it reaches zero.

This eliminates the majority of per-node allocation and handle-resolution work during config updates while preserving identical semantics. No dataplane or ABI impact — the counter data pages were already shared.

GeorgyKirichenko:
Every config generation install rebuilds the entire ectx tree, including spawning fresh counter_storage wrappers for every node — even though counter value pages are already COW-shared.

For a typical single-module config update, 99%+ of modules are unchanged (same cp_module pointer shared via registry refcounting). Their counter_registry pointer is identical, so the counter layout is also identical. The wrapper rebuild is pure waste.

- Adds a refcnt field to counter_storage.
- In counter_storage_spawn, if the old storage registry pointer matches, bumps refcnt and returns the old storage directly.
- counter_storage_free decrements refcnt and skips reclamation until zero.
- Eliminates the majority of per-node allocation and handle-resolution work during config updates.
- No dataplane or ABI impact — counter data pages were already shared.

Closes #1631.

### DIFF
--- a/lib/counters/counters.c
+++ b/lib/counters/counters.c
@@ -331,6 +331,17 @@ counter_storage_spawn(
 	struct counter_storage *old_counter_storage,
 	struct counter_registry *counter_registry
 ) {
+	// Fast path: if the old generation's storage was built for the same
+	// counter_registry (same pointer — the registry item was shared by
+	// refcounting across generations), the counter layout is identical.
+	// Bump the refcount and return the old storage directly, skipping the
+	// wrapper rebuild, pool copy, and handle resolution.
+	if (old_counter_storage != NULL &&
+	    ADDR_OF(&old_counter_storage->registry) == counter_registry) {
+		old_counter_storage->refcnt += 1;
+		return old_counter_storage;
+	}
+
 	struct counter_storage *new_counter_storage = (struct counter_storage *)
 		memory_balloc(memory_context, sizeof(struct counter_storage));
 	if (new_counter_storage == NULL)
@@ -339,6 +350,8 @@ counter_storage_spawn(
 	counter_storage_init(
 		memory_context, new_counter_storage, counter_registry
 	);
+
+	new_counter_storage->refcnt = 1;
 
 	for (uint64_t pool_idx = 0; pool_idx < COUNTER_POOL_SIZE; ++pool_idx) {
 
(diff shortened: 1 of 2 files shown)


GeorgyKirichenko:
## Correctness review

The pointer comparison (`old_counter_storage->registry == counter_registry`) is a valid identity check: registry items are shared by pointer across generations via `registry_copy` (`registry.h:129-137`), so an unchanged module passes the same `&cp_module->counter_registry` address to both generations. Replaced modules have a different address, so the fast path is correctly skipped.

Counter data pages were already COW-shared via `block->refcnt` (`counters.c:389`), so the fast path only skips wrapper/handle reconstruction, not data sharing. The refcount correctly extends storage lifetime across the generation boundary: old gen free goes 2→1 (skip), new gen free goes 1→0 (reclaim). Per-worker isolation is preserved (fast path fires only between the same worker's old and new ectx).

### One latent code-quality concern

`refcnt = 1` is set in `counter_storage_spawn` (line 354), not in `counter_storage_init` (line 320). If a future caller uses `counter_storage_init` directly without going through `spawn`, `refcnt` would be 0 from memset, and `counter_storage_free` would underflow to `UINT64_MAX`, leaking the storage. Currently safe because `counter_storage_init` is only called from `counter_storage_spawn`. Moving the init to `counter_storage_init` would be more defensive.

PR #1635, opened 2026-08-03
fix(web): cross-check the shared-memory inventory on the fwstate page

3Hren:
A module control plane rebuilds its configuration map empty on startup, while the dataplane keeps its configurations in shared memory across that restart. The firewall-state page listed configurations only from the control plane, so after a restart it showed a plain empty page and offered to create a first configuration while the dataplane was still holding a live one.

The page now combines the control plane's own list with the shared-memory inventory the inspect API exposes, the same cross-check the acl and route pages already do. A configuration the control plane no longer knows is still dropped from the view, but the page now warns instead of pretending the instance is empty. Creating a configuration stays available, since re-applying one is how an operator recovers.

Verified in a browser against a stubbed backend: with an empty control-plane list and a non-empty inventory the page warns that the control plane does not know the configuration, and with both empty it still shows the plain empty placeholder and no warning.

### DIFF
--- a/modules/fwstate/web/FWStatePage.tsx
+++ b/modules/fwstate/web/FWStatePage.tsx
@@ -15,7 +15,7 @@ import {
 import { CircleInfo, Plus } from '@gravity-ui/icons';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 import { useConfigListCache, useSearchParamHelpers, usePageContribution, useContainerHeight, useTabCycle, useUnsavedChangesBlocker } from '@yanet/core/hooks';
-import { API, ApiError, loadKnownConfigs } from '@yanet/core/api';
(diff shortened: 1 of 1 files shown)


3Hren:
Accepted, and the local review reached the same conclusion independently. The partial-loss case is real: once at least one configuration survives, the all-dropped hook stays silent and an inventory-only configuration disappears without a warning.

The fix does not belong here, though. The hook fires from the shared configuration loader, and all five pages that combine a control-plane list with the shared-memory inventory inherit exactly the same blind spot, so fixing it inside this one page would leave the other four wrong and duplicate the logic. It also changes when every one of those pages warns, which is a behaviour decision of its own rather than part of extending the cross-check to this page.

Landing it as an immediate follow-up against the shared loader instead, so the warning covers a partial loss on all five pages at once.

PR #1644, opened 2026-08-03
chore(cli): require core over std across the rust workspace

3Hren:
Enable the `std_instead_of_core` clippy lint at deny level for the whole Rust workspace, so anything `core` provides is imported from `core` rather than `std`. The lint is declared once in the root manifest and inherited by every member through a `[lints]` table, which is the mechanism Cargo offers for workspace-wide lint policy.

The lint belongs to the restriction group, which `clippy::all` does not cover. Several generated protobuf wrappers already carried a blanket allow for `clippy::all` and were still emitting hits, so the suppression had to be named explicitly.

Every generated protobuf wrapper now carries that explicit allow, because the gRPC client codegen emits `std` paths that are not ours to change. All remaining occurrences are hand-written and were converted, covering error, formatting, networking, string-parsing, duration, memory, iterator and operator imports.

One case intentionally keeps `std`: a caller that reads the system clock, since the wall-clock types have no `core` equivalent while the address type alongside them does.

The Rust conventions document previously prescribed the exact `std` import the lint now rejects, so it is corrected and the new rule recorded alongside it.


3Hren:
@codex review

PR #1647, opened 2026-08-03
feat(unrdup): redistribute balancer ICMP errors to peer balancers

OrIX219:
When a backend answers with an oversized packet, a router returns an ICMP error addressed to the virtual service. With anycast it frequently lands on a balancer that holds no session for the flow, and the error is lost.

This adds an `unrdup` module that catches those errors and tunnels a copy to every peer balancer that may hold the session.

Configuration is a source network per family plus the services, each with its peers and the endpoints it serves:
```
source_v4: "10.0.0.0/30"
source_v6: "2001:db8:a::/96"
services:
  - vip: "192.0.2.1"
    peers: ["10.0.0.10", "2001:db8:b::11"]
    endpoints:
      - { port: 443, proto: tcp }
```
Whatever the source mask leaves free carries entropy from the address that sent the error, so a peer spreads the clones across its receive queues.

The session lookup is a TODO in the handler, blocked until balancer2 keeps a session table.


GeorgyKirichenko (dismissed):

GeorgyKirichenko:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
Still not decided how to mitigate fanout loop

modules/unrdup/dataplane/dataplane.c
```diff
@@ -0,0 +1,661 @@
...
+// ethernet padding or trailer begins.
+//
+// The version is checked here because the shared parser takes the header
+// layout from the ethernet type alone, so a frame naming one family and
+// carrying another would be read at the wrong offsets throughout.
+static int
+unrdup_outer_end(struct packet *packet, uint32_t *end) {
+	struct rte_mbuf *mbuf = packet_to_mbuf(packet);
+	uint16_t offset = packet->network_header.offset;
+	uint16_t data_len = packet_data_len(packet);
```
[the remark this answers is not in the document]
OrIX219: Not handling mbuf segments yet

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

modules/unrdup/fuzzing/unrdup.c
```diff
@@ -0,0 +1,192 @@
...
+	}
+	SET_OFFSET_OF(
+		&fuzz_params.module_ectx.counter_storage, counter_storage
+	);
+
+	*cp_module = &config->cp_module;
+	return 0;
+
+error:
+	memory_bfree(
```
moonug, line 192: **Unwind config allocations on setup failure**

After `unrdup_module_config_update_services()` succeeds, `config` owns the allocated service, peer, and port arrays. Any later failure in counter registration/linking or `counter_storage_spawn()` reaches this label and frees only the root object, leaving those nested allocations in `fuzz_params.mctx`. The initialized counter registry also needs its matching cleanup. The caller exits immediately, so this is limited to fuzz setup, but it still breaks the allocator-pairing contract and prevents leak checks from catching the problem. Please track the initialized stages and release them before freeing `config`.

modules/unrdup/api/controlplane.h
```diff
@@ -0,0 +1,68 @@
...
+	enum ip_family family;
+
+	const struct unrdup_peer_config *peers;
+	uint64_t peer_count;
+
+	const struct unrdup_port_config *ports;
+	uint64_t port_count;
+};
+
+// A failure leaves the module on its previous configuration.
```
moonug, line 61: The public API documents that a failed service update leaves the module on its previous configuration. The implementation builds the new service table before replacing the old one, but the allocation-failure path is not covered by a test.

If the existing test allocator supports fault injection, could we add a test that fails while allocating the new service, peer, or port data and verifies that:

- the update returns an error;
- all temporary allocations are released;
- the previous service table remains intact.

This would protect the atomic-publication guarantee against future changes to the C update path.

modules/unrdup/tests/dataplane/unrdup_test.c
```diff
@@ -0,0 +1,1475 @@
...
+			IPPROTO_TCP,
+			UNSERVED_PORT,
+			OFFENDING4_FULL
+		),
+		"build"
+	);
+	return assert_passed_through(&packet);
+}
+
+static int
```
moonug, line 540: UDP is currently covered only by a negative test: the configuration serves TCP, while the UDP packet is expected to pass through.

Could we add a positive UDP forwarding test where the configured endpoint uses `IPPROTO_UDP`, the quoted packet contains a UDP header, and the module produces a tunnel clone for the peer? This would cover the UDP parsing, protocol matching, and forwarding path, which is currently untested.

modules/unrdup/api/controlplane.c
```diff
@@ -0,0 +1,312 @@
...
+	if (unrdup_module_config_register_counters(&config->cp_module, err)) {
+		unrdup_module_config_free(&config->cp_module);
+		return NULL;
+	}
+
+	return &config->cp_module;
+}
+
+int
+unrdup_module_config_register_counters(
```
moonug, line 282: The module registers and updates eight counters, but the dataplane tests never assert their values.

Could we add assertions for the counters exercised by the existing fixtures, such as `redistributed`, `clones_sent`, `peer_no_source`, `unserved`, `misaddressed`, and `malformed`? The packet counters should verify both packet and byte counts.

For example, a single input error sent to two peers should distinguish one redistributed input packet from two successfully sent clones. Without these assertions, a regression in observability can pass while packet behavior tests remain green.

modules/unrdup/controlplane/service_test.go
```diff
@@ -0,0 +1,407 @@
...
+
+	_, err = service.UpdateConfig(t.Context(), request)
+	require.NoError(t, err)
+
+	require.Len(t, backend.handles, 2)
+	require.True(t, backend.handles[0].freed, "the replaced handle is released")
+	require.False(t, backend.handles[1].freed, "the live handle is kept")
+}
+
+func TestUpdateConfigBackendFailure(t *testing.T) {
```
moonug, line 365: `TestUpdateConfigBackendFailure` covers a backend failure only when no configuration has been stored yet. It does not verify the replacement failure contract.

Could we first install a valid configuration, then make the backend fail for the next update, and verify that:

- the update returns `Internal`;
- the previous handle is not freed;
- `ShowConfig` still returns the previous configuration.

This is important because `service.go:129-141` intentionally publishes the new module before freeing the previous one, so the existing configuration must remain active when the replacement fails.

modules/unrdup/tests/dataplane/unrdup_test.c
```diff
@@ -0,0 +1,1475 @@
...
+		rte_cpu_to_be_16(RTE_ETHER_TYPE_IPV4),
+		"network header type"
+	);
+
+	free_clones(&packet_front);
+	free_packet(&packet);
+	return TEST_SUCCESS;
+}
+
+static int
```
moonug, line 1016: `test_fanout_v6` builds the input packet with `build_icmp4`, so it verifies an IPv4 ICMP error being sent through an IPv6 tunnel. It does not verify fanout for an actual ICMPv6 input packet.

Could we add a separate test using `build_icmp6`, an IPv6 VIP, and an IPv6 peer? The test should verify the resulting tunnel protocol and ensure that the IPv6 error payload is preserved. This would cover the `IPPROTO_IPV6` branch in `dataplane.c:541-549`.

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

Cargo.toml
```diff
@@ -35,6 +35,7 @@ members = [
     "modules/pdump/cli",
     "modules/route-mpls/cli",
     "modules/route/cli/route",
+    "modules/unrdup/cli",
```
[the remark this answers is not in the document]
OrIX219: unrdup is exercised by the in-process harness and has no VM test

OrIX219:
@moonug review pls

GeorgyKirichenko (approved):

PR #1664, opened 2026-08-04
chore(gitignore): keep skills private by default

3Hren:
- Switch the skill directories from a blacklist of individually named private skills to a default-deny allowlist, so a new private skill is excluded automatically instead of needing a matching entry.
- Mirror the shape the repository already uses for out-of-tree modules, keeping the two policies expressed the same way.
- Invert which mistake is possible: forgetting an entry previously left a private skill committable, and now it makes adding a public skill fail loudly instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,11 +137,19 @@ deploy/packages/
 # checkout, and a directory-only pattern does not match a symlink.
 .claude/agent-memory
 .claude/settings.*.json
-.claude/skills/loadtest-hw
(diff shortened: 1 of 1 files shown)


3Hren:
The approach here changed after the first push: instead of naming each private skill in an ignore entry, the skill directories are now default-deny with the tracked public skills explicitly re-included, matching the shape already used for out-of-tree modules. @codex review

PR #1703, opened 2026-08-05
feat(cp): link modules to objects with per-worker link ectx

GeorgyKirichenko:
- A cp_module declares links to cp_objects by (type, name) via cp_module_link_object, mirroring the device linkage.
- At execution-context build time each link resolves to a module_object_link_ectx inside the module_ectx, owning a counter storage spawned from the linked object's dedicated relation counter registry and referencing that object's per-worker execution context.
- Links are reached at runtime through object_link_get_address, which returns NULL when the module has no link at that index.
- cp_object gains the dedicated relation counter registry alongside its own, with init, fini, and cross-generation linking.
- Bumps YANET_MODULE_ABI_VERSION for the cp_module and module_ectx layout changes.

Builds on the per-worker object ectx work (merged in #1702). Link counter storages spawn without old-storage reuse for now, so their values reset on a generation swap; tag-indexed reuse is a follow-up.

### DIFF
--- a/lib/controlplane/config/cp_counter.c
+++ b/lib/controlplane/config/cp_counter.c
@@ -475,10 +475,7 @@ cp_config_counter_storage_registry_lookup_module(
 	struct counter_tag tags[] = {
 		{.key = "device", .value = device_name},
 		{.key = "pipeline", .value = pipeline_name},
-		{
-			.key = "function",
-			.value = function_name,
-		},
+		{.key = "function", .value = function_name},
 		{.key = "chain", .value = chain_name},
 		{.key = "module_type", .value = module_type},
 		{.key = "module_name", .value = module_name},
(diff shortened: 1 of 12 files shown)


GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

The three open findings are addressed in c5b02e5f:

- Preserve link counters across generations: the link counter storage now reuses the prior generation's storage via `cp_config_counter_storage_registry_lookup_module_object_link` (`econtext.c`).
- Exclude link storages from module-only queries: `yanet_get_module_counters` now adds an `object_type` exclusion tag (`agent.c`).
- Set device_count before fallible object builds: already in place (`econtext.c`).

GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

PR #1792, opened 2026-08-07
perf(ci): install clang-tidy from the distro package

3Hren:
Stacked on #1791 — merge that one first. This pull request's base is its branch, so the diff shown here is only this change.

The job downloaded and unpacked a full upstream toolchain release, roughly one and a half gigabytes, in order to use a single binary from it. That cost about 116 seconds of a 205-second job and could not be cached, because the release asset alone exceeds the repository's remaining Actions cache allowance.

The distribution archive carries that binary at exactly the version the download was pinned to, in the updates and security pockets. So the download step is deleted outright and the package is added to the list the job already installs, leaving the job with one step fewer than before this stack and two fewer than the branch it targets.

Because the version is identical, this raises no question of the linter's findings drifting.

## What else falls out

The invocation loses everything that existed only to reach into an unpacked tree: a directory variable, an existence check, two environment exports, and an indirection through a variable holding the binary's path.

It also loses a branch that would have run an upstream helper tool if that tool were found. Two job logs show it never ran — the marker it would print appears only where the step's own script is echoed, never as output — because the release tarball installs that helper under a different name than the branch looked for. The packaged toolchain does ship it, under the name the branch would find, so keeping the branch would have quietly moved linting onto a path this repository has never exercised. The proven path stays and the dead one goes.

## Coherence, and one cost worth naming

The result is that continuous integration now runs the same binary the repository's own local lint recipe runs, and installs it the same way the development container image and the functional-test virtual machine already do. The deleted download step was the only place doing it differently.

The cost: adding a package changes that install step's cache key, so it mints a fresh entry of roughly seventy to ninety megabytes compressed, while the previous, much smaller one lingers until it ages out. The Actions cache is close to its ceiling, so this will bring forward the eviction of something else. It is not the driver of that pressure — a substantial share of the cache is duplicate entries under a handful of keys — but it is not free either, and the duplication deserves its own change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -282,20 +282,13 @@ jobs:
       cache_name: clang-tidy
 
     steps:
-      - name: Install LLVM 19 toolchain
-        id: llvm
-        uses: KyleMayes/install-llvm-action@v2
-        with:
-          version: "19.1.1"
-          directory: ${{ runner.temp }}/llvm
-
       - uses: actions/checkout@v4
         with:
           submodules: true
 
       - uses: ./.github/actions/apt-packages
         with:
-          packages: meson clang cmake python3-pyelftools libnuma-dev libpcap-dev git protobuf-compiler jq
+          packages: meson clang clang-tidy-19 cmake python3-pyelftools libnuma-dev libpcap-dev git protobuf-compiler jq
 
       - uses: actions/setup-go@v5
         with:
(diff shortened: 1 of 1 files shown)


3Hren:
Closing unmerged: the job this speeds up is being removed instead.

Measuring it to justify the speedup is what settled that. The job emits 233 warnings and reports success, so it enforces nothing. All 233 come from a single enabled check, 208 of them from one test file. Its header filter uses a lookahead the regex engine cannot compile, so headers have never been linted at all, and the braces check sits under the options block rather than the check list, so it has never been enabled either. Nothing consumes the output — no artifact, no annotation, no required check. What remains is an expensive parse of every C file, which the build jobs already prove by compiling.

This branch also demonstrated a real cost. Adding the package to the cached apt composite rather than installing it directly made that step run for 51 minutes, because the caching action tars the whole toolchain closure package by package. That was my error, not the reviewer's — a plain install was the recommendation.

PR #1794, opened 2026-08-07
chore(ci): remove the clang-tidy job

### LINKED ISSUE #1783: perf(ci): stop clang-tidy paying for an uncached LLVM install and an unused Go build
## Motivation

The `clang-tidy` job is 5.8 min end to end (avg over the 12 most recent successful `pull_request` runs), but the step-level replay of run 28576043639 shows the actual lint takes 20 s of that:

| Phase | s |
| --- | --- |
| LLVM 19 toolchain install | 116 |
| checkout + apt packages + ccache setup | 43 |
| meson setup | 18 |
| ninja C targets | 17 |
| ninja Go targets (never read by this job) | 104 |
| clang-tidy itself | 20 |

[build.yml#L285-L332](https://github.com/yanet-platform/yanet2/blob/6bf3c29ff7f481305f8d6b141915bfd609a16fc9/.github/workflows/build.yml#L285-L332): the LLVM install step (`KyleMayes/install-llvm-action`) carries no cache configuration, so it downloads and unpacks the toolchain fresh on every run. `make dataplane` at [build.yml#L334-L338](https://github.com/yanet-platform/yanet2/blob/6bf3c29ff7f481305f8d6b141915bfd609a16fc9/.github/workflows/build.yml#L334-L338) builds both C and Go targets, but the `clang-tidy` step below it (`build.yml#L340-L398`) only ever reads `build/compile_commands.json` and `.c` files — the 104 s of Go compilation this job pays for is never consumed.

220 of the job's 348 measured seconds (63%) is toolchain install plus dead Go compilation, on a job that does 20 s of real work.

## Acceptance

- [ ] LLVM 19 toolchain install is cached across runs (action-native cache or an `actions/cache` step keyed on the pinned version).
- [ ] The job no longer builds Go targets it never reads — either compile C-only, or otherwise avoid `make dataplane`'s Go build.
- [ ] `clang-tidy` job wall clock re-measured against the 5.8 min baseline.

Source: deferred remainder of the CI performance review behind PR #1775 (`perf(ci): stop rebuilding Go twice and shrink the actions cache`), measured 2026-08-07. Tracked under #1777.

3Hren:
Removing the job rather than continuing to optimise it. Measuring it in order to justify a speedup is what showed there was nothing to speed up.

## What it actually did

On a real run it emitted 233 warnings and reported success, because nothing turns warnings into failures. Every one of those came from the single check the configuration enables, and 208 of them from one test file. Its header filter is a negative lookahead, which the regex engine behind the tool cannot compile, so no header has ever been inspected. A second check was written under the options block instead of the check list, so it was never enabled either.

Nothing consumed the result: no artifact, no annotation, no other workflow, and no merge rule. The repository's branch ruleset has no required-status-check rule at all, so the job could never block anything.

What remained was an expensive parse of every C file, which the build jobs already prove by compiling them.

## What stays

The configuration file and the local recipes are untouched, as is the package in the development image and the test virtual machine, so the tool remains available to run on demand. Only the continuous-integration job goes.

## A coverage consequence worth naming

This job was one of two places where continuous integration parsed the whole C tree with a different compiler from the one the build jobs use. The other never runs on a pull request. So a front-end break specific to that compiler now surfaces at most nightly rather than on the pull request that introduces it.

That is not an argument for keeping this job, which only ever caught front-end errors and enforced none of its findings, but it is a real change in what continuous integration notices and is worth deciding about separately.

Closes #1783.


3Hren:
Rebased onto current `main`. The branch conflicted with #1799, which migrated `actions/cache@v3` to `@v4` across this workflow including one line inside the job being deleted. The resolution keeps every surviving `@v4` line from `main` and removes only the job, so the diff is still a pure 110-line deletion with no insertions.

Also corrected two imprecisions in the description above: the job configured its build with the same compiler as the build jobs and only ran a front end over the tree, so it parsed rather than compiled it, and the other clang site is reachable by manual dispatch as well as by schedule.

@codex review

PR #1803, opened 2026-08-07
chore(ci): repair the dead cache restore-key ladders in the deb build path

### LINKED ISSUE #1800: chore(ci): build-deb Go cache restore-key ladder is dead by construction
## Motivation

`build-deb`'s Go module cache has a `restore-keys` fallback that can never match, so every cache miss on the exact key is a full cold miss instead of a partial restore.

[.github/workflows/build-deb.yml#L106-L112](https://github.com/yanet-platform/yanet2/blob/2e762d45a84680b89231d0cc8c0d7d4b00515eec/.github/workflows/build-deb.yml#L106-L112) sets the primary key to `${{ env.cache_name }}-${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-${{ hashFiles('**/go.mod') }}-${{ matrix.unbtver }}` but its only `restore-keys` entry to `${{ env.cache_name }}-${{ runner.os }}-go-${{ matrix.unbtver }}` — i.e. `deb-build-24.04-Linux-go-24.04`. `actions/cache` restore-keys match by prefix, and `…-go-24.04` can never be a prefix of `…-go-<hash>-<hash>-24.04`, since the hashes sit strictly between the shared `go-` prefix and the trailing `-24.04`. The fallback can therefore never fire.

[.github/workflows/build-push-images.yml#L57-L64](https://github.com/yanet-platform/yanet2/blob/2e762d45a84680b89231d0cc8c0d7d4b00515eec/.github/workflows/build-push-images.yml#L57-L64) has the identical defect and deliberately shares the same `deb-build-24.04-...` key namespace with `build-deb.yml`.

Contrast with the four `build.yml` jobs, which use a correct two-level ladder and score exact primary-key hits: [.github/workflows/build.yml#L78-L81](https://github.com/yanet-platform/yanet2/blob/2e762d45a84680b89231d0cc8c0d7d4b00515eec/.github/workflows/build.yml#L78-L81) restores from `<name>-<os>-go-<go.sum hash>-` (partial) before falling back to `<name>-<os>-go-` (bare). `build-deb` needs the equivalent ladder: `deb-build-<ver>-<os>-go-<go.sum hash>-` then `deb-build-<ver>-<os>-go-`.

This is pre-existing and independent of the `actions/cache@v3`→`v4` migration in #1786/#1799 — it is reproduced identically on both major versions:

- Post-bump (v4), PR #1799's `build-deb (24.04)` job (run 92957374100) logged `Cache not found for input keys: deb-build-24.04-Linux-go-<h1>-<h2>-24.04, deb-build-24.04-Linux-go-24.04`, then saved under the primary key.
- Pre-bump (v3), the identical miss on the identical key pair appears on run 31194827824 (job 92920425671, branch `chore/drop-clang-tidy-job`, still on `actions/cache@v3`).

`build-deb (24.04)` is the longest job in CI at roughly 14 minutes, and it re-downloads the whole Go module cache on every `go.mod`/`go.sum` change instead of restoring the previous one's modules and only re-fetching the delta.

## Acceptance

- [ ] `build-deb.yml`'s Go cache `restore-keys` includes a partial-hash rung (`…-go-${{ hashFiles('**/go.sum') }}-`) ahead of the bare `…-go-` rung, matching `build.yml`'s ladder shape.
- [ ] `build-push-images.yml`'s Go cache gets the same fix, since it shares the `deb-build-24.04-...` key namespace.
- [ ] A run after the fix, on a `go.mod`/`go.sum` change, restores from the partial-hash rung instead of missing cold (verified from job logs).

Source: found while verifying PR #1799 (the v3→v4 cache-action bump that fixed #1786); the defect is independent of that migration and reproduced identically on the pre-bump v3 run.

3Hren:
The deb and image build paths appended the Ubuntu version after the dependency hashes in their Go and Cargo cache keys. Restore keys match by prefix, and the hashes sit strictly between the shared prefix and that trailing version, so no fallback rung could ever be formed. Every miss on the exact key was therefore a full cold miss instead of a partial restore, on the longest job in CI.

Fixed by applying one rule at all four sites: the variable-length hash components come last, and any discriminator sits ahead of them. Each cache then gets the same two-level fallback the main build workflow already uses.

- Moved the Ubuntu discriminator ahead of the hashes in both Cargo caches, keeping the two Ubuntu releases partitioned so they never share a target directory.
- Dropped the redundant trailing Ubuntu version from the Go cache key in the deb workflow, since the cache name already carries it.
- Added the missing fallback ladder to the image workflow's Cargo cache, and aligned its key with the deb workflow so the two genuinely share one namespace as intended.

Also removes a pre-existing hazard in the opposite direction: the release build's bare Cargo fallback could previously restore a target directory built on a different Ubuntu release.

Verified by rendering every cache key and rung in the repository and asserting each rung is a real prefix of its key, for both Ubuntu values, plus a cross-entry collision matrix in both directions. The same check reports the four sites as dead before the change.

Closes #1800.

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -84,10 +84,10 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('**/Cargo.toml') }}-${{ matrix.unbtver }}
+          key: ${{ runner.os }}-${{ matrix.unbtver }}-cargo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('**/Cargo.toml') }}
           restore-keys: |
-            ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-${{ matrix.unbtver }}
-            ${{ runner.os }}-cargo-${{ matrix.unbtver }}
+            ${{ runner.os }}-${{ matrix.unbtver }}-cargo-${{ hashFiles('**/Cargo.lock') }}-
+            ${{ runner.os }}-${{ matrix.unbtver }}-cargo-
 
       - name: Install packaging tools
         run: |
@@ -107,9 +107,10 @@ jobs:
         with:
           path: |
             ${{ env.modcache }}
-          key: ${{ env.cache_name }}-${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-${{ hashFiles('**/go.mod') }}-${{ matrix.unbtver }}
+          key: ${{ env.cache_name }}-${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-${{ hashFiles('**/go.mod') }}
           restore-keys: |
-            ${{ env.cache_name }}-${{ runner.os }}-go-${{ matrix.unbtver }}
+            ${{ env.cache_name }}-${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-
+            ${{ env.cache_name }}-${{ runner.os }}-go-
       - name: Install Go Protobuf Plugins
         run: |
           go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
--- a/.github/workflows/build-push-images.yml
+++ b/.github/workflows/build-push-images.yml
@@ -36,7 +36,10 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-24.04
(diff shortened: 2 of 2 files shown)


3Hren:
Acceptance evidence from this PR's own `build-deb (24.04)` run (job 92983269677).

The Go cache fell back through the new ladder and restored from the bare rung, matching an entry left under the old key namespace:

```
Cache hit for restore-key: deb-build-24.04-Linux-go-<gosum>-<gomod>-24.04
Cache restored from key:   deb-build-24.04-Linux-go-<gosum>-<gomod>-24.04
```

The same situation before the change produced a full cold miss, because both listed keys were unreachable by prefix:

```
Cache not found for input keys: deb-build-24.04-Linux-go-<gosum>-<gomod>-24.04, deb-build-24.04-Linux-go-24.04
```

So a fallback rung fired here for the first time. The Cargo cache correctly reports a one-time cold miss, since its key namespace moved and nothing had been written under it yet.

Issue #1801, opened 2026-08-07
chore(filter,lpm): name shared-memory contexts distinctly so per-subsystem metrics can attribute them


3Hren:
## Motivation

[`lib/filter/compiler.h`](https://github.com/yanet-platform/yanet2/blob/2c0d1f99960a0f0360cd59deba3a74864edafe52/lib/filter/compiler.h#L144) names every filter memory context with the hardcoded literal `"filter"`, and [`common/lpm.h`](https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/common/lpm.h#L125) does the same with `"lpm"`. Modules routinely build several of these under one module memory context, so the memory-context tree contains multiple sibling nodes that are indistinguishable by name:

- [`modules/acl/api/controlplane.c`](https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/modules/acl/api/controlplane.c#L348) creates five filter contexts under one module context (`filter_vlan`, `filter_ip4`, `filter_ip4_port`, `filter_ip6`, `filter_ip6_port`).
- [`modules/forward/api/controlplane.c`](https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/modules/forward/api/controlplane.c#L194) and [`modules/mirror/api/controlplane.c`](https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/modules/mirror/api/controlplane.c#L189) each create three.
- [`modules/route-mpls/api/controlplane.c`](https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/modules/route-mpls/api/controlplane.c#L195) creates two.
- [`modules/route/api/controlplane.c`](https://github.com/yanet-platform/yanet2/blob/e211334a99d1b54e5696aaa8b99ac04d99184686/modules/route/api/controlplane.c#L131) creates two lpm contexts (`lpm_v4`, `lpm_v6`) from the same parent; [`modules/nat64/api/nat64cp.c`](https://github.com/yanet-platform/yanet2/blob/d17d03b634ceea7b10c9bc7127f60a4b42480bc7/modules/nat64/api/nat64cp.c#L100) creates three (`v4_to_v6`, `v6_to_v4`, `v6_prefixes`).

[...]

PR #1810, opened 2026-08-07
feat(controlplane): expose per-agent shared-memory arena utilization as metrics

### LINKED ISSUE #1731: feat(controlplane): expose per-subsystem shared-memory utilization as a metric
## Motivation

`struct memory_context` (`common/memory.h`) already tracks `balloc_count`/`bfree_count`/`balloc_size`/`bfree_size` per node, so a node's live shared-memory usage is derivable as `balloc_size - bfree_size`. The memory-context tree already carries `parent`/`first_child`/`next_sibling` links, landed by an earlier merged change, so the tree structure needed to walk an agent's whole allocation graph already exists in main. What is missing is any surface that actually walks that tree and reports utilization: today there is no signal at any layer for "how full is the shared-memory arena," even though shared-memory exhaustion is a real failure mode — route tables grow with the global BGP table, ACL rulesets grow, and an agent's `memory_limit` can be approached or hit with no visibility into which subsystem is consuming it.

The blocking prerequisite this task originally identified — that both `memory_context_init_from` and `memory_context_fini` mutate the parent/child/sibling links without any lock, so a scrape-time tree walk could race a concurrent module config build or free — has since been resolved generically: an allocator-internal lock introduced by a separate merged change now serializes those operations across every module's build/free path, not just one module's. That lock's coverage should be re-verified against every module's actual build/free call sites before being fully relied on, rather than assumed complete because one module's changes exercised it.

This is distinct from a related but separate concern already tracked elsewhere: moving a specific subsystem's storage into agent-scoped memory in the first place (allocation *placement*) is not the same question as reporting utilization of whatever is already there (allocation *visibility*), and the two should not be conflated.

## Scope

- A generic surface (a shared Go helper, or an FFI walk) that walks an agent's guarded `memory_context` tree once and emits `balloc_size - bfree_size` per subsystem, labeled by node name — not hidden inside a single module's own metrics surface.
- Exposed as a reusable metric type (or wired directly into a future Prometheus-style exporter) so any module or the platform layer can surface it, not a one-off per-module counter.

## Constraints

- Re-verify that the generic allocator lock actually covers every module's build/free call sites, not only the call sites exercised by the change that introduced it, before treating the concurrency prerequisite as fully discharged.
- A separate, already-merged inspect-API change introduces its own tree walk at the CLI/web layer; before building a new walk, confirm whether that surface is reusable here or whether a distinct, metrics-oriented walk is still needed.

## Acceptance

- [ ] The concurrency prerequisite (safe tree walk under concurrent module build/free) is confirmed covering every module's build/free path, not just a sampled subset.
- [ ] A generic surface walks an agent's memory-context tree once and emits per-subsystem utilization, reusing the existing inspect-API walk if it fits, or a dedicated one if it doesn't.
- [ ] The utilization figures are exposed as a reusable metric surface (not module-specific), ready for a future exporter to pick up.

Source: packet-path safety backlog migration; originally surfaced while designing a module's own metrics-service inventory and needing a cross-module answer to "how full is the arena."

3Hren:
Shared-memory exhaustion is a real failure mode that had no signal at any layer. Route tables grow with the global BGP table and ACL rulesets grow, so an agent can approach its memory limit with nothing reporting how full the arena is or which subsystem is consuming it.

- Add a platform-level memory service that reports arena occupancy and per-subsystem attribution for every agent of the instance, reusing the memory-context tree the info API already exports rather than introducing a second walk.
- Emit the figures through the gateway's existing metrics service, so the generic metrics command surfaces them alongside the server metrics with no client change.
- Serve an arena listing over the new service as well, giving a planned arena grow and shrink API its read side. It deliberately omits the context tree, which the inspect API already carries.

Occupancy and attribution are separate metric families on purpose. The allocator rounds every request up to a pool size and an out-of-order teardown can drop a live subtree from the walk, so the free-byte count is authoritative for how full an arena is, while the context tree only attributes what is inside it.

Attribution is keyed by each context's full path from the root. An agent holds one context per named module configuration, so a name-and-parent key collapsed same-named leaves from different configurations onto one series and made the most useful question unanswerable. Contexts that still resolve to an identical path are summed, because sibling contexts can share a hardcoded name and names truncate to a fixed length, and a duplicate label set costs a scraper the entire scrape rather than just the affected series. Naming those contexts distinctly is tracked in #1801.

Arena figures cover the live generation, with superseded generations reported separately so a stalled reclamation stays visible. Labels are limited to the agent and the context path, keeping the series set bounded by configuration.

Closes #1731.

### DIFF
--- a/controlplane/builtin/memory.go
+++ b/controlplane/builtin/memory.go
@@ -1,0 +1,152 @@
+package builtin
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+
+	"github.com/yanet-platform/yanet2/common/commonpb/v1"
+	"github.com/yanet-platform/yanet2/controlplane/ffi"
+	ynpb "github.com/yanet-platform/yanet2/controlplane/ynpb/v1"
+)
+
+// Memory is an in-process gRPC service for reporting per-agent
+// shared-memory arena utilization.
+type Memory struct {
+	ynpb.UnimplementedMemoryServiceServer
+
+	instanceID uint32
+	shm        *ffi.SharedMemory
+}
+
+// NewMemory creates a new Memory service.
+func NewMemory(instanceID uint32, shm *ffi.SharedMemory) *Memory {
+	return &Memory{
+		instanceID: instanceID,
+		shm:        shm,
+	}
+}
+
+// Name returns the service name.
+func (m *Memory) Name() string { return "memory" }
+
+// Endpoint returns empty string indicating in-process service.
+func (m *Memory) Endpoint() string { return "" }
+
+// ServicesNames returns the gRPC service names served by this service.
+func (m *Memory) ServicesNames() []string { return []string{"controlplane.ynpb.v1.MemoryService"} }
+
+// RegisterService registers the service on the given gRPC server.
+func (m *Memory) RegisterService(server *grpc.Server) {
+	ynpb.RegisterMemoryServiceServer(server, m)
+}
+
+// ListArenas returns arena utilization for every agent generation.
+func (m *Memory) ListArenas(
+	ctx context.Context,
+	request *ynpb.ListArenasRequest,
+) (*ynpb.ListArenasResponse, error) {
+	dpConfig := m.shm.DPConfig(m.instanceID)
+	agents := dpConfig.Agents()
+
+	arenas := make([]*ynpb.ArenaInfo, 0)
+	for _, agent := range agents {
+		for genIdx, instance := range agent.Instances {
+			arenas = append(arenas, &ynpb.ArenaInfo{
+				Agent:       agent.Name,
+				Generation:  instance.Gen,
+				Pid:         instance.PID,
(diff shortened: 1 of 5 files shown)


3Hren:
Force-pushed a rename: the four metrics drop their `agent_` name prefix, since the `agent` label already carries that dimension. No other change.

@codex review

3Hren:
Parking this as a draft rather than merging. CI is green across all 16 checks and the review found nothing, so it is mergeable whenever we want it — the hold is deliberate.

The open question is ergonomic, not correctness: whether these four series are actually pleasant to query and alert on against a real workload. That is best answered by putting them on the stand and looking at the output, and the answer may well argue for reshaping the labels or the split between occupancy and attribution before any of it becomes a contract someone builds a dashboard on.

Marking ready for review again will re-trigger a fresh review round, which is what we want after any reshaping.

Issue #1782, opened 2026-08-07
chore(ci): decide whether build-deb gets a warm base-branch compiler cache


3Hren:
## Motivation

`build-deb.yml` has no `push: branches` trigger — over its last 100 runs: 96 `pull_request`, 4 `workflow_dispatch`, 0 `push` ([build-deb.yml#L3-L18](https://github.com/yanet-platform/yanet2/blob/96fb8bfd8cf3719fe0e5ece257b5a49e1a587f9c/.github/workflows/build-deb.yml#L3-L18)). GitHub Actions caches are ref-scoped, so a cache written on a `refs/pull/N/merge` ref is invisible to every other pull request. Of the job's 7 ccache entries, 6 were created on pull-request refs. Cross-PR restore therefore almost never hits, and this is the dominant cause of the job's 19.5% overall ccache hit rate (separate from the 0%-direct-mode issue tracked in #1776's other children).

Every other heavy job in `build.yml` gates its cache *write* to `github.ref == 'refs/heads/main'` (PR #1775) precisely so a warm base-branch entry exists for pull requests to restore from. `build-deb` cannot use that same gate as-is, because it never runs on a main push at all — a main-gated write would simply never fire, and because the cache is ref-scoped, the workflow would stay stranded on whatever stale entry it last wrote from a PR ref.

This has a real, non-free cost to weigh: giving `build-deb` a warm cache means either running it (or an equivalent compile) on every push to main, adding a full 15-minute job to every merge, or restructuring what it caches from.

## Acceptance

- [ ] A decision is recorded: run `build-deb` (or a cache-warming equivalent) on main pushes to seed a warm cache, or explicitly accept the current cold-cache cost as a rejected trade-off.
- [ ] If warming is chosen: implemented, and the ccache hit rate re-measured against the 19.5% baseline.
- [ ] If rejected: the reason and the measured cost of running it on every merge are recorded here.

Source: deferred remainder of the CI performance review behind PR #1775 (`perf(ci): stop rebuilding Go twice and shrink the actions cache`), measured 2026-08-07. Tracked under #1776.

3Hren:
Two local measurements to weigh in the decision here, both on Ubuntu 24.04 / gcc 13.3.0 / ccache 4.9.1 (the exact CI version) at `-j4` to match the 4-vCPU runners — from the DPDK-artifact-cache evaluation closed as #1835:

| Scenario | ninja wall at `-j4` | DPDK CPU time |
[...]

PR #1834, opened 2026-08-08
perf(ci): stop the deb build poisoning its own ccache keys

### LINKED ISSUE #1781: chore(ci): diagnose why build-deb's ccache has zero direct-mode hits
## Motivation

`build-deb`'s ccache ([build-deb.yml#L59-L65](https://github.com/yanet-platform/yanet2/blob/96fb8bfd8cf3719fe0e5ece257b5a49e1a587f9c/.github/workflows/build-deb.yml#L59-L65)) sees 2486 cacheable calls at a 19.5% hit rate. Of the 162 hits it does get, 0.0% are direct-mode — every single one is a slower preprocessed-mode hit. `build-release-artifacts`, `build-asan-test` and `build-tsan-test` all sit at 86-98% direct-hit rates on the same ccache action version in the same repository, so the 0% is specific to this job, not the tool.

Unexplained. Suspected causes, neither confirmed: compile-flag divergence from the packaging hardening options ([debian/rules#L8-L12](https://github.com/yanet-platform/yanet2/blob/a11373cfa76696e74573d87aa4dcf2bfda764207/debian/rules#L8-L12), `DEB_BUILD_MAINT_OPTIONS = hardening=+all` plus `-fno-lto`), or the ccache configuration not resolving correctly under the `sudo -E` invocation in the build step ([build-deb.yml#L115-L118](https://github.com/yanet-platform/yanet2/blob/96fb8bfd8cf3719fe0e5ece257b5a49e1a587f9c/.github/workflows/build-deb.yml#L115-L118)).

Direct mode skips preprocessing entirely on a hit; recovering it would cut real compile time inside `dh_auto_build`'s 473 s ninja/cargo phase, the single largest phase in `build-deb`.

## Acceptance

- [ ] Root cause identified (ccache config, `sudo` environment, or hardening-flag divergence) with evidence from a real run, not inference.
- [ ] A fix is proposed and its expected direct-hit-rate improvement is stated, or the cause is recorded as unfixable with why.

Source: deferred remainder of the CI performance review behind PR #1775 (`perf(ci): stop rebuilding Go twice and shrink the actions cache`), measured 2026-08-07. Tracked under #1776.

3Hren:
- `build-deb`'s ccache produced zero direct-mode hits — 0 of 1605 direct lookups — and a 20% overall hit rate, against 77-98% direct in the three build jobs using the same action version. Two independent causes, both confirmed by reproducing the exact signature locally rather than by inference.
- Ubuntu's fixfilepath feature maps the build directory into the DWARF source path and stamps it with the package version, which the CI build script makes unique for every commit. ccache folds that mapped directory into both its manifest key and its result key, so every run computed fresh keys and the restored cache was dead on arrival. Pinning the debug path keeps it absolute, as DWARF requires for the compilation directory, while dropping the version.
- The ccache action's compiler-symlink option additionally routed Go cgo compiles through ccache. cgo names synthetic pseudo-files in its preprocessed output, ccache cannot stat them, and each such compile paid a full wasted preprocess before falling back to the real compiler — 782 errors, 32% of all invocations. Meson already prefixes ccache on its own compile rules whenever ccache is on the path, so the symlinks bought the C build nothing, and the three healthy jobs set no such option either.
- Measured on a local A/B in which only the changelog version differed between two runs: direct hits went from 0 of 1110 to 527 of 1110, and the overall hit rate from 11.8% to 93.5%. That is a floor taken after a single populating run, so the steady state should approach the rate the other jobs already reach.
- Only the 24.04 leg is affected. Jammy's dpkg has neither the debug-path override nor the versioned path, so the export is inert there and there was nothing to fix.
- Verifiable from the ccache action's own verbose summary on the second run after this lands: the error block should collapse to roughly two, and direct hits should be non-zero.

Closes #1781.

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -65,7 +65,6 @@ jobs:
           key: ${{ runner.os }}-deb-ccache-${{ matrix.unbtver }}
           max-size: 2G
           verbose: 2
-          create-symlink: true
 
       - name: update apt (act hack)
         if: ${{ env.ACT }}
--- a/debian/rules
+++ b/debian/rules
@@ -11,6 +11,16 @@ export DEB_LDFLAGS_MAINT_APPEND = -fno-lto
 # Enable all hardening options
 export DEB_BUILD_MAINT_OPTIONS = hardening=+all
 
+# Pin the DWARF source path so it does not carry the package version.
+#
+# Ubuntu's fixfilepath feature maps the build directory to
+# /usr/src/<source>-<version>, and the CI build script gives every commit a
+# distinct +dev version. ccache hashes that mapped directory into both its
+# manifest and result keys, so a version-bearing path makes every key unique
+# and no cache entry is ever reused. Turning the feature off instead would
+# leave a relative path, which DWARF forbids for DW_AT_comp_dir.
+export DEB_BUILD_DEBUGPATH = /usr/src/yanet2
+
 # Set parallel build options
 export DEB_BUILD_OPTIONS = parallel=$(shell nproc)
 


3Hren:
Measured on this PR's own `build-deb (24.04)` runs, against the baseline job cited in #1781 (92976317468).

| | baseline | this PR, cold cache | this PR, warm cache |
|---|---|---|---|
| ccache invocations | 2414 | 683 | 683 |
| Errors — "could not read or parse input file" | 782 (32.39%) | 2 (0.29%) | 2 (0.29%) |
| Called for preprocessing | 152 | 0 | 0 |
| Called for linking | 590 | 2 | 2 |
| Overall hit rate | 163/810 (20.12%) | 18/665 (2.71%) | 617/665 (92.78%) |
| Direct-mode lookups | **0 / 1605 (0.00%)** | 0 / 680 | **526 / 680 (77.35%)** |
| "Build packages" step | 914 s | 581 s | 460 s |

- Dropping the compiler symlinks is visible immediately, on the cold run: invocations fall by 72% and the error block collapses from 782 to 2. The 152 "called for preprocessing" going to zero also confirms the job was running two nested ccache processes per compile, since meson prefixes ccache on its own rules and the symlink made the compiler it invoked ccache again.
- Direct mode reaches 77.35%, which lands exactly on the 77.5% that `build-release-artifacts` already scores — the healthy figure this job never reached.
- One honest caveat on what the warm column proves. It is a re-run of the same commit, so the package version is identical across the two runs and it does not by itself isolate the version from the debug path. That isolation comes from the local A/B in which the changelog version was the only variable (0 of 1110 direct before, 527 of 1110 after) together with ccache folding the mapped directory into both key derivations. Cross-PR warmth remains gated on #1782, since this workflow never runs on main and its caches stay ref-scoped.

PR #1841, opened 2026-08-08
perf(build): keep the build stamp out of the widely included config header

### LINKED ISSUE #1837: chore(build): per-second build timestamp caps ccache's ceiling and makes packages unreproducible
## Motivation

[meson.build#L52-L53](https://github.com/yanet-platform/yanet2/blob/545d22affa68e6c277ad92277ba60e652f8c2f68/meson.build#L52-L53) runs `date -u +%Y-%m-%d %H:%M:%S` at configure time and bakes the result into `yanet_build_config.h` as `YANET_BUILD_DATE`, at second granularity, regenerated on every `meson setup`.

[lib/logging/log.h#L6](https://github.com/yanet-platform/yanet2/blob/1b2c0101a8047bba6017c16772243e7ee1c423f0/lib/logging/log.h#L6) includes that header, and 31 `.c` files include `logging/log.h` directly (more reach it transitively), so a large share of yanet's own translation units record a manifest include-hash that changes on every single build and can never score a ccache direct hit across runs, no matter what else is fixed.

Calibration so this is not oversold: yanet's own TUs are roughly 10% of the ~1605 ccache lookups (DPDK dominates the rest), and `build.yml` has the identical exposure while still scoring 77-98% direct. So this is a ceiling on the achievable hit rate, not the cause of the `build-deb` gap that #1781 fixed.

Second consequence, arguably the bigger one: `SOURCE_DATE_EPOCH` is honoured nowhere in `meson.build` or `debian/rules`, so the produced Debian packages are not reproducible. Debian/Ubuntu export `SOURCE_DATE_EPOCH` for exactly this, and debhelper computes it already.

Two independent directions, without prescribing one:

- Honour `SOURCE_DATE_EPOCH` when it is set, deriving the stamp from the changelog rather than wall-clock time.
- Confine the stamp to a single translation unit so the rest of the tree stops depending on it (and therefore stops losing ccache direct-mode eligibility).

The two are complementary: the first fixes reproducibility, the second raises the achievable hit-rate ceiling; either alone is a real improvement.

## Acceptance

- [ ] A decision is recorded on which direction(s) to take (honour `SOURCE_DATE_EPOCH`, confine the stamp to one TU, or both).
- [ ] Debian packages built with `SOURCE_DATE_EPOCH` set are reproducible (identical output across two builds), or the reason they cannot be is recorded.
- [ ] If the stamp is confined to one TU, the ccache direct-hit rate for yanet's own TUs is re-measured against this baseline.

Source: DPDK build-caching RnD pass behind PR #1834 / issue #1781, measured 2026-08-08.

3Hren:
- The generated config header mixed three values that change only when a build option changes against six that change on essentially every configure, one of them a wall-clock timestamp at second granularity. The logging header includes it, and 31 source files include the logging header directly with more reaching it transitively, so a large share of the tree recorded a compiler-cache manifest that moved every build and could never score a direct-mode hit across runs. The cost is loss of direct-mode eligibility, which degrades to a preprocessor-mode fallback rather than a full recompile.
- The six volatile values move to their own header, included only by the dataplane version banner that prints them. That banner is their sole consumer anywhere in either repository, so nothing else has to change.
- Proven by execution rather than inference: after a reconfigure, the stable header is byte- and mtime-identical, so it does not even dirty ninja, while the new stamp header changes — the positive control showing the comparison can fail.
- The build date now honours the source date the packaging tools export, which is the standard contract for a reproducible timestamp. Scoped honestly: this repository's own deb path rewrites the changelog with a wall-clock entry immediately before invoking the build, so that value is effectively "now" there and two builds of one commit still differ. The change makes the stamp a function of the changelog, and delivers a reproducible timestamp for a rebuild against a fixed changelog. Making the deb path itself reproducible needs the changelog rewrite addressed separately.
- An empty source date is handled explicitly, which is the case a known dpkg bug produces, and a malformed one fails the configure rather than silently falling back.
- The direct-hit rate for this repository's own translation units will be re-measured on this pull request's own build the way it was for #1834, and posted here.

Closes #1837.

### DIFF
--- a/dataplane/main.c
+++ b/dataplane/main.c
@@ -6,7 +6,7 @@
 #include "config.h"
 #include "dataplane.h"
 #include "logging/log.h"
-#include "yanet_build_config.h"
+#include "yanet_build_stamp.h"
 
 static void
 print_version(void) {
--- a/meson.build
+++ b/meson.build
@@ -21,39 +21,57 @@ endif
 # The aarch64 branch appends its own ISA flag to yanet_project_args
 # further down, once the DPDK subproject() has resolved its ISA baseline.
 
+# Stable config: changes only when a build option changes, so it stays a
+# widely-included header without dragging per-build values into every
+# translation unit's ccache manifest.
 yanet_conf = configuration_data()
 yanet_conf.set('DEBUG_NAT64', get_option('buildtype') == 'debug')
 yanet_conf.set('ENABLE_TRACE_LOG', get_option('trace').enabled())
 yanet_conf.set('MBUF_MAX_SIZE', get_option('mbuf_max_size'))
 
+build_cfg = 'yanet_build_config.h'
+configure_file(output: build_cfg, configuration: yanet_conf)
+
+# Volatile config: version/commit/timestamp change on essentially every
+# build. Kept out of yanet_build_config.h and included only by the version
+# banner, so it does not poison ccache manifests across the rest of the tree.
+yanet_stamp = configuration_data()
+
 # Version and build info
 yanet_version = get_option('version') == '' ? meson.project_version() : get_option('version')
 
 cc = meson.get_compiler('c')
-yanet_conf.set_quoted('YANET_VERSION', yanet_version)
-yanet_conf.set_quoted('YANET_COMPILER_ID', cc.get_id())
-yanet_conf.set_quoted('YANET_COMPILER_VERSION', cc.version())
-yanet_conf.set_quoted('YANET_BUILD_TYPE', get_option('buildtype'))
+yanet_stamp.set_quoted('YANET_VERSION', yanet_version)
+yanet_stamp.set_quoted('YANET_COMPILER_ID', cc.get_id())
+yanet_stamp.set_quoted('YANET_COMPILER_VERSION', cc.version())
+yanet_stamp.set_quoted('YANET_BUILD_TYPE', get_option('buildtype'))
 
 # Git commit (if available)
 git = find_program('git', required: false)
 if git.found()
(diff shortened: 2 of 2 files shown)


3Hren:
Re-measured the way #1834 was, on this PR's own `build-deb (24.04)` run and a re-run of it against the cache the first one wrote.

| | #1834 warm, before this change | this PR, warm |
|---|---|---|
| Direct-mode lookups | 526 / 680 (77.35%) | **590 / 653 (90.35%)** |
| Preprocessed hits | 91 | **0** |
| Direct share of all hits | 526 / 617 (85.25%) | **590 / 590 (100.0%)** |
| Overall hit rate | 92.78% | 92.48% |
| Errors | 2 | 2 |
| Cacheable calls | 665 | 638 |

- The prediction this change was built on was that the 91 leftover preprocessed hits were translation units whose direct lookup failed because the generated header moved under them. They go to **zero**, and every hit is now direct.
- The overall hit *rate* is deliberately flat, and that is the honest reading: those 91 were already hits, just through the slower path that runs the preprocessor first. The saving is the preprocessor pass on each, not additional hits.
- Direct-mode lookups reaching 90.35% puts this job above the 77-98% band the other build jobs sit in, rather than merely inside it.
- The 27 fewer cacheable calls are #1838 landing in between, which stopped building testpmd — an independent corroboration of that change's 28-edge delta.
- Wall clock is not being claimed either way: the "Build packages" step was 492 s here against 460 s on #1834's warm run, which is runner-to-runner variation on a step whose critical path is cargo, not ninja.

PR #1863, opened 2026-08-08
test(dataplane-ut): share worker round preparation

### LINKED ISSUE #1755: test(dataplane): dataplane_ut hand-mirrors worker_loop_round instead of calling it
## Motivation

Two facts, verified against current HEAD (`bb3dc1cd`, line numbers have drifted from when this was first found but the facts still hold):

1. `lib/dataplane_ut/dataplane_ut.c` (around line 495-496, comment reads "Mirror worker_loop_round housekeeping: time, gen, iterations.") sets `dp_worker->current_time = ut->mock_time_ns` directly, then hand-duplicates the rest of `worker_loop_round`'s per-round preamble (generation store, iteration increment) instead of calling the real function. Consequence: `dataplane_ut` tests and `PipelineRound`-style benches execute zero of the real per-round preamble in `dataplane/worker.c`'s `worker_loop_round` (currently starting at `worker.c:325`), and the harness can silently drift from the real worker loop with no test catching it.

2. `lib/dataplane/time/clock.h` documents `dataplane_time_ns_fn` (the process-global function pointer) as an override point that "tests and alternative harnesses may install a substitute" for. Verified: exactly one call site in the entire tree (`dataplane/worker.c:331`) plus its default assignment (`lib/dataplane/time/clock.c:44`) — zero reassignments anywhere. `dataplane_ut` bypasses it entirely (fact 1) rather than using it as documented. The dead indirect call also blocks inlining of the clock read on every round, a pre-existing, small (sub-0.1%-class) cost not worth a PR on its own for the perf reason alone.

Fix shape (not yet designed in detail): add a `dataplane_ut` mode that actually calls `worker_loop_round`, using `dataplane_time_ns_fn` for the mock-time substitution, instead of hand-mirroring it. This closes the coverage gap and simultaneously makes the override point either genuinely used or safely deletable.

Caveat for future perf/behaviour work in this area: any pipeline-level benchmark (`PipelineRound` and siblings) that claims to cover the per-round preamble is a phantom result until this harness gap is fixed — the preamble code they think they are measuring never actually runs under the harness today.

## Acceptance

- [ ] Re-confirm the current line numbers for the hand-mirrored block in `dataplane_ut.c` and the real preamble in `worker.c` at whatever HEAD this is picked up at, since they have already drifted once.
- [ ] `dataplane_ut` has a mode that calls the real `worker_loop_round` (or an equivalent that routes through `dataplane_time_ns_fn`) instead of hand-mirroring its housekeeping.
- [ ] `dataplane_time_ns_fn` is either genuinely exercised by that mode (mock substitute installed through it) or removed if the chosen fix shape makes it redundant — this is decided and documented, not left dead either way.
- [ ] No regression to existing `dataplane_ut` callers that rely on the current hand-mirrored shortcut (call sites checked before changing shared harness code).

Source: migrated from the local task tracker's TASK-0073, filed 2026-07-16, migrated 2026-08-06.

3Hren:
- Shares worker-round preparation between the production loop and the in-process harness.
- Keeps deterministic harness clocks instance-local while preserving the direct TSC read in the dataplane.
- Verifies time, generation publication, iteration counting, and invalid-worker behavior.

Closes #1755.

### DIFF
--- a/dataplane/worker.c
+++ b/dataplane/worker.c
@@ -53,6 +53,7 @@
 #include "lib/controlplane/config/zone.h"
 #include "lib/dataplane/worker/counters.h"
 #include "lib/dataplane/worker/pipeline_round.h"
+#include "lib/dataplane/worker/round.h"
 
 #include "common/data_pipe.h"
 #include "logging/log.h"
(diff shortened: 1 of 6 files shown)


3Hren:
@codex review

PR #1878, opened 2026-08-09
fix(dataplane): transfer mbuf ownership to the consumer on the worker tx pipe

### LINKED ISSUE #1718: fix(dataplane): tail segments of a chained mbuf can be freed twice on a rejected tx
## Motivation

The pending-mbuf FIFO in `dataplane/worker.c` pins only the **head** segment of a chained mbuf. `worker_connection_push_cb` calls `rte_mbuf_refcnt_update(push_ctx->mbuf, 1)` on the mbuf handed to it, which is the head pointer alone — `packet_to_mbuf` in `worker_send_to_port` never walks `->next`. `worker_tx_pipe_reclaim` later frees that same head pointer once its refcount falls back to the recorded baseline.

The producer therefore holds a reference it believes protects the whole packet, expecting to free it itself once the consumer's NIC transmit completes. If the consumer instead classifies the packet as REJECTED and calls `rte_pktmbuf_free()` directly, that call walks the segment chain independently of the producer's pin: every segment is decremented, and any tail segment reaching zero returns to the mempool immediately. The producer, still waiting on what it believes is an ordered transmit completion, later frees the head again — and that walk follows `head->next` into segments that are already back in the pool. Double free, and a use-after-free of whatever has since been allocated there.

Reachable with an MTU-7000 KNI relay carrying a multi-segment mbuf, whenever the downstream transmit burst rejects it.

Adjacent but distinct: PR #1241 addresses the same file's multi-segment hazard on the packet **clone** path by deep-copying. This is the FIFO refcount scheme itself, so that fix does not cover it. The mechanism has been confirmed by reading the code and remains present at `bb3dc1cd`; it has **not** been reproduced dynamically, because the test that surfaced the area used single-segment packets only.

## Acceptance

- [ ] Bug-hunter confirmation: build a multi-segment mbuf, force a REJECTED classification on the consumer side, and show the tail-segment double free under ASan.
- [ ] A design is chosen — either account for every segment in the pending-mbuf scheme, or change the producer/consumer contract so a rejected packet is handed back rather than freed.
- [ ] Fix implemented, with a regression test exercising the multi-segment rejected-transmit path under ASan.

Source: analytical finding from a bug-hunter pass on 2026-07-21, surfaced while reviewing the separate virtio transmit-stall fix; code citations re-verified against `bb3dc1cd` before filing.

3Hren:
The producer pinned only the head segment of a chained mbuf pushed to another worker, recorded a refcount baseline, and performed the final free itself. Both the producer and the consumer then called the chain-walking free, so every tail segment of a multi-segment packet was returned to the mempool twice.

Confirmed under ASan on both transmit paths, not only the rejected one the issue describes. When the consumer accepts, the PMD's completion free is the same per-segment free, so the tails go back to the pool while the head survives at its baseline with its next pointer intact, and the producer's later free walks into pool memory.

The pin turned out to serve exactly one undocumented purpose: the per-worker rx mempool was created with no cache and `MEMPOOL_F_SP_PUT`, which DPDK maps to a single-producer ring, so only the owning lcore could ever return an mbuf to it. The pin kept every consumer-side free short of the pool. Dropping that flag makes cross-core puts legal and lets the consumer own the mbuf outright, which deletes the pending ring, the reclaim scan and two atomics per remote packet.

- Drop `MEMPOOL_F_SP_PUT` from the per-worker rx pool, keeping single-consumer gets, which stay sound because the rx-ring prefill runs before the worker thread exists and every later allocation is that worker's own.
- Remove the producer-side pin, the pending-mbuf ring and the reclaim loop, leaving only the free-phase call that advances the pipe's free position.
- Move the tx pipe into its own unit so a test can drive the real producer and consumer paths, which were previously static.
- Add Go regression tests over the real pipe covering a rejected transmit, an accepted transmit and the single-segment control, plus a case pinning the free-phase call that would otherwise wedge the pipe.

At source level the original defect is a wrong-primitive mistake: `rte_mbuf_refcnt_update` pins one segment where `rte_pktmbuf_refcnt_update` pins the whole chain. Correcting just the primitive was considered and rejected, and the reason is not the primitive. With every segment pinned, the producer's readiness test can still observe the head back at its baseline while the consumer is midway through decrementing the tail segments, so the two walks race on a tail and the losing side's mempool put can land on the consumer lcore. That is the very invariant the pin exists to protect, and against a single-producer ring it corrupts silently rather than crashing, turning an ASan-visible double free into an invisible one. Closing the race needs the readiness test to check every segment, which re-reads N cachelines the consumer core has just dirtied, and the scheme also rests on three unenforced invariants: that no module ever shares an mbuf segment, that `RTE_ETH_TX_OFFLOAD_MBUF_FAST_FREE` is never enabled, and that consumers free head to tail.

Verification: the multi-segment test was written against the unfixed code and fails there with a use-after-free, so the before and after are both demonstrated rather than reconstructed. A bug-hunter pass reproduced all four original scenarios against the fixed code with a positive control, a randomized stress and a two-thread run over roughly 1.1M mbufs. On `lab-vla-fw3` at a saturating 20 Mpps the change is within the noise floor measured between two builds of identical source, and microbenchmarks put the removed cross-core cacheline round trip at 118-207 cycles per remote packet against 12-26 cycles added per mempool put.

Closes #1718.

### DIFF
--- a/bindings/go/dataplane_ut/tx_pipe.go
+++ b/bindings/go/dataplane_ut/tx_pipe.go
@@ -1,0 +1,92 @@
+package dataplaneut
+
+/*
+#include <stdint.h>
+
+#include "lib/dataplane_ut/tx_pipe.h"
+*/
+import "C"
+
+import "fmt"
+
+// TxPipe is a handle to a dataplane_ut tx-pipe harness: one worker_tx_pipe
+// plus its own malloc-backed mock mempool, letting a test drive the tx pipe
+// producer/consumer paths directly.
+//
+// Free must be called when the test is done.
+type TxPipe struct {
+	ptr *C.struct_dataplane_ut_tx_pipe
+}
+
+// NewTxPipe constructs a TxPipe whose pipe holds (1 << pipeSize) entries,
+// per worker_tx_pipe_init.
+func NewTxPipe(pipeSize int) (*TxPipe, error) {
+	ptr := C.dataplane_ut_tx_pipe_new(C.size_t(pipeSize))
+	if ptr == nil {
+		return nil, fmt.Errorf("failed to create tx pipe harness: dataplane_ut_tx_pipe_new returned NULL")
+	}
+	return &TxPipe{ptr: ptr}, nil
+}
+
+// Free tears down the harness. Nil-safe.
+func (m *TxPipe) Free() {
+	if m.ptr == nil {
+		return
+	}
+	C.dataplane_ut_tx_pipe_free(m.ptr)
+	m.ptr = nil
+}
+
+// TxMbuf is an opaque handle to a chained mbuf allocated from a TxPipe's
+// mock mempool by AllocMbuf, to be handed to Push.
+type TxMbuf struct {
+	ptr *C.struct_rte_mbuf
+}
+
+// AllocMbuf allocates a chained mbuf of segCount segments from the harness
+// mempool, each segment holding segSize bytes of payload.
+func (m *TxPipe) AllocMbuf(segCount, segSize uint32) (*TxMbuf, error) {
+	ptr := C.dataplane_ut_tx_pipe_alloc_mbuf(m.ptr, C.uint32_t(segCount), C.uint32_t(segSize))
+	if ptr == nil {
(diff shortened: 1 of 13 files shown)


3Hren:
@codex review

3Hren:
@codex review

PR #1885, opened 2026-08-10
fix(acl): label metrics client errors correctly

### LINKED ISSUE #1744: fix(acl): label CLI metrics-client errors with the correct service name
## Motivation

`modules/acl/cli/src/main.rs` defines `SERVICE_NAME = "modules.acl.controlplane.aclpb.v1.ACLService"` (line 291) and passes it to both `Service::new` calls — the ACL service client (line 301) and the metrics service client (line 308). Any error surfaced by the metrics client (connection failure, RPC error) is labeled `ACLService` instead of the actual target (`…aclpb.v1.MetricsService`), which misleads whoever reads the error message during an incident. Confirmed still present on current `main`.

Data correctness is unaffected — this only mislabels the service name inside error text.

## Acceptance

- [ ] The metrics client's `Service::new` call uses the correct `MetricsService` FQN, not the ACL service's, either via a second constant or by deriving the FQN from the generated proto descriptor.

Source: local task tracker migration, item dated 2026-07-09, verified against current `main` on 2026-08-06.

3Hren:
- Uses the metrics service identity when labeling ACL metrics client failures.

Closes #1744.

### DIFF
--- a/modules/acl/cli/src/main.rs
+++ b/modules/acl/cli/src/main.rs
@@ -289,6 +289,7 @@ pub struct Cmd {
 
 /// The fully-qualified gRPC service name used in error messages.
 const SERVICE_NAME: &str = "modules.acl.controlplane.aclpb.v1.ACLService";
+const METRICS_SERVICE_NAME: &str = "modules.acl.controlplane.aclpb.v1.MetricsService";
 
 pub struct ACLService {
     service: Service<AclServiceClient<LayeredChannel>>,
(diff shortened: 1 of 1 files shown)


3Hren:
Simplified the change to the direct service-name fix and rebased it onto current `main`.

@codex review

3Hren:
Rebased the unchanged minimal fix onto the latest `main`; local workspace gates and independent review are green.

@codex review

3Hren:
@codex review

PR #1894, opened 2026-08-10
refactor(pdump)!: make logging optional

### LINKED ISSUE #1727: refactor(pdump,bird-adapter): design logger sourcing for non-constructor methods
## Motivation

The `logger:` ledger's last 5 rows are methods, not constructors, so the mechanical
"thread a `WithLog` constructor option" pattern that already retired the rest of the ledger
doesn't apply as-is:

- `modules/pdump/controlplane/ffi.go:ModuleConfig.SetupRing`
- `operators/bird-adapter/service.go:AdapterService.processBirdImport`
- `operators/bird-adapter/service.go:AdapterService.reconnectStream`
- `operators/bird-adapter/service.go:AdapterService.runBirdImportLoop`

Verified today directly against `lint/style/allowlist.txt`: exactly these 5 rows remain under the
`logger:` check in the whole repository, matching this issue's scope precisely.

Bundled into the same change: `modules/pdump/controlplane/service.go:NewPdumpService`,
mechanically identical to the constructor-option pattern already used elsewhere, grouped here
only because it shares a package with `SetupRing` and would otherwise collide with a second PR
touching the same ledger lines.

Each of the 4 methods needs a per-case decision: either the logger belongs on the struct (set
once via the constructor's option, the method reads it off the receiver), or it is a genuinely
per-call derived logger — built from request-scoped fields — that belongs constructed at the
call site instead of threaded in as a parameter. This is a design decision, not a mechanical
rename.

## Scope

Likely two changes once the design is settled — pdump (`SetupRing` plus `NewPdumpService`) and
bird-adapter (3 methods) touch disjoint packages — but keep as one issue until a closer look at
the design says otherwise.

## Acceptance

- [ ] Each of the 4 methods' logger sourcing is decided (struct field via constructor option vs.
      call-site-derived) and implemented.
- [ ] `NewPdumpService` migrated to the options pattern in the same change.
- [ ] All 5 corresponding rows deleted from `lint/style/allowlist.txt`.
- [ ] `make lint-go` and the affected packages' `go test` pass.

Source: migrated from the retired local task tracker's task filed 2026-07-15, on 2026-08-06.

3Hren:
- Makes service and module-config logging optional with no-op defaults.
- Moves ring logger ownership to the Go wrapper without changing C or shared-memory state.
- Removes the final two `logger:` style-debt entries.
- Requires downstream `NewPdumpService` callers to use `WithPdumpServiceLog`.

Closes #1727.

### DIFF
--- a/lint/style/allowlist.txt
+++ b/lint/style/allowlist.txt
@@ -35,8 +35,6 @@
 
 barenew:common/go/grpcmetrics/grpcmetrics.go:New:1  # legacy: rename to a descriptive constructor name
 barenew:common/go/xbackoff/backoff.go:New:1  # legacy: rename to a descriptive constructor name
-logger:modules/pdump/controlplane/ffi.go:ModuleConfig.SetupRing:1  # legacy: migrate to the options pattern
-logger:modules/pdump/controlplane/service.go:NewPdumpService:1  # legacy: migrate to the options pattern
 loggerlast:common/go/readiness/readiness.go:options.Log:1  # legacy: move the *zap.Logger field to the end of the struct
 loggerlast:controlplane/gateway/gateway.go:gatewayOptions.Log:1  # legacy: move the *zap.Logger field to the end of the struct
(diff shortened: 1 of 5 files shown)


3Hren:
@codex review

Issue #1887, opened 2026-08-10
fix(controlplane): make generation reclamation safe for both C and Go owners


3Hren:
## Motivation

[`cp_config_gen_install`](https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/lib/controlplane/config/zone.c#L200-L229) publishes the new configuration generation, then waits for every worker to acknowledge it, then reclaims the superseded generation:

```c
ATOMIC_SET_OFFSET_OF(&cp_config->cp_config_gen, new_config_gen);  // L224 publish
dp_config_wait_for_gen(dp_config, new_config_gen->gen);            // L225 wait (the grace period)
cp_config_gen_free(cp_config, old_config_gen);                     // L226 reclaim
```

The wait at line 225 is the grace period. The finding: **once line 224 has executed, there is no value the function can return that is safe.**

### Returning FAILURE is unsafe

All ten in-file callers of `cp_config_gen_install` respond to a nonzero return with `goto error_free`, which calls `cp_config_gen_free(cp_config, new_config_gen)` on the generation workers are currently executing — a use-after-free of the live configuration. This is the reason #1676 deliberately keeps `cp_config_gen_install` returning success even when the worker-quiescence wait itself times out.

It is unsafe a second time, in Go: [`backend.UpdateModule`](https://github.com/yanet-platform/yanet2/blob/5759684ba3e0dc15cfd35c36513b6f4bd7f05faf/modules/decap/controlplane/backend.go#L36-L41)'s error path calls `mod.Free()` on the **new** module config:

```go
if err := m.agent.UpdateModules(
	[]ffi.ModuleConfig{mod.AsFFIModule()},
); err != nil {
	mod.Free()
	return nil, fmt.Errorf("failed to update module %q: %w", name, err)
}
```

### Returning SUCCESS before quiescence is unsafe in the other direction

The `cp_module` blocks are owned by Go, not by `cp_config_gen`. [`DecapService.updateConfig`](https://github.com/yanet-platform/yanet2/blob/f56fe6ae96abae02498418382c9032f59585bfdf/modules/decap/controlplane/service.go#L151-L167) calls `old.Free()` unconditionally as soon as the backend reports no error:

```go
if old, ok := m.configs[name]; ok {
	old.Free()
}
```

[...]

Issue #1882, opened 2026-08-10
fix(dataplane): fail startup when a worker thread cannot be created


3Hren:
## Motivation

`dataplane_worker_start` creates the worker thread with a raw, unchecked `pthread_create`:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/dataplane/worker.c#L602-L604

The function always returns 0 regardless of the outcome, and its own caller discards the return too:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/dataplane/device.c#L80-L83

The thread is pinned to `worker->config.core_id`, which comes straight from config parsing with no bound check against the online CPU set -- only against trailing garbage in the string:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/dataplane/config.c#L250-L257

The worker is already counted before its thread is created: `dp_config->worker_count` is incremented inside `dataplane_init`, while threads are only launched later from `dataplane_start`:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/dataplane/worker.c#L444-L457

## Consequence

A `core_id` naming a CPU that is not online makes `pthread_create` fail. Because the failure is silently discarded, `dp_config->worker_count` permanently counts a worker that has no thread. That worker's `gen` never advances past 0. `dp_config_wait_for_gen` then spins on it forever:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/lib/dataplane/config/zone.c#L14-L29

and it is called while `cp_config_gen_install` holds `cp_config_lock`:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/lib/controlplane/config/zone.c#L222-L225

So the very first config install after startup wedges the entire control plane at 100% CPU, with nothing logged anywhere.

Measured on this dev box (40 online CPUs): a `core_id = 40` config makes `pthread_create` fail with `rc=22` (EINVAL) -- the thread never runs, and `worker_count` still includes it as live.

## Scope

- [ ] Propagate the `pthread_create` failure up through `dataplane_worker_start` and `dataplane_device_start` instead of discarding it.
[...]

Issue #1881, opened 2026-08-10
fix(dataplane_ut): worker_round_prepare overwrites the generation sentinel, hanging later config updates


3Hren:
## Motivation

`dataplane_ut_new` primes each synthetic worker's generation counter with a sentinel far above any real config generation, specifically so the harness never blocks waiting for a config republish:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/lib/dataplane_ut/dataplane_ut.c#L348

That sentinel does not survive processing. Every call that drives a packet round through the harness calls `worker_round_prepare`:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/lib/dataplane_ut/dataplane_ut.c#L497

and `worker_round_prepare` unconditionally overwrites the worker's generation with the *real* current generation, release-ordered:

https://github.com/yanet-platform/yanet2/blob/8a051bcf9ea2882a9002bedbf8b823d9c2d02829/lib/dataplane/worker/round.h#L31-L33

So after a single `HandlePackets*` call, the sentinel is gone. A subsequent config write that goes through `cp_config_gen_install` → `dp_config_wait_for_gen` now waits for this worker's generation to reach the newly installed one — a bump that will never arrive on its own, because nothing drives another processing round for that purpose. `dp_config_wait_for_gen` itself has no bound or yield (`lib/dataplane/config/zone.c`, tracked separately in #1676), so the wait does not fail, it spins forever.

**This is not theoretical.** A private module's functional test suite hit exactly this sequence — an `agent.Update*` call issued after a `HandlePackets*` call on the same harness instance — and the install call hung for the full 300s test timeout with the control plane spinning under the config lock. The only way that suite could proceed was a test-only mechanism that pokes the generation counter directly rather than exercising the real republish path, plus a standing rule to never call `agent.Update*` after `HandlePackets*` on one harness. That is a live workaround for a defect in the shared test harness, not a documented limitation of the design.

Existing suites in this repository do not hit this because they perform their whole configuration before the first processing round. For example, `modules/fwstate/tests/functional/fwstate_test.go` issues every `agent.Update*` call inside `setupFWStateHarness`, before any subtest exercises the harness. That is why a defect with a confirmed victim is nonetheless invisible in this repo's own CI.

## Relationship to #1676

[...]

Issue #1676, opened 2026-08-05
fix(dataplane): bound the config-generation handshake instead of spinning under the lock


3Hren:
## Motivation

`dp_config_wait_for_gen` (`lib/dataplane/config/zone.c`) spins until every worker republishes its generation counter. The loop has no yield — the code says `// TODO cpu yield` — no timeout, and no failure path. It is called from `cp_config_gen_install` (`lib/controlplane/config/zone.c`) while the config lock is held, which puts it on the path of every module config write.

Today workers tick roughly a million times a second, so the wait is sub-microsecond and the defect is invisible in the common case. The trigger that makes it not invisible is not merely a slow tick: a counted dataplane worker can have no thread at all if its `pthread_create` call fails and the failure is discarded, which pins that worker&#39;s `gen` at 0 forever and turns the very first config install after startup into a permanent, silent, 100%-CPU control-plane hang. See #1882 for that trigger.

## Evidence

A standalone harness driving the real `dp_config_wait_for_gen` against synthetic workers at a controlled tick period, 4 workers, 20 installs:

| worker tick | mean wait / install | worst | control-plane CPU during the wait |
| --- | --- | --- | --- |
| busy loop | 0.001 ms | 0.008 ms | — |
| 1 ms | 1.071 ms | 1.131 ms | ~100% |
| 100 ms | 99.4 ms | 135.8 ms | ~100% |

The last column is the point: the control plane is spinning, not blocking, for the entire wait, once per config install, with the lock held.

## A second, smaller defect in the same function

The loop is a `do { ... } while (idx &lt; dp_config-&gt;worker_count)`. With `worker_count == 0` and a valid, non-NULL `workers` array, the body executes once before the condition is evaluated, but `continue` inside the `do/while` jumps straight to the controlling expression, so the function returns cleanly — `worker_count == 0` alone is not what crashes.

The crash comes solely from `dp_config-&gt;workers` itself being a zero offset. `ADDR_OF` (`common/memory_address.h`) returns NULL for a zero offset because of its ternary, so `workers` is NULL, `workers + 0` is NULL, and `ADDR_OF`&#39;s first statement loads through it — a SIGSEGV inside the control-plane process, not a stray read. Both conditions co-occur on a zeroed segment, which is why they looked coupled. Reproduced under ASan: `lib/dataplane/config/zone.c:19:30: runtime error: load of null pointer of type &#39;struct dp_worker *&#39;`.

[...]

PR #1898, opened 2026-08-10
fix(dataplane): yield while waiting for a generation acknowledgement

3Hren:
Addresses part of #1676. Deliberately does not close it — the bounding half stays open, see below.

The wait for every worker to acknowledge a new configuration generation spun with no yield and no pause, so the control-plane thread burned a full core for as long as the slowest worker took, all of it while holding the configuration lock. It now checks, then yields for a bounded phase, then sleeps at a fixed one-millisecond interval. Measured against a worker that never acknowledges, the waiting thread drops from 99.35% of a core to between 1.6% and 1.8%.

The healthy path is unaffected — a worker that has already acknowledged is noticed on the first check, before any yield, measured at 34, 46 and 55 nanoseconds for one, four and sixteen workers.

The walk is also bounded by the worker count rather than run at least once. That removes an unconditional first-iteration dereference on an instance with no workers, which #1901 has since made unreachable in production, so it is hardening rather than a fix.

Limits, stated because they are easy to read into this and are not here:

- The wait still blocks indefinitely, by design. Once the new generation is published there is no safe return value: reporting failure makes every caller free the generation the workers are executing, and reporting success makes the module services free the superseded one while an unacknowledged worker may still read it. Bounding the wait is blocked on #1887.
- Only the waiting thread stopped spinning. The configuration lock is still a bare compare-and-swap spin, so a concurrent writer keeps burning a core while the holder sleeps.
- The real-world trigger for an unbounded wait is a counted worker whose thread was never created, which is #1882 and is not addressed here.

On testing, which is the part worth knowing before approving:

- This change ships no test, and the existing suite does not exercise the new code. Every harness worker is seeded past any real generation, so the first check returns and nothing reaches the yield or sleep rungs. A green suite here means the change compiles and breaks nothing, not that the backoff works.
- That is deliberate rather than an omission. Earlier revisions added a harness entry point so a test could drive the wait directly, and three cases that pinned the comparison, the recheck and the walk over every worker. None of them failed against the unfixed revision, and reaching the interesting states needs a worker held behind while another thread advances it, which deadlocks in this harness for the reason tracked as #1881. The test-only surface was judged disproportionate for what remains and was removed.
- What remains is a change in how the wait spends its time, whose effect is a measurement, and a loop shape. Neither is something a test in this repository can pin today.

### DIFF
--- a/lib/dataplane/config/zone.c
+++ b/lib/dataplane/config/zone.c
@@ -1,5 +1,7 @@
 #include "zone.h"
 
+#include <sched.h>
+#include <time.h>
 #include <unistd.h>
 
 struct dp_config *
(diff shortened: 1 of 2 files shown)


3Hren:
Reworked after #1901 landed: the zero-worker state this branch used to defend against is now rejected at dataplane startup, so the guard, its test and the harness support for that state were removed. What remains is the yielding backoff and the loop rewrite. @codex review

3Hren:
Cut again: the harness entry point, its Go bindings and the three tests are gone, since reaching the interesting states through the harness deadlocks and the trivial one is already covered by every module functional suite. What is left is the backoff and the loop bound, two files. @codex review

PR #1903, opened 2026-08-10
feat(fwstate): configurable sync suppression timeout

### LINKED ISSUE #1902: fwstate: configurable sync suppression timeout to cut redundant state-sync refreshes
## Problem

In a state-driven firewall the ACL module emits a state-sync frame for every rule that triggers `CREATE_STATE`/`CHECK_STATE`, and the fwstate module then updates the corresponding fwmap entry **and** forwards the packet to peer firewalls. For a frequently touched session this produces a ~2x packet amplification (one data packet in, one sync packet processed+forwarded out), burning dataplane CPU on redundant fwmap refreshes and consuming output bandwidth for state refreshes that carry no new information.

## Proposal

Add a configurable **sync suppression timeout** to the fwstate module. When a sync frame arrives for an entry that already exists and is alive:

- If the new entry's expiry deadline would land within the suppression window of the entry's *current* expiry deadline, the frame is **discarded** — the fwmap record is left untouched (no write, no counter bump beyond a `suppressed` counter).
- Otherwise the frame is applied as today.

This debounces refreshes so a busy session is refreshed at most once per suppression window instead of once per packet, while a state transition that changes the TTL (e.g. SYN → established) still refreshes immediately because its deadline jumps past the window.

To keep the effective session keep-alive no shorter than the configured timeout, every fwstate TTL is increased by the suppression delta (`effective_ttl = configured_ttl + sync_suppress_timeout`). This guarantees that even when one refresh is suppressed, the entry retains at least `configured_ttl` of remaining life.

A fully suppressed *internal* sync packet (all of its frames suppressed) is dropped instead of forwarded, sparing downstream pipeline and inter-firewall bandwidth.

### Configuration

New `SyncConfig` field:
- `sync_suppress_timeout` (uint64, nanoseconds). `0` disables suppression (default, preserves current behaviour).

### Scope

- `lib/fwstate/config.h` — new `sync_suppress_timeout` member on `fwstate_sync_config`.
- `modules/fwstate/dataplane/dataplane.c` — TTL bump by the delta + per-frame suppression check before the fwmap put; drop fully-suppressed internal packets.
- `modules/fwstate/api/fwstate_cp.c` + `dataplane/config.h` — new `fwstate_sync_v4_suppressed` / `fwstate_sync_v6_suppressed` counters.
- proto + Go bindings + convert — plumb the new field end to end.
- Tests covering suppression, TTL bump, and the disabled (`0`) default.

## Out of scope

- Changing ACL-side sync frame *creation* frequency (suppression is local to fwstate).
- Per-frame rebuild of batched multi-frame sync packets (ACL currently emits one frame per packet).

GeorgyKirichenko:
The ACL module emits a state-sync frame for every stateful rule hit, and fwstate refreshes the fwmap and forwards the packet, producing roughly 2x packet amplification for frequently touched sessions.

- Add a configurable `sync_suppress_timeout` to the fwstate sync config (nanoseconds, zero disables).
- A sync frame for an alive entry whose new expiry lands within the window of the current deadline is discarded, leaving the fwmap record untouched, so a busy session is refreshed at most once per window instead of once per packet.
- Every entry TTL is inflated by the window so the effective keep-alive never drops below the configured timeout.
- A fully suppressed internal sync packet is dropped instead of forwarded, sparing downstream and inter-firewall bandwidth.
- New per-family `fwstate_sync_v4_suppressed` / `fwstate_sync_v6_suppressed` counters are exposed.
- Validate that each timeout plus the suppress window still fits the 48-bit stored TTL.

Plumbed end to end: C config struct, dataplane suppression logic, proto field, Go bindings, and converters.

Closes #1902.


GeorgyKirichenko:
Addressed all review findings: (P1) suppression predicate now only fires in the small-extension band, so shorter-TTL teardown frames (FIN/RST) and longer-TTL state progressions are always applied; (P2) sync_suppress_timeout is a direct override so 0 disables it; (P2) overflow-safe validation; (P2) fuzzer now registers the new counters; (P2) stale-layer entries are never suppressed (always promoted); (P2) suppressed counters wired into the metrics adapter. @codex review

GeorgyKirichenko:
Round 2: (P1) suppression now also skips when the frame carries flag bits the entry hasn't seen, so same-TTL handshake transitions (SYN->ACK) always merge; (P2) sync_suppress_timeout is now proto3 optional — absent inherits the current window (older clients unaffected), explicit zero disables. @codex review

GeorgyKirichenko:
Round 3: (P2) flags are now snapshotted under the read lock before the suppress decision, closing the race with a concurrent put; (P2) equal-deadline duplicates are now suppressed too (the shorter-TTL teardown path is preserved); (P2) added `--sync-suppress-timeout` to the fwstate CLI update command. @codex review

GeorgyKirichenko:
Round 4: stale-layer flag snapshots are now skipped (those entries carry no lock and are never suppressible).

On the mixed-packet finding: the only internal sync packets fwstate receives are crafted by the ACL module one frame per packet (sync.c appends a single fw_state_sync_frame), and external (potentially batched) packets are dropped after processing rather than forwarded. So a partially-suppressed internal packet that is still forwarded does not occur on the current paths — the mbuf-compaction remedy would be disproportionate. Filed as a follow-up consideration in the issue if multi-frame internal sync packets are ever introduced. @codex review

PR #1929, opened 2026-08-11
refactor(counters)!: rename the worker drops metric to disposed

3Hren:
- Renamed the aggregate worker metric from `worker_drops` to `worker_disposed`, carrying the operator terminology introduced for the counters CLI column into the metric surface.
- Marked as breaking: the rename has no deprecation window and no dual emission, matching the precedent set when the redundant metric-name prefix was dropped, so any external dashboard or alert reading the old name has to be updated.
- Left the counter itself, its protobuf field, the dataplane counter name and the reason-specific transmit drop metrics unchanged, so the reported value is identical.

There are no consumers of the old name inside either repository.


3Hren:
Dropped the explanatory comment — the rename stands on its own. @codex review

PR #1938, opened 2026-08-11
fix(dscp): allocate test config from shared arena

### LINKED ISSUE #1876: fix(dscp): test harness splices Go-heap memory contexts into shared-memory arena
## Motivation

`modules/dscp/tests/dataplane/dscp.go` builds its module config as a Go-heap `C.struct_dscp_module_config{}` composite literal, then calls `lpm_init` on two `struct lpm` fields inside it. `struct lpm` embeds a `struct memory_context` ([common/lpm.h#L45](https://github.com/yanet-platform/yanet2/blob/513e6f178af10f5e145fbd855961d02c0d2528c5/common/lpm.h#L45)), so the harness splices two Go-heap-backed memory contexts into a C arena's child list through the offset-pointer (`ADDR_OF`) machinery — the same shared-memory ownership contract the shm offset-pointer convention enforces for production code.

https://github.com/yanet-platform/yanet2/blob/513e6f178af10f5e145fbd855961d02c0d2528c5/modules/dscp/tests/dataplane/dscp.go#L87-L91

PR #1159 removed the identical hazard from the nat64 test harness, replacing a Go-heap composite literal with `memory_balloc` from the test memory context. dscp's own sibling harness already does this correctly: [modules/fwstate/tests/dataplane/fwstate.go#L35-L38](https://github.com/yanet-platform/yanet2/blob/faec6fea7ca3c50a34dbf6564aff35fa0492476f/modules/fwstate/tests/dataplane/fwstate.go#L35-L38). dscp is the module AGENTS.md names as the canonical reference harness other modules are told to copy, so leaving it as the outlier risks the pattern propagating into every new module test suite.

It is latent rather than live: Go's GC is non-moving today and the calls into C are synchronous, so no relocation or premature free has been observed. But the fix is small and mechanical, and has two working precedents already in the tree (fwstate, and nat64 as of #1159).

## Acceptance

- [ ] dscp test harness allocates `struct dscp_module_config` via `memory_balloc` from the test memory context, not a Go-heap composite literal.
- [ ] The freshly `memory_balloc`'d block is `memset` to zero before use — `memory_balloc` does not zero, and the Go-heap literal it replaces was zero-filled.
- [ ] Existing dscp dataplane tests still pass.

Source: review of #1159 — PR did not touch dscp, but surfaced the same Go-heap-into-C-arena hazard class it fixed in nat64's harness.

3Hren:
- Allocate the DSCP dataplane test configuration from its shared memory arena.
- Zero newly allocated configuration storage before building lookup tables.
- Preserve arena ownership for embedded lookup-table memory contexts.

Closes #1876.

### DIFF
--- a/modules/dscp/tests/dataplane/dscp.go
+++ b/modules/dscp/tests/dataplane/dscp.go
@@ -10,6 +10,7 @@ package dscp_test
 #include "modules/dscp/dataplane/config.h"
 
 #include <stdlib.h>
+#include <string.h>
 
 uint8_t dscp_mark_never = DSCP_MARK_NEVER;
 uint8_t dscp_mark_default = DSCP_MARK_DEFAULT;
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

3Hren:
@codex review

3Hren:
@codex review

3Hren:
@codex review

3Hren:
@codex review

PR #1939, opened 2026-08-11
feat(commonpb): add a contiguous IP network message

### LINKED ISSUE #1908: feat(commonpb): add a contiguous IP network message
## Motivation

`commonpb` has a shared host-address message and an inclusive address-range message, but no shared message for one contiguous CIDR network. Callers that need a prefix therefore cannot reuse a canonical commonpb wire type.

[common/commonpb/v1/ipaddr.proto:7-20](https://github.com/yanet-platform/yanet2/blob/7c0e93017abc390f6b623cd22c887d941ed36dab/common/commonpb/v1/ipaddr.proto#L7-L20)

[common/commonpb/v1/iprange.proto:9-18](https://github.com/yanet-platform/yanet2/blob/7c0e93017abc390f6b623cd22c887d941ed36dab/common/commonpb/v1/iprange.proto#L9-L18)

## Scope

- Add `ContiguousIPNetwork` to `common.commonpb.v1` for one IPv4 or IPv6 CIDR prefix.
- Define its address-family and prefix-length validity contract consistently with `IPAddress`.
- Provide the normal generated bindings and Go conversion/validation helpers used by the existing commonpb IP types.
- Do not migrate unrelated module-specific prefix messages in this change.

## Acceptance

- [ ] The proto message represents both IPv4 and IPv6 CIDR prefixes without string parsing.
- [ ] Invalid address lengths and out-of-range prefix lengths return errors rather than producing ambiguous values.
- [ ] Go helpers round-trip representative `netip.Prefix` values for both families.
- [ ] JSON behavior is defined and covered consistently with the existing commonpb IP types.
- [ ] Generated protobuf artifacts and Meson/proto registration are updated as required.

Source: user backlog item, confirmed against [common/commonpb/v1/ipaddr.proto:7-20](https://github.com/yanet-platform/yanet2/blob/7c0e93017abc390f6b623cd22c887d941ed36dab/common/commonpb/v1/ipaddr.proto#L7-L20) and [common/commonpb/v1/iprange.proto:9-18](https://github.com/yanet-platform/yanet2/blob/7c0e93017abc390f6b623cd22c887d941ed36dab/common/commonpb/v1/iprange.proto#L9-L18).

3Hren:
Add a shared wire type for a single IPv4 or IPv6 CIDR prefix, so callers that need a prefix can reuse a canonical commonpb message instead of a module-specific one or a plain string.

The address family is carried by the address byte length, consistently with the existing host-address message, and the prefix length is validated against that family on decode. A malformed address length, a missing address, and an out-of-range prefix length all return errors rather than an ambiguous value.

Host bits are normalised on both construction and decoding, in Go and in Rust alike, so the same message means the same thing in both languages. That choice is forced by the Rust network type, which cannot represent host bits at all, and it makes the round trip deliberately lossy for an input that carries them.

Both language bindings get parsing from a string, human-readable rendering, and conversion to the language's own network type, following the existing address and hardware-address types. JSON rendering matches each side's established shape for those types.

Existing module-specific prefix messages are deliberately left alone here and tracked as separate follow-ups.

Closes #1908.

### DIFF
--- a/common/commonpb/meson.build
+++ b/common/commonpb/meson.build
@@ -7,12 +7,12 @@ common_proto_files = [
     join_paths(common_proto_dir, 'v1', 'macaddr.proto'),
     join_paths(common_proto_dir, 'v1', 'ipaddr.proto'),
     join_paths(common_proto_dir, 'v1', 'iprange.proto'),
+    join_paths(common_proto_dir, 'v1', 'ipnetwork.proto'),
 ]
 
-# Generate protobuf files. The root include dir is required so that
-# iprange.proto can import ipaddr.proto via the canonical path
-# "common/commonpb/v1/ipaddr.proto", which matches the import path used by
-# consumers outside this directory.
+# Generate protobuf files. The root include dir is required so that protos
+# in this directory can import each other via the canonical path that
+# consumers outside this directory use.
 common_protoc_gen = custom_target(
     'common-protoc',
     output: [
@@ -21,6 +21,7 @@ common_protoc_gen = custom_target(
         'macaddr.pb.go',
         'ipaddr.pb.go',
         'iprange.pb.go',
+        'ipnetwork.pb.go',
     ],
     input: common_proto_files,
     command: [
--- a/common/commonpb/v1/ipnetwork.go
+++ b/common/commonpb/v1/ipnetwork.go
@@ -1,0 +1,99 @@
+package commonpb
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/netip"
+)
+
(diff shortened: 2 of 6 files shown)


3Hren:
Addressed the review feedback:

- `NewContiguousIPNetworkFromPrefix` now returns an error instead of yielding an empty message that only failed later at decode, following the fallible `NewIPRange` in the same package.
- The overlong error in `ToPrefix` is reflowed one argument per line, matching the existing style in `iprange.go`.
- JSON marshalling now goes through the same intermediate type both directions share, instead of hand-formatting the wire shape.
- serde's traits are imported rather than path-qualified across the whole crate, matching `filterpb`.

@codex review

PR #1940, opened 2026-08-11
fix(counters): stop holding the config lock across counter value copies

### LINKED ISSUE #1914: fix(counters): stop holding the config lock across counter value copies
## Motivation

A production control plane livelocked. An untagged metrics scrape read every per-rule ACL counter (78 535 per module position, 671 874 across 12 positions), doing an allocation and memcpy per counter per worker — roughly 16M allocations — inside a single hold of `cp_config_lock`, the one global shared-memory spinlock that every shared-memory read and write path takes. The first uncontended scrape measured 7.8 s against a 4-6 s scrape interval, so the duty cycle exceeded 100% immediately. Each blocked cgo caller permanently pins one OS thread because a cgo call cannot be cancelled, and the process reached 5678 threads against Go's default 10000 limit.

[`yanet_get_counters_by_tags`](https://github.com/yanet-platform/yanet2/blob/6057bbcaa8c95e2fbaa121f69ccd3dd7176765ea/lib/controlplane/agent/agent.c#L1734-L1744) takes `cp_config_lock` and holds it through the whole matching-and-copy sequence, including the [per-counter copy loop](https://github.com/yanet-platform/yanet2/blob/6057bbcaa8c95e2fbaa121f69ccd3dd7176765ea/lib/controlplane/agent/agent.c#L1709-L1728), which does a `calloc` for the per-worker pointer array plus a `malloc` and `memcpy` for every worker's value block, for every counter. Cost is counters times workers, with one allocation per element.

## Scope

Resolve the matching storage set under `cp_config_lock`, release the lock, then snapshot values outside it.

## Acceptance

- [ ] `cp_config_lock` is released before the per-counter value copy loop runs.
- [ ] A regression test asserts on **lock-hold time** — for example that `cp_config_try_lock` from another thread succeeds within a bound during a large counter read — not on total call duration, since a fast lock-free copy phase could otherwise hide a regression.
- [ ] Existing counter-read semantics are unchanged: a config generation swap racing the read either serves a consistent pre-swap snapshot or a consistent post-swap one, never a mix.

Source: production incident investigation, 2026-08-10. See also #1739 (bounding reader concurrency and lock fairness on the same lock) and #1913 (incident epic).

3Hren:
- Releases the shared-memory config lock before the per-counter value copy loop in the tagged counter read path: the matching storages and their metadata are resolved under the lock, each matched storage is pinned by reference count, and the allocation-heavy value copies run unlocked against the pinned storages.
- Adds a counted storage release variant so dropping a pin after a racing config update does not read the registry retired together with its owning object.
- Adds a regression test asserting the lock becomes acquirable from another thread while a large read is still copying values, with a positive control requiring witnessed lock contention first.
- Adds a racing-update regression test in which concurrent reinstalls genuinely retire the pinned storages while a reader snapshots them, guarding the release path under the address sanitizer.

Closes #1914.

### DIFF
--- a/lib/controlplane/agent/agent.c
+++ b/lib/controlplane/agent/agent.c
@@ -1665,7 +1665,7 @@ cp_counter_storage_copy_tags(const struct cp_counter_storage *storage) {
 	return tags;
 }
 
-// Free a per-instance snapshot array as produced by fill_counter_handle.
+// Free a per-instance value snapshot array.
 static void
 free_counter_values(uint64_t **values, uint64_t instance_count) {
 	if (values == NULL) {
@@ -1680,11 +1680,9 @@ free_counter_values(uint64_t **values, uint64_t instance_count) {
 // Snapshot a counter's per-worker values into heap memory and stash the
 // result behind the opaque counter_handle.values.
 //
-// Each worker has its own single-instance storage. The caller gathers one
-// storage per worker into worker_storages (a plain C-pointer array). The
-// values are copied out of generation-owned shm while the caller still holds
-// cp_config_lock, so the snapshot (worker_count entries) stays valid across
-// controlplane updates until it is released by yanet_counter_handle_list_free.
+// Used only for the dataplane-owned worker and port counter storages, which
+// are not generation-scoped, so no locking or pinning is needed around this
+// call.
 static int
 fill_counter_handle(
 	struct counter_handle *dst,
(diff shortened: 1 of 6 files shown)


3Hren:
@codex review

3Hren:
@codex review

3Hren:
An alternative implementation of the same fix is now up as #1955: instead of pinning individual counter storages, it adds a reference count to whole configuration generations, so the read performs zero allocations under the config lock and needs no counted release variant. This PR stays parked as the fallback approach until one of the two is chosen.

3Hren:
Closing: the generation-refcount alternative #1955 merged as the chosen implementation of #1914, making this per-storage pinning approach obsolete.

PR #1955, opened 2026-08-11
fix(counters): pin the config generation for unlocked counter value reads

### LINKED ISSUE #1914: fix(counters): stop holding the config lock across counter value copies

3Hren:
- Adds a reference count to configuration generations: the tagged counter read pins the active generation under a brief lock hold, runs matching, allocation and value copies entirely unlocked against the pinned generation, and drops the pin under another brief hold — zero allocations happen under the shared-memory config lock.
- A racing config update retires the replaced generation through the same release path once the last reader lets go, so a reader may perform the deferred teardown that the update path performs today.
- Decomposes the read into separate matching, metadata and value-copy passes, sharing one value-copy helper with the dataplane-owned worker and port counter paths.
- Adds regression tests covering the lock hold during a large read, a racing generation swap under the address sanitizer, and generation leaks across repeated swaps.
- The new field is appended at the end of the generation structure, so a running dataplane keeps reading existing offsets unchanged; control-plane processes should restart together on upgrade.
- A process killed inside the unlocked window now leaks one pinned generation instead of poisoning the lock for every other process, a strictly smaller failure than the previous dead-holder hang.
- Supersedes #1940, which stays open as the fallback approach.

Closes #1914.

### DIFF
--- a/lib/controlplane/agent/agent.c
+++ b/lib/controlplane/agent/agent.c
@@ -1677,14 +1677,48 @@ free_counter_values(uint64_t **values, uint64_t instance_count) {
 	free(values);
 }
 
+// Copy one already-described counter's per-worker value snapshot into
+// newly allocated memory, stashed behind the handle's opaque value array.
+//
+// The caller guarantees the per-worker storages stay alive for the
+// duration of this call.
+static int
+counter_handle_copy_values(
+	struct counter_handle *dst,
+	struct counter_storage **worker_storages,
+	uint64_t worker_count,
+	uint64_t idx
+) {
+	uint64_t **values = calloc(worker_count, sizeof(*values));
+	if (values == NULL) {
+		return -1;
+	}
+	for (uint64_t w_idx = 0; w_idx < worker_count; ++w_idx) {
+		if (dst->size == 0) {
+			continue;
+		}
+		values[w_idx] = malloc(dst->size * sizeof(uint64_t));
+		if (values[w_idx] == NULL) {
+			free_counter_values(values, worker_count);
+			return -1;
+		}
+		struct counter_value_handle *handle =
+			counter_get_value_handle(idx, worker_storages[w_idx]);
+		memcpy(values[w_idx],
+		       counter_handle_get_value(handle),
+		       dst->size * sizeof(uint64_t));
+	}
+	dst->values = values;
+	return 0;
+}
+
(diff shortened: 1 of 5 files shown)


3Hren:
@codex review

3Hren:
@codex review

3Hren:
@codex review

3Hren:
@codex review

PR #1964, opened 2026-08-11
perf(counters): bulk-copy counter values per worker

### LINKED ISSUE #1916: perf(counters): bulk-copy counter values per worker
## Motivation

The [per-counter copy loop](https://github.com/yanet-platform/yanet2/blob/6057bbcaa8c95e2fbaa121f69ccd3dd7176765ea/lib/controlplane/agent/agent.c#L1709-L1728) in `yanet_get_counters_by_tags` costs 24 allocations per counter today: one `calloc` for the worker pointer array, and one `malloc` per worker for that worker's value block. Against the incident's per-position counter count (78 535), this is the bulk of the roughly 16M allocations measured inside the single `cp_config_lock` hold described in #1913.

## Scope

A single allocation sized for the whole per-worker block, filled with one pass, removes almost all of these allocator round trips per scrape round.

## Acceptance

- [ ] The per-counter copy path makes one allocation for all workers' values instead of one per worker.
- [ ] Existing counter-read tests pass unchanged; a new test asserts the allocation count for an N-worker, M-counter read no longer scales as `O(N*M)` allocations.

Source: production incident investigation, 2026-08-10. See also #1913 (incident epic) and #1914 (removing the lock hold this copy happens under).

3Hren:
- The tagged counter read snapshotted each counter's values with one allocation per worker plus one more for the pointer array, which dominated the allocator traffic of a scrape over a large rule set.
- A single allocation per counter now holds the pointer array and every worker's values, so a counter's snapshot costs one allocation whatever the worker count.
- A new Go regression test drives the read through the shared-memory bindings at two worker counts, counting the allocations the read performs, and pins the per-counter cost with a floor that fails if the measurement ever stops observing the copy.

Closes #1916.

### DIFF
--- a/lib/controlplane/agent/agent.c
+++ b/lib/controlplane/agent/agent.c
@@ -1665,20 +1665,11 @@ cp_counter_storage_copy_tags(const struct cp_counter_storage *storage) {
 	return tags;
 }
 
-// Free a per-instance snapshot array as produced by fill_counter_handle.
-static void
-free_counter_values(uint64_t **values, uint64_t instance_count) {
-	if (values == NULL) {
-		return;
-	}
-	for (uint64_t i = 0; i < instance_count; ++i) {
-		free(values[i]);
-	}
-	free(values);
-}
-
 // Copy one already-described counter's per-worker value snapshot into
 // newly allocated memory, stashed behind the handle's opaque value array.
+//
+// The pointer array and every worker's value block live in one allocation:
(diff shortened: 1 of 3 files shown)


3Hren:
The regression test moved from C to Go: the C additions to the counter lock test were reverted and a new `tests/counters` package now measures the read's allocation count through the shared-memory bindings.

@codex review

PR #1984, opened 2026-08-12
fix(dataplane): propagate port start failures

### LINKED ISSUE #1944: fix(dataplane): propagate dpdk_port_start failure instead of discarding it
## Motivation

`dataplane_device_start` calls `dpdk_port_start` and drops the `int` it returns:

https://github.com/yanet-platform/yanet2/blob/87c6050ec0e3f9a496ef67edc7487ea7f9d9b0fd/dataplane/device.c#L63-L88

The rest of the function propagates failures correctly — a `dataplane_worker_start` failure a few lines later returns `-1` immediately (L80-L84) — so this is the one unpropagated start failure left in a function whose purpose, after #1941, is exactly to propagate worker/device start failures instead of proceeding past them.

If the port fails to start, the function still spawns every worker for that device and polls a dead port, and `dataplane_start` reports success to the control plane. This is the same defect class #1882/#1941 fixed for worker thread creation, one layer over: a start failure discarded instead of aborting dataplane startup.

## Scope

- [ ] Check `dpdk_port_start`'s return value and return the failure from `dataplane_device_start` instead of continuing to `dataplane_device_start_query_rss` and worker spawn.

## Acceptance

- [ ] A port-start failure aborts dataplane startup with a logged error naming the device, rather than the dataplane starting with a dead port and live workers polling it.

Source: surfaced by the reviewer sweep on PR #1941 (`fix(dataplane): refuse to start when a worker thread cannot be created`, closing #1882), which judged this out of scope for that PR.

3Hren:
- Treat any nonzero DPDK port-start result as a fatal device-start failure.
- Report the affected device and error before RSS discovery or worker creation.
- Preserve existing successful startup behavior without API, ABI, or configuration changes.
- Validate the failure path with a deterministic harness and the C build, test, lint, and fuzz gates.

Closes #1944.

### DIFF
--- a/dataplane/device.c
+++ b/dataplane/device.c
@@ -69,7 +69,15 @@ dataplane_device_start(
 	    "start dataplane device id=%u with %d workers",
 	    device->device_id,
 	    device->worker_count);
-	dpdk_port_start(device->port_id);
+	int rc = dpdk_port_start(device->port_id);
+	if (rc) {
+		LOG(ERROR,
+		    "failed to start dataplane device id=%u port=%s: error=%d",
+		    device->device_id,
+		    device->port_name,
+		    rc);
+		return -1;
+	}
 
 	// Query the NIC's RSS hash key and redirection table for later use by
 	// RSS-aware worker-affinity lookups. This is a best-effort query: no


3Hren:
Rebased onto current main with unchanged candidate content after fresh local verification and whole-candidate approval. @codex review

PR #1988, opened 2026-08-12
refactor(dataplane): route packets straight to device entry schedules

GeorgyKirichenko:
- Drops the pending input/output lists from the packet front, so a packet a module routes to another device lands straight on that device's entry schedule instead of being collected and demultiplexed on the next round.
- Adds device-entry routing helpers that preserve per-level packet accounting while placing packets directly on the target device, dropping them on the originating schedule when the device is gone from this generation.
- Worker read now deposits RX onto the target device input schedule and frees pre-configuration traffic itself, and the pipeline round polls device entry schedules directly rather than draining pending lists.
- Updates the forward, mirror, and route modules plus the unit-test and fuzz harnesses, and bumps the module ABI version for the struct layout change.

### DIFF
--- a/dataplane/worker.c
+++ b/dataplane/worker.c
@@ -50,6 +50,7 @@
 #include "lib/dataplane/pipeline/pipeline.h"
 #include "lib/dataplane/time/clock.h"
 
+#include "lib/controlplane/config/econtext.h"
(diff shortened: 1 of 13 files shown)


GeorgyKirichenko:
@codex review

PR #2009, opened 2026-08-13
fix(forward): allow concurrent metrics reads

### LINKED ISSUE #1922: fix(forward): use a read lock for metrics collection
## Motivation

[`ForwardService`](https://github.com/yanet-platform/yanet2/blob/f532febc8a22530c5eeaf5149f7650bc8e5d61df/modules/forward/controlplane/service.go#L48-L54) declares a plain `sync.Mutex`, and [`collectDataplaneMetrics`](https://github.com/yanet-platform/yanet2/blob/97e8b836d0764c44bd4e46064bc604d0bf6031e2/modules/forward/controlplane/metrics.go#L58-L69) takes it exclusively for what is a pure read of `m.configs`. In a dump taken during the incident described in #1913, 16 goroutines were queued behind a single holder stuck in a cgo call.

The route module already uses a `sync.RWMutex` with `RLock` on its read path — a working precedent for the same shape in-repo.

## Scope

Change `ForwardService.mu` to a `sync.RWMutex`, take a read lock in `collectDataplaneMetrics`, and keep the exclusive lock only where the config map is actually mutated.

## Acceptance

- [ ] `collectDataplaneMetrics` no longer excludes concurrent metrics reads from each other.
- [ ] Every mutation of `m.configs` still holds the exclusive lock.
- [ ] A regression test confirms concurrent metrics reads no longer serialize behind one another.

Source: production incident investigation, 2026-08-10. See also #1913 (incident epic) and #1919 for the same shape recurring across other modules.

3Hren:
- Lets read-only forward service operations share the config lock while preserving exclusive configuration mutation and module lifetime protection.
- Adds deterministic race-safe regression coverage proving concurrent metrics reads both reach the dataplane counter backend.
- Assumes backend counter snapshots are safe in parallel because each call pins its own configuration generation and returns call-owned copies.
- Keeps the read lock across counter snapshots so configuration replacement or deletion cannot release module state mid-read.

Closes #1922.

### DIFF
--- a/modules/forward/controlplane/metrics.go
+++ b/modules/forward/controlplane/metrics.go
@@ -56,8 +56,8 @@ func (m *ForwardService) Metrics(tags ...*commonpb.MetricTag) ([]*commonpb.Metri
 // per-module counter such as "rx" — leaves that config with nothing to
 // read.
 func (m *ForwardService) collectDataplaneMetrics(tags []*commonpb.MetricTag) ([]*commonpb.Metric, error) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
+	m.mu.RLock()
+	defer m.mu.RUnlock()
 
 	result := make([]*commonpb.Metric, 0)
 	for configName, config := range m.configs {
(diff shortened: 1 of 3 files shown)


3Hren:
Rebased onto the current `main`; the approved three-file diff is byte-identical and all affected local gates pass on head `a1dc2e8567`. @codex review

Issue #1757, opened 2026-08-06
feat(cli): add an explicit ipv4 selector to fwstate entries


3Hren:
## Motivation

`modules/fwstate/cli/src/args.rs` (`EntriesCmd`) has a single boolean `ipv6` (doc: "Use IPv6 map instead of IPv4"), aliased to short `-6`. There is no `--ipv4`/`-4` spelling — verified against current HEAD (`bb3dc1cd`), `EntriesCmd` still only declares `pub ipv6: bool`. `main.rs`'s `list_entries()` sends it straight through as `ListEntriesRequest { is_ipv6: cmd.ipv6, ... }`; `controlplane/service.go` branches on `req.GetIsIpv6()` to pick one of the two maps.

This is a selector, not a filter, and that asymmetry is the actual gap: `modules/route` (`FibShowCmd`) and `operators/route` (`RouteShowCmd`) treat `--ipv4`/`--ipv6` as filters over one dump where both-absent means "show everything". fwstate dumps exactly one family per call and IPv4 is the unnameable default — there is no `-4`, and no way to dump both families in one invocation (must run `entries` twice). A user who learned `-4`/`-6` on route will reasonably expect the same behavior here and be surprised.

Not packet-path, not urgent — CLI ergonomics/consistency only.

## Decomposition

Three options, not yet chosen between:

1. Add `--ipv4`/`-4` as an explicit peer, `conflicts_with` `--ipv6`. Cheapest, CLI-only, no proto change. The `conflicts_with` is structurally justified here because the wire field is a single bool, so "both" has no representation without a proto change.
2. Make both true filters matching route's semantics (both-absent = both families). Needs either two CLI-side streams concatenated, or a proto change (`is_ipv6` to a tri-state or repeated address-family field) plus a matching `service.go` change.
3. Replace both with `--af <ipv4|ipv6|both>` as a clap `ValueEnum`. Clearest semantics but a breaking CLI change for anyone scripting `--ipv6` today, including `tests/functional/main/fwstate_test.go` which invokes `--ipv6` literally.

Options 2 and 3 cross the CLI -> proto -> Go boundary; option 1 does not. That cost delta is the main thing to weigh before choosing.

## Acceptance

- [x] One of the three options above is chosen and recorded.
- [x] Chosen option implemented, with the family selection/filtering behavior documented in `--help` output.
[...]

PR #2040, opened 2026-08-13
feat(fwstate): filter listed state entries by address family

sakateka:
Replace the single family switch on the entries command with a pair of independent filters. Passing neither, or both, lists IPv4 then IPv6 under a shared record limit, while the cursor and direction stay per map.

Bracket IPv6 endpoints in human-readable output and widen the address columns to fit.

*Issue: https://github.com/yanet-platform/yanet2/issues/1757*

### DIFF
--- a/modules/fwstate/cli/src/args.rs
+++ b/modules/fwstate/cli/src/args.rs
@@ -138,22 +138,46 @@ pub enum DirectionArg {
     Backward,
 }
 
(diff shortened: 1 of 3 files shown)


sakateka:
@codex review

sakateka:
@codex review

3Hren (approved):

PR #2047, opened 2026-08-14
refactor(counters,acl): replace config counter registry with named module registries

GeorgyKirichenko:
Follow-up to #2042 that generalizes the single config counter registry into a set of named per-module registries.

- A module can now hold several named runtime counter registries instead of a single config registry, created lazily via cp_module_counter_registry (get-or-create by tag).
- Each registry is tagged, gets its own per-worker storage expanded on config-gen installation, and is linked against the same-tag registry of the displaced module on config rebuild so per-rule counters survive layout-preserving updates.
- Tagged storages are addressed by kind=runtime plus a config tag naming their registry, so lookups select them explicitly instead of via an absent predicate; predefined module counters keep kind=module and yanet_get_module_counters returns only those.
- The ACL module moves its per-rule counters into a dedicated rules registry, separating them from the predefined module counters.
- ACL GetMetrics no longer returns per-rule counters; the new GetRulesCounters RPC serves them, reading exactly the runtime-kind storages through the tag query, with a matching yanet-cli-acl rule-counters subcommand.
- The route and route-mpls modules move their per-route and per-nexthop counters into a dedicated routes registry. Route metrics read them through a new ModuleRuntimeCounters query beside the unchanged module-level read, and ShowFIB resolves nexthop counter names out of the new registry.
- Module ABI version is bumped for the changed shared-memory structures.

Verified with a clean meson compile, meson test 68/68, go clean -cache plus acl/controlplane/forward/route/route-mpls/bindings Go tests (including new functional coverage for GetRulesCounters and the module/rule counter read split), cargo build/clippy/fmt/test for the CLI, and the comment, proto, and clang-format lints.

### DIFF
--- a/api/counter.h
+++ b/api/counter.h
@@ -59,7 +59,9 @@ yanet_get_chain_counters(
 	const char *chain_name
 );
 
-// Get module counters, optionally filtered by name.
+// Get module counters, optionally filtered by name. Returns the module's
+// predefined counters only; its per-rule counters live on runtime-kind
+// storages read through yanet_get_counters_by_tags.
 struct counter_handle_list *
 yanet_get_module_counters(
 	struct dp_config *dp_config,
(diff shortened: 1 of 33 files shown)


GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

PR #2054, opened 2026-08-15
fix(common): detect optimized sanitizer builds

### LINKED ISSUE #2050: chore(common): non-production build check only detects AddressSanitizer
## Motivation

`common/buildinfo.h`'s `build_is_optimized()` (added by #2049, not yet merged) only checks `__OPTIMIZE__` and `HAVE_ASAN` — a build with optimizations enabled but ThreadSanitizer or UndefinedBehaviorSanitizer turned on reports as production and the `dataplane/main.c` startup WARN stays silent. Today this is harmless and unreachable: every sanitizer target this repository's own `Makefile` produces (`make setup-asan`, `make test-tsan`, etc.) also disables optimizations, so such builds still trip the optimization arm of the check. The gap becomes live the day a future sanitizer target keeps optimizations on — at that point the check silently stops doing its job, exactly the accidental-production-sanitizer-build risk #1750 introduced the warning to catch.

Related and cosmetic: two user-facing messages describe the same predicate with different precision. The `dataplane/main.c` startup WARN and the benchmark-harness warning in [bindings/go/dataplane_ut/dataplane_ut.go:615-616](https://github.com/yanet-platform/yanet2/blob/742c1f8b3d44a8a60cd8289ae35afddd1eab834e/bindings/go/dataplane_ut/dataplane_ut.go#L615-L616) both say "sanitizers" generically, while `common/buildinfo.h`'s doc comment (added by #2049; not yet on `main`, see the PR diff) states the AddressSanitizer-only semantics precisely. If the predicate's coverage is ever tightened, both user-facing messages should change together so the same check isn't described two different ways.

## Scope

- Either broaden `build_is_optimized()` to cover TSan/UBSan (a meson-sourced `YANET_SANITIZE`-style signal, deliberately left out of #2049's first cut), or explicitly document the ASan-only scope as a permanent constraint tied to the Makefile invariant it currently relies on.
- Whichever is chosen, reconcile the two warning strings with the header's actual semantics.

## Acceptance

- [ ] `build_is_optimized()` either detects TSan/UBSan builds that keep optimizations on, or the header/warning strings are made to state plainly that only ASan is covered and why that is currently safe.
- [ ] The `dataplane/main.c` startup warning and the `bindings/go/dataplane_ut` benchmark warning describe the same predicate at the same precision.

Source: deferred by the architect as non-blocking during review of #2049 (closes #1750), which added the non-production-build startup warning.

3Hren:
- Treats every Meson sanitizer mode as non-production even when optimization remains enabled.
- Preserves compiler-detected AddressSanitizer handling for fuzzing builds configured outside the Meson sanitizer option.
- Keeps startup and benchmark warnings sanitizer-generic because both continue to consume the same shared predicate.
- Assumes all non-`none` Meson sanitizer modes are unsuitable for production throughput measurements.
- Verified optimized unsanitized, ThreadSanitizer, and UndefinedBehaviorSanitizer code generation, release tests, fuzz builds, Clang syntax, and clean AddressSanitizer/CGO tests.

Closes #2050.

### DIFF
--- a/common/buildinfo.h
+++ b/common/buildinfo.h
@@ -5,7 +5,8 @@
 #include "asan.h"
 
 // Reports whether this translation unit was compiled as a production build:
-// optimizations enabled and AddressSanitizer disabled.
+// optimizations enabled and neither AddressSanitizer nor a Meson-selected
+// sanitizer enabled.
 //
 // The answer reflects the compile flags of whichever translation unit
 // includes this header, so it is only meaningful from a meson-built C
@@ -13,7 +14,7 @@
 // with its own, independent flags.
 static inline bool
 build_is_optimized(void) {
-#if defined(HAVE_ASAN)
+#if defined(HAVE_ASAN) || defined(YANET_SANITIZE)
 	return false;
 #elif defined(__OPTIMIZE__)
 	return true;
--- a/lib/dataplane_ut/dataplane_ut.h
+++ b/lib/dataplane_ut/dataplane_ut.h
@@ -107,10 +107,11 @@ dataplane_ut_run(
 );
 
 // Report whether this dataplane build is suitable for benchmarking:
-// compiled with optimizations enabled and without AddressSanitizer.
+// compiled with optimizations enabled and neither AddressSanitizer nor a
+// Meson-selected sanitizer enabled.
 //
-// Benchmarks against a debug or AddressSanitizer build produce misleading
-// numbers, so the Go harness warns when this returns 0.
+// The Go harness warns when this returns 0 because the benchmark is not
+// representative for that build.
 int
 dataplane_ut_build_optimized(void);
 
--- a/meson.build
+++ b/meson.build
@@ -89,6 +89,7 @@ yanet_cgo_cflags = []
 yanet_cgo_ldflags = []
 b_sanitize = get_option('b_sanitize')
 if b_sanitize != 'none'
+    yanet_project_args += '-DYANET_SANITIZE'
     sanitize_flags = []
     if b_sanitize.contains('address')
         sanitize_flags += 'address'


3Hren:
Addressed the GLM review against current head `fd70195ad8ab`.

- Refuted the CGO propagation concern after tracing the complete include/compile population: `common/buildinfo.h` is included only by `dataplane/main.c` and `lib/dataplane_ut/dataplane_ut.c`, both Meson-built. The Go preamble includes only the external `dataplane_ut.h` declaration and links the Meson-built implementation, so adding this define to CGO flags would not affect the predicate.
- Tightened both comments to describe exactly compiler-detected AddressSanitizer plus Meson-selected sanitizers, including the external-UBSan counterexample from the review.
- Kept permanent matrix coverage out of this low-effort change. The suggested static assertion would only restate the macro after it had arrived and would not test propagation. The exact optimized object path was instead exercised for `b_sanitize=none`, `thread`, and `undefined`, including archive disassembly and runtime verification.

A fresh independent whole-candidate review of the amended bytes is APPROVED.

3Hren:
@codex review

3Hren:
Follow-up on the refreshed GLM comment for `fd70195ad8ab`:

- The stated falsification path is the actual build path. `lib/dataplane_ut/meson.build` lists `dataplane_ut.c` in the Meson `static_library('dataplane_ut', ...)`; top-level `add_project_arguments(yanet_project_args, language: 'c')` runs before `subdir('lib')`. The cgo preamble contains no C source inclusion and links this archive via `#cgo LDFLAGS: -ldataplane_ut`. Therefore cgo never recompiles the predicate implementation, and mirroring the define into `CGO_CFLAGS` would not change it.
- The optimized `thread` and `undefined` matrix compiled that exact Meson archive, verified its compile command contained `-DYANET_SANITIZE`, and verified the exported function's object code returned 0. The optimized `none` control omitted the define and returned 1.
- The absence of a permanent CI matrix remains an explicit non-blocking residual risk. A fresh independent review accepted the real-object matrix for #2050's scope, and current-head Codex review reported no major issues.

PR #2058, opened 2026-08-15
ci(notifications): use slim GitHub runners

3Hren:
- Run Telegram notification jobs on GitHub's lightweight hosted runner.
- Preserve the existing PR and issue notification behavior.

### DIFF
--- a/.github/workflows/notifications.yml
+++ b/.github/workflows/notifications.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   notify:
     if: github.event_name == 'pull_request'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     env:
       Message: |
         🚀 PR created by ${{ github.actor }}:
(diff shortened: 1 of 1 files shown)


3Hren:
Verified against the current GitHub-hosted runner reference: `ubuntu-slim` is listed for both public and private repositories, and the current image includes `curl` and `ca-certificates`. More importantly, the `notify` job on head `e4ff5f297634` completed successfully on this PR, proving runner pickup and execution of the Telegram `curl` step: https://github.com/yanet-platform/yanet2/actions/runs/31877621720/job/94995574939. The skipped issue job uses the same runner and binary; only its unchanged event fields differ. No code change is needed.

References: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories and https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md.

Issue #2055, opened 2026-08-15
fix(acl,forward,route-mpls): avoid zero-length VLAs on an empty poll tick


3Hren:
## Motivation

#2052 fixed this class in `mirror`; it is not mirror-specific. Three more module handlers size local variable-length arrays by the per-round input count and therefore declare zero-bound arrays on every force-polled empty tick:

- [modules/acl/dataplane/dataplane.c:207-230](https://github.com/yanet-platform/yanet2/blob/8615c60b808029c9d666d63b44dc8be6195f4d19/modules/acl/dataplane/dataplane.c#L207-L230) — 11 arrays (`vlan_packets`/`vlan_result`, `ip4_packets`/`ip4_result`, `ip4_port_packets`/`ip4_port_result`, `ip6_packets`/`ip6_result`, `ip6_port_packets`/`ip6_port_result`, plus `ip6_port_pos`).
- [modules/forward/dataplane/dataplane.c:36-45](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/modules/forward/dataplane/dataplane.c#L36-L45) — 6 arrays.
- [modules/route-mpls/dataplane/dataplane.c:45-50](https://github.com/yanet-platform/yanet2/blob/aa5470361bcdde62ba880c410ec6b7de4332100b/modules/route-mpls/dataplane/dataplane.c#L45-L50) — 4 arrays.

This is confirmed by execution, not inferred from a grep: building with `-Db_sanitize=address,undefined` and running the mirror module's own empty-round test (added for #2052) emits six "variable length array bound evaluates to non-positive value 0" reports from `forward`, because the forward sink sits behind mirror in the same chain and is force-polled with an empty front. The worker force-polls every module once per tick by design — [lib/dataplane/worker/pipeline_round.c:21-30](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/lib/dataplane/worker/pipeline_round.c#L21-L30) — so an empty input front is normal steady state for every module handler, not an edge case.

Severity matches #2052: undefined behaviour in C with no observed misbehaviour, since the arrays are never indexed when the count is zero. The cost is sanitizer noise that can mask real findings on any force-polled sanitizer run — which is every one of them.

## Scope

[...]

Issue #2056, opened 2026-08-15
ci(sanitizer): fail the sanitizer job on a UBSan/ASan diagnostic instead of discarding it


3Hren:
## Motivation

Nothing today converts a UBSan/ASan diagnostic emitted during a sanitizer test run into a CI failure. #2055 (and #2052 before it) went undetected for however long the zero-length-VLA class existed because a sanitizer report on its own is not a gate — it is only visible if the offending package's tests fail outright, or someone happens to run the suite verbosely and reads the output by hand. A green sanitizer job today provides no assurance that it is actually diagnostic-free.

## Constraints

This is **blocked on #2055**: the sanitizer job currently runs the Go test runner over a package list without verbose output, so a passing package's stdout/stderr — where the diagnostic lives — is discarded today. Enabling any such gate before #2055 lands would immediately turn the sanitizer job red for `acl`, `forward`, and `route-mpls`.

The non-trivial part is visibility, not just severity: any gate has to make the test runner retain and surface sanitizer diagnostics from passing packages, not merely treat existing test failures more strictly. A naive "fail on nonzero exit" change is insufficient — a package producing UBSan output while its tests still pass needs to fail too.

## Scope

- Make the sanitizer job's Go test runner invocation capture output from every package, not only failing ones, and scan it for sanitizer diagnostic markers.
- Fail the job when a diagnostic is found, even if all tests in that package passed.
- Land only after #2055 closes, so the gate starts clean.

## Acceptance

- [ ] The sanitizer CI job fails when any package under test emits a UBSan or ASan diagnostic, independent of that package's own test pass/fail result.
- [ ] A deliberately reintroduced instance of the #2055 class (or an equivalent synthetic diagnostic) is verified to redden the job before landing.

Source: noted alongside #2055 while fixing #2052 — the fix removes the known instances of the defect class, but nothing stops a new one from landing silently, since a passing sanitizer run today discards diagnostic output from packages whose tests pass.

3Hren:
Carried over from #2060 (closed as a duplicate of this issue), which added execution-proven detail this issue didn't yet have:

[...]

PR #2062, opened 2026-08-15
fix(forward): avoid zero-length VLAs on an empty poll tick

3Hren:
The worker force-polls every module in a non-empty chain once per tick regardless of input, so a handler's per-round input count is legitimately zero in normal steady state rather than as an edge case. Sizing the handler's six local variable-length arrays directly by that count declared them with zero length on every such tick, which C leaves undefined.

The handler now reads the count once, returns early when it is zero, and sizes the arrays by that local. This is the fix already merged for the mirror module, applied unchanged. Returning early skips nothing: this handler does no work after its packet loop, and the counters read before the guard are pointer lookups with no side effects. The per-tick bookkeeping belongs to the caller, which already guards its own histogram on a non-zero packet count.

Measured rather than inferred. A sanitizer build of the base emitted six variable-length-array-bound reports from this handler, reached with an empty front from seven independent test packages. The same build of this change emits none, with the instrumentation confirmed still compiled in so that the silence is real.

No new test accompanies this change, deliberately. Those seven packages already reach this handler with an empty front, so a forward-owned empty-round test would duplicate an existing vehicle rather than add coverage.

The fuzz build was not run locally, because the local target wipes and rebuilds the whole tree while this change alters no signature or symbol. The fuzz workflow is nightly rather than pull-request triggered, so it was dispatched manually against this branch instead.

Part of #2055.

### DIFF
--- a/modules/forward/dataplane/dataplane.c
+++ b/modules/forward/dataplane/dataplane.c
@@ -33,16 +33,26 @@ forward_handle_packets(
 		cp_module
 	);
 
-	struct packet *vlan_packets[packet_front_input_count(packet_front)];
-	uint32_t vlan_result[packet_front_input_count(packet_front)];
+	// A force-polled tick can hand this module an empty front, and sizing
+	// the arrays below directly by that count would then declare them with
+	// zero length, which is undefined behavior. The early return is only
+	// safe because nothing runs after this handler's final packet loop —
+	// any code added after that loop must run before the return.
+	uint64_t count = packet_front_input_count(packet_front);
+	if (count == 0) {
+		return;
+	}
+
+	struct packet *vlan_packets[count];
+	uint32_t vlan_result[count];
 	uint64_t vlan_idx = 0;
 
-	struct packet *ip4_packets[packet_front_input_count(packet_front)];
-	uint32_t ip4_result[packet_front_input_count(packet_front)];
+	struct packet *ip4_packets[count];
+	uint32_t ip4_result[count];
 	uint64_t ip4_idx = 0;
 
-	struct packet *ip6_packets[packet_front_input_count(packet_front)];
-	uint32_t ip6_result[packet_front_input_count(packet_front)];
+	struct packet *ip6_packets[count];
+	uint32_t ip6_result[count];
 	uint64_t ip6_idx = 0;
 
 	for (struct packet *packet = packet_list_first(&packet_front->input);


3Hren:
Both review points addressed.

**Unpinned invariant behind the early return** — taken, and applied to all three modules in this series rather than only here. The comment now states what keeps the guard correct, not just why it exists: the early return is sound only because nothing runs after the handler's final packet loop, so trailing work added later must run before the return. It says "final packet loop" deliberately, since these handlers have a gather loop and a pop loop with the filter queries in between, and the unqualified phrasing was falsifiable in one lookup. The merged mirror fix carries the earlier wording and could be brought into line separately.

**Regression guard is sanitizer-only** — correct, and worth stating precisely. The sanitizer configuration does run in CI and passes on this pull request, so the measurement is not local-only. But nothing converts a sanitizer diagnostic into a failure: it defaults to reporting and continuing, and the single place this repository sets that option sets it to continue. Verified by execution — a zero-length array compiled with the repository's own sanitizer flags prints its diagnostic and still exits zero. So reintroducing this class would print and stay green, and this class is indeed protected only by review today. That gap is tracked as #2056, which is blocked on #2055 precisely so the gate starts clean.

One note for the record, since this pull request was red for a while and the failure looked incriminating. `TestForward/Test_ICMP6_Echo` failed twice here while main passed at the same base. It is not caused by this change: the subtest echoes to an autoconfigured link-local address and can race duplicate-address detection, during which the guest silently drops packets addressed to it — the captured trace shows the guest's neighbour-discovery traffic arriving after the echo request had already been sent. The IPv4 sibling sets its address explicitly and cannot hit it, which explains the family asymmetry. Two passing runs on unrelated branches went through the identical restart path, and a fresh run here is green. Filed as #2063, with the upstream snapshot problem as #2064. Re-running only the failed jobs reuses the same baked baseline, so it reproduces rather than resamples — worth knowing before reading two failures as determinism.

@codex review

3Hren:
On the tail claim — the demand is fair, and it has been discharged. It is not verifiable from the hunk, which is exactly why it was checked against the function rather than the diff.

The handler ends at its pop loop. After the loop's closing brace comes the function's closing brace and nothing else — no per-call bookkeeping, no aging tick, no front manipulation, no dispatch. Three independent passes established this: a reviewer read each of the three handlers to its closing brace, a dynamic-analysis pass traced it separately, and the same pass confirmed the caller owns all per-tick work — it stamps device identifiers, adds the receive counters, brackets the handler for timing, and derives transmit, drop and pending counters from the front afterwards, with its own latency histogram already guarded on a non-zero packet count. So the tail that would be skipped is empty, and the bookkeeping that must not be skipped was never in the handler.

The downstream-cascade concern is answered by the same structure. Modules do not poll each other. The worker force-polls every module in a non-empty chain once per tick by design, so a module behind this one is reached whether or not this handler runs, and skipping an empty front here cannot starve it. That is the property that made an empty front normal steady state in the first place.

On the suggested alternative of clamping the size to one: declined, with reasons rather than preference. It preserves the tail's execution, but the tail is empty, so it preserves nothing while allocating six one-element arrays and running three classification queries on every idle tick — which is steady state, on the packet path. It also diverges from the merged mirror fix, leaving two shapes for one defect class across four modules. The guard is the shape the issue specifies and the one already on main.

On the second point — the sanitizer configuration is PR-triggered and passes on this pull request, so the measurement is not local-only. But the distinction drawn is the right one and I want to agree with it explicitly rather than let it read as answered: those packages reach the handler with an empty front, they do not detect a reintroduction, because a sanitizer diagnostic does not fail anything here. A forward-owned test would not change that — it would reach, not detect. What closes the gap is the gate in #2056, and until it lands this class is protected by review. That is stated plainly rather than papered over with a test that would only look like coverage.

Issue #2061, opened 2026-08-15
fix(route-mpls): invalid filter verdict drops the popped packet without output or drop


3Hren:
## Motivation

In [modules/route-mpls/dataplane/dataplane.c](https://github.com/yanet-platform/yanet2/blob/aa5470361bcdde62ba880c410ec6b7de4332100b/modules/route-mpls/dataplane/dataplane.c#L88-L108), inside the pop loop of `route_mpls_handle_packets`, both address families do the same thing on an invalid filter verdict: the packet is popped off the input list at [line 88](https://github.com/yanet-platform/yanet2/blob/aa5470361bcdde62ba880c410ec6b7de4332100b/modules/route-mpls/dataplane/dataplane.c#L88), then on `FILTER_RULE_INVALID` the code advances its index and `continue`s — [IPv4 at lines 93-96](https://github.com/yanet-platform/yanet2/blob/aa5470361bcdde62ba880c410ec6b7de4332100b/modules/route-mpls/dataplane/dataplane.c#L93-L96), [IPv6 at lines 105-108](https://github.com/yanet-platform/yanet2/blob/aa5470361bcdde62ba880c410ec6b7de4332100b/modules/route-mpls/dataplane/dataplane.c#L105-L108) — without ever calling `packet_front_output` or `packet_front_drop`. The packet is neither forwarded nor dropped through the packet_front machinery: it is simply lost, which looks like an mbuf leak on that path. Every other exit from the loop (`target == NULL`, zero nexthop map, encap failure, success) calls one of the two.

Raised by the reviewer as out of scope for #2055 and independently confirmed by reading the code during this session.

## Constraints

Untestable in-process today: route-mpls is absent from the `dataplane_ut` link group (see the companion coverage-gap issue this session also filed), so this path's reachability and the leak's real-world rate have not been established by execution — only by code reading.

## Acceptance

- [ ] Bug-hunter confirmation: reproduce the loss/leak (in-process test once coverage exists, or a targeted probe) rather than relying on code reading alone.
- [ ] The invalid-verdict branch calls `packet_front_drop` (or `packet_front_output`, whichever is correct for this verdict) for both IPv4 and IPv6 before `continue`.
- [ ] A regression test exercises an invalid verdict and asserts the packet is accounted for (dropped or output), not silently lost.

Source: raised by the reviewer during #2055 as out of scope for that fix, independently confirmed by reading modules/route-mpls/dataplane/dataplane.c.

Issue #2059, opened 2026-08-15
test(route-mpls): wire module into dataplane_ut so in-process tests can execute it


3Hren:
## Motivation

`route-mpls` is absent from the cgo link group in [bindings/go/dataplane_ut/dataplane_ut.go](https://github.com/yanet-platform/yanet2/blob/742c1f8b3d44a8a60cd8289ae35afddd1eab834e/bindings/go/dataplane_ut/dataplane_ut.go#L10-L45): neither its library search path (`-L.../modules/route-mpls/dataplane`) nor its `-lroute_mpls_dp` flag is present, and `new_module_route_mpls` is missing from the [extern declarations and module table](https://github.com/yanet-platform/yanet2/blob/742c1f8b3d44a8a60cd8289ae35afddd1eab834e/bindings/go/dataplane_ut/dataplane_ut.go#L115-L139) — acl, forward and mirror are all present there. The module also has no `tests/` and no `fuzzing/` directory. Its only test vehicle in the repository is `tests/functional/main/route_mpls_test.go`, the QEMU VM suite, which runs the real binary uninstrumented and without a sanitizer build.

Measured during #2055: a UBSan sweep over the whole Go test suite produced zero diagnostics from route-mpls, not because it was clean but because nothing executes it in-process. That fix's route-mpls half had to be published verified by inspection rather than by execution — the weakest of the three modules #2055 touched.

## Scope

- Add `-L../../../build/modules/route-mpls/dataplane` and the correct `-l` flag to the link group in `dataplane_ut.go`, plus the `new_module_route_mpls` extern declaration and module-table entry.
- Add a `modules/route-mpls/tests/functional` (or equivalent in-process) package exercising the module the way acl/forward/mirror are exercised.

## Constraints

The built static library is named with a hyphen, `libroute_mpls-dp.a` — [modules/route-mpls/dataplane/meson.build:13-14](https://github.com/yanet-platform/yanet2/blob/b98b0fe10ff72a61d76e04ccb28eab15bdf7a733/modules/route-mpls/dataplane/meson.build#L13-L14) names the static library `route_mpls-dp`, while every sibling module (`acl_dp`, `forward_dp`, `mirror_dp`, …) uses an underscore. The link flag must match `-lroute_mpls-dp`, or the meson library must be renamed to `route_mpls_dp` first, before the new `-l` flag is added.

[...]

PR #2065, opened 2026-08-15
fix(route-mpls): avoid zero-length VLAs on an empty poll tick

3Hren:
The worker force-polls every module in a non-empty chain once per tick regardless of input, so a handler's per-round input count is legitimately zero in normal steady state rather than as an edge case. Sizing the handler's four local variable-length arrays directly by that count declared them with zero length on every such tick, which C leaves undefined.

The handler now reads the count once, returns early when it is zero, and sizes the arrays by that local. This is the shape already merged for the mirror and forward modules. The comment states the invariant the guard rests on as well as the reason for it: the early return is sound only because nothing runs after the handler's final packet loop, so trailing work added later must run before the guard. That claim was checked against the function rather than the diff, and the per-tick bookkeeping lives in the caller, which already guards its own histogram on a non-zero packet count.

One limitation is worth stating plainly rather than leaving to be discovered. Unlike its two siblings, this fix is not backed by an execution measurement. A sanitizer sweep of the base emitted zero reports from this module — not because it is clean, but because nothing executes it: the module is absent from the link list the in-process test harness uses, it has no test or fuzzing directory, and its only vehicle is the virtual-machine functional suite, which runs uninstrumented. So this change is verified by inspection and by being byte-for-byte the same guard as the two modules where the reports did appear and then disappeared. Giving the module a real in-process vehicle is tracked as #2059.

While reading this handler for that check, a separate pre-existing defect surfaced: on an invalid filter verdict the packet has already been popped from the input list and is then skipped without being forwarded or dropped, so it is neither. That is filed as #2061 and is deliberately not addressed here.

Part of #2055.

### DIFF
--- a/modules/route-mpls/dataplane/dataplane.c
+++ b/modules/route-mpls/dataplane/dataplane.c
@@ -42,12 +42,22 @@ route_mpls_handle_packets(
 		cp_module
 	);
 
-	struct packet *ip4_packets[packet_front_input_count(packet_front)];
-	uint32_t ip4_result[packet_front_input_count(packet_front)];
+	// The worker's per-tick force-poll can reach this handler with an empty
+	// front, and the arrays below would then be declared with zero length,
+	// which is undefined behavior. The early return is only safe because
+	// nothing in this handler runs after its final packet loop — any code
+	// added after that loop must run before the return.
+	uint64_t count = packet_front_input_count(packet_front);
+	if (count == 0) {
+		return;
+	}
+
+	struct packet *ip4_packets[count];
+	uint32_t ip4_result[count];
 	uint64_t ip4_idx = 0;
 
-	struct packet *ip6_packets[packet_front_input_count(packet_front)];
-	uint32_t ip6_result[packet_front_input_count(packet_front)];
+	struct packet *ip6_packets[count];
+	uint32_t ip6_result[count];
 	uint64_t ip6_idx = 0;
 
 	for (struct packet *packet = packet_list_first(&packet_front->input);


3Hren:
Three notes, all answerable without a change.

**Coverage gap** — agreed as stated, and the sequencing suggestion is the right one. When #2059 gives this module an in-process vehicle, the empty-front call is worth adding across the siblings together rather than only here. Noting for that issue that such a test reaches the path but does not detect a reintroduction until #2056 makes a sanitizer diagnostic fail the run — the two are complementary, and a test added before the gate would look like coverage it does not yet provide.

**The invariant, confirmed** — the handler ends at its packet loop. The last statement inside the loop hands the packet to the output front, then the loop closes and the function closes immediately after it. No trailing bookkeeping, nothing to skip. The per-tick work lives in the caller, which stamps device identifiers, adds the receive counters, times the handler and derives the transmit, drop and pending counters from the front afterwards, with its own histogram already guarded on a non-zero packet count. Verified by reading the function, not the hunk.

**Sibling sweep — done, and this pull request plus its remaining sibling closes the class.** The defining search for an array sized directly by the input-count call across the module, device, library and dataplane trees returns nothing on this branch, which carries the last outstanding module's fix alongside this one. Mirror was fixed in #2057, forward in #2062, this is the third, and the fourth is the acl change that follows — after it lands the pattern has no remaining instance. Worth adding that acl was the one that needed care: it holds nineteen such arrays rather than the four here, eight of them further down on a conditional path that no test in the repository reached, so a guard placed at the first declaration rather than at the top of the function would have left those undefined.

PR #2066, opened 2026-08-15
fix(acl): avoid zero-length VLAs on an empty poll tick

### LINKED ISSUE #2055: fix(acl,forward,route-mpls): avoid zero-length VLAs on an empty poll tick

3Hren:
The worker force-polls every module in a non-empty chain once per tick regardless of input, so a handler's per-round input count is legitimately zero in normal steady state rather than as an edge case. Sizing the handler's local variable-length arrays directly by that count declared them with zero length on every such tick, which C leaves undefined.

This module needed more care than its siblings. It has nineteen such arrays, not the eleven visible in the obvious block: eight more sit further down on the shared-classification path, sized by a second local that re-reads the same count. A guard placed at the first declaration would have left those eight undefined, and nothing in the repository would have noticed. The guard therefore sits above every declaration, so one return dominates all nineteen.

Measured rather than inferred. A sanitizer build of the base emitted eleven variable-length-array-bound reports from this handler; the same build of this change emits none, with the instrumentation confirmed still compiled in so that the silence is real. The eight on the shared-classification path emitted nothing anywhere, because reaching them needs the shared trie built and an empty front at the same time and no test supplied both.

That is what the new test is for. It compiles a ruleset that builds the shared trie, asserts the trie really was built rather than trusting the ruleset to have done it, and then runs one force-polled round with no packets. Before it, this module's own suite never reached its handler with an empty front — the only vehicle in the repository was a pipeline test nobody would associate with this module. Run against the pre-fix handler the new test lights all nineteen sites, including the eight, so it is a real vehicle and not a decoration.

It cannot assert the defect directly, and its doc comment says so rather than implying otherwise. A zero-length array here is undefined behaviour the handler never indexes, and on an empty round every module counter takes a zero delta while the per-module latency histogram is skipped entirely. What the test asserts instead is the reject path's one observable postcondition, that the module's whole counter vector is untouched — which also pins the caller's own guard against timing an empty round. Its value is reach: under a sanitizer build it is what turns a regression into a diagnostic. It is not a gate, because nothing here converts a sanitizer diagnostic into a failure — that gap is #2056.

With this the defect class is closed. The defining search for an array sized directly by the input-count call now returns nothing across the module, device, library and dataplane trees, mirror having been fixed in #2057, forward in #2062 and route-mpls in #2065.

Closes #2055.

### DIFF
--- a/modules/acl/dataplane/dataplane.c
+++ b/modules/acl/dataplane/dataplane.c
@@ -204,30 +204,39 @@ acl_handle_packets(
 	 * For the second option we have to split v4 and v6 processing.
 	 */
 
-	struct packet *vlan_packets[packet_front_input_count(packet_front)];
-	uint32_t vlan_result[packet_front_input_count(packet_front)];
+	// A force-polled tick can reach this handler with an empty front, and
+	// zero-sizing every variable-length array declared below is undefined
+	// behavior. The early return is only safe because everything below is
+	// per-packet — trailing work added later must go above the guard.
+	uint64_t count = packet_front_input_count(packet_front);
+	if (count == 0) {
+		return;
+	}
+
+	struct packet *vlan_packets[count];
+	uint32_t vlan_result[count];
 	uint64_t vlan_idx = 0;
 
-	struct packet *ip4_packets[packet_front_input_count(packet_front)];
-	uint32_t ip4_result[packet_front_input_count(packet_front)];
+	struct packet *ip4_packets[count];
+	uint32_t ip4_result[count];
 	uint64_t ip4_idx = 0;
 
-	struct packet *ip4_port_packets[packet_front_input_count(packet_front)];
-	uint32_t ip4_port_result[packet_front_input_count(packet_front)];
+	struct packet *ip4_port_packets[count];
+	uint32_t ip4_port_result[count];
 	uint64_t ip4_port_idx = 0;
 
-	struct packet *ip6_packets[packet_front_input_count(packet_front)];
-	uint32_t ip6_result[packet_front_input_count(packet_front)];
+	struct packet *ip6_packets[count];
+	uint32_t ip6_result[count];
 	uint64_t ip6_idx = 0;
 
-	struct packet *ip6_port_packets[packet_front_input_count(packet_front)];
-	uint32_t ip6_port_result[packet_front_input_count(packet_front)];
+	struct packet *ip6_port_packets[count];
+	uint32_t ip6_port_result[count];
 	uint64_t ip6_port_idx = 0;
 
 	// Position of each ip6_port batch packet within the ip6 batch, filled
 	// only on the shared-classification path where every ip6_port packet
 	// is by construction also an ip6 packet.
-	uint32_t ip6_port_pos[packet_front_input_count(packet_front)];
+	uint32_t ip6_port_pos[count];
 
 	for (struct packet *packet = packet_list_first(&packet_front->input);
 	     packet != NULL;
@@ -287,17 +296,14 @@ acl_handle_packets(
 	);
 
 	if (net6_share) {
-		const uint32_t batch_count =
-			packet_front_input_count(packet_front);
-
 		struct net6_share_dir *share_src = &acl_config->net6_share_src;
 		struct net6_share_dir *share_dst = &acl_config->net6_share_dst;
 
 		// Classify each v6 address half once on the union tries.
-		uint32_t src_hi[batch_count];
-		uint32_t src_lo[batch_count];
-		uint32_t dst_hi[batch_count];
-		uint32_t dst_lo[batch_count];
+		uint32_t src_hi[count];
+		uint32_t src_lo[count];
+		uint32_t dst_hi[count];
+		uint32_t dst_lo[count];
 
 		for (uint64_t idx = 0; idx < ip6_idx; ++idx) {
 			struct rte_mbuf *mbuf =
@@ -338,8 +344,8 @@ acl_handle_packets(
 		struct net6_classifier *ip6_dst_cls = (struct net6_classifier *)
 			ADDR_OF(&acl_config->filter_ip6.v[ip6_dst_leaf].data);
 
-		uint32_t ip6_src_slots[batch_count];
-		uint32_t ip6_dst_slots[batch_count];
+		uint32_t ip6_src_slots[count];
+		uint32_t ip6_dst_slots[count];
 
 		for (uint64_t idx = 0; idx < ip6_idx; ++idx) {
 			ip6_src_slots[idx] = value_table_get(
@@ -371,8 +377,8 @@ acl_handle_packets(
 					 .data
 			);
 
-		uint32_t ip6_port_src_slots[batch_count];
-		uint32_t ip6_port_dst_slots[batch_count];
+		uint32_t ip6_port_src_slots[count];
+		uint32_t ip6_port_dst_slots[count];
 
 		for (uint64_t idx = 0; idx < ip6_port_idx; ++idx) {
 			uint32_t pos = ip6_port_pos[idx];
--- a/modules/acl/tests/functional/acl_net6_share_test.go
+++ b/modules/acl/tests/functional/acl_net6_share_test.go
@@ -462,3 +462,93 @@ func TestACL_Net6Share_VerdictParity_ScaleHeavy(t *testing.T) {
 	)
 	runNet6ShareVerdictParity(t, 40000, 4096, scaleCPMemory, scaleAgentMemory)
 }
+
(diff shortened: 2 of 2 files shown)


3Hren:
All six findings from both reviewers addressed.

**The test passed identically whether or not the handler ran** — the important one, and correct. Empty output, empty drops and an all-zero counter vector hold just as well if the harness never force-polls at all, so the test would have degraded silently if that contract ever shifted. There is now a positive control after the empty round: one packet chosen to match exactly one rule, then an assertion that that rule's counter advanced. It was verified by measurement rather than by reading the rule set — only the intended counter moves, its baseline zero is the very value the vector assertion above already pins, and the packet is port-scoped so its verdict is produced through the shared-classification path, riding the one array that exists only there. So the control exercises the branch this test is about, not an ordinary path.

**The assertion message stated a contract the code does not owe** — fixed. It now describes current behaviour and explicitly declines to call a future per-round counter a defect, which is exactly what the guard's own remedy invites someone to add.

**The guard rationale was over budget** — fixed, 6 prose lines down to 4. The enumeration of the shared-classification arrays is gone because the sentence quantifies over every variable-length array declared below, which covers them without listing them, and enumerated cases are the first thing the convention says to cut. Worth noting the domain is unchanged: the enumeration was emphasis inside an already-universal claim, not an extension of it.

**Abbreviated locals** — spelled out in the new test. The pre-existing siblings keep theirs, since the convention governs new code rather than rewriting neighbours.

**The doc comment overclaimed** — fixed. It no longer says the test would catch a regression, since in an ordinary build nothing converts a sanitizer diagnostic into a failure and it would catch nothing. The exclusivity claim was re-grepped rather than assumed and holds: this is the only zero-packet round in the module's test tree, and no acl test uses a multi-device topology, which was the other route to an empty front.

**On the environment check failing rather than skipping** — deliberate, and matching the precedent directly above it in the same file. The parity test already fails on the same condition for the same reason: the kill switch keys off the variable being present rather than its value, so if it is set the test silently compiles the unshared path and covers eight fewer arrays while still reporting success. A skip trades a loud failure for silent miscoverage, which is the failure mode this whole change exists to remove. Nothing in CI sets it.

@codex review

3Hren:
Both rounds of findings addressed. Three changes.

**The eight shared-path arrays are no longer protected positionally.** The separate local that re-read the input count is gone, and all nineteen arrays are now sized by the one hoisted local. The concern that a non-empty round could reach those declarations with size zero does not arise today — the classification loop only walks the list and never diverts a packet, so the first removal happens far below — but that made their safety an argument a reader has to reconstruct rather than something they can see. It is now the guarded variable itself, by name, so the compiler enforces what a comment used to assert. I had declined this as out of scope earlier and was wrong to.

It was verified as a pure substitution rather than assumed: eight sites, no other use of the removed local, no loop bound or argument depending on its narrower type. At the instruction level the two builds differ in 140 of 7,646 instructions and every difference is a 32-to-64-bit width change — no arithmetic, no scale factor, no branch structure, no call — with the instruction count falling by exactly the deleted zero-extensions. The only direction the widening could change behaviour is above the 32-bit limit, where the old code truncated and under-allocated, so it can remove a latent overflow but not introduce one. Behaviourally, both shared-versus-unshared parity tests pass at 384 and 12,000 rules, comparing packet-by-packet and counter-by-counter over exactly these arrays.

**The positive control's comment claimed more than the control proves** — correct, and caught independently by both reviewers. The control invokes a separate packet-bearing round, so a harness that returned early on a zero-packet call would pass every assertion in this test. It therefore establishes nothing about the empty round. That was my error in specifying it. Since no empty-round signal is observable from Go at all, the fix is to narrow the claim rather than manufacture an observation, and deliberately not to add a production counter to create one. The comment now claims only what the control establishes: the wiring is live, this rule set really does route through the shared-classification path, and the guard did not break the non-empty path. That last one earns more than it did before, since this revision changes executable code on exactly that path.

**Abbreviated locals in the packet-building block** are spelled out, including the three protocol layer variables.

On the environment precondition failing rather than skipping — answered above and unchanged: the kill switch keys off presence rather than value, so with it set the test silently compiles the unshared path and covers eight fewer arrays while still reporting success. A skip trades a loud failure for exactly the silent miscoverage this change exists to remove, and the parity test directly above it already fails on the same condition for the same reason.

@codex review

3Hren:
Three remaining notes, all confirmation requests rather than defects. Answering with what was checked rather than changing anything.

**The function tail** — confirmed, and this is the fourth independent check of it across this series. The handler ends at its packet loop: the last branch bumps a no-match counter and drops the packet, the loop closes, and the function closes immediately after. There is no epilogue — no free, no flush, no telemetry bump. The per-tick work lives in the caller, which stamps device identifiers, adds the receive counters, times the handler, and derives the transmit, drop and pending counters from the front afterwards, with its own histogram already guarded on a non-zero packet count. The note is right that this is the one place a silently skipped side effect could hide, which is exactly why the comment states the invariant rather than leaving it implicit.

**The environment guard** — the premise here is mistaken, so worth correcting rather than just answering. The rest of this file does not tolerate the variable with a skip: the parity test directly above uses the same hard failure, for the same reason. The only skip in the file guards a different variable entirely, an opt-in for a production-scale run, which is a genuine opt-in rather than a coverage switch. And nothing in the workflows or the makefile sets the kill switch, which was the condition attached to the note. The hard failure is the defensible case: with the variable set this test compiles the unshared path, silently covers eight fewer arrays and still reports success, which is precisely the silent-miscoverage failure this change exists to remove.

**The empty-round half being vacuous without force-poll** — conceded, stated in the test's own documentation, and unchanged by anything available here. There is no empty-round signal observable from Go: every module counter takes a zero delta and the per-module latency histogram is skipped entirely. That is a property of the medium, not an omission, and the honest response is the one already in the code — the test documents that it buys reach rather than detection, and the positive control's comment was narrowed in the previous revision precisely so it does not claim to cover this gap. Manufacturing an observation would mean adding a production counter to serve a test, which is worse than the gap it closes. What actually closes it is #2056, making a sanitizer diagnostic fail the run, which is blocked on this issue and unblocks when it merges.

PR #2067, opened 2026-08-15
refactor(gateway): enforce package encapsulation

### LINKED ISSUE #1733: refactor(controlplane): encapsulate gateway and bundle reach-ins and externalize gateway's test package
## Motivation

The API gateway proxies every module backend; its exported-surface decisions are also read by
the client-side registrar/registration-loop helpers that import it. `private:` rows re-counted
directly against `lint/style/allowlist.txt` today: `gateway.go` (8 — `Gateway.Run` reaching
`runner.module`, `NewGateway` reaching `entry.kind` twice and `entry.service` five times),
`registry.go` (7 — `BackendRegistry.GetBackend`/`ListBackends`/`Renew`/`registerBackend` reaching
`entry.backend`/`.service`/`.lastSeenAt` and their `existing.*` counterparts),
`metrics_service.go` (1 — `metricsCollectors` reaching `entry.service`), and `bundle/bundle.go`
(2 — `buildServices` reaching `factory.name`/`.new`) — 18 total, not the 12 this item was
originally filed under; the count has drifted upward twice as later work reshaped `gateway.go`
and `registry.go`.

Two earlier merged changes already retired part of this pile by giving the registry's
backend-entry type exported accessors — one added an exported close method for a new eviction
sweep, another paid one more `registry.go` row incidentally. The same shape (exported accessors
on the entry type rather than field reach-ins) should retire what remains.

There are also 4 `testpkg:` rows: `gateway_test.go`, `readiness_test.go`, `registry_test.go`,
and `runner_test.go` all need to move from `package gateway` to an external `gateway_test`
package.

## Acceptance

- [ ] Re-count both checks' rows for these files in `lint/style/allowlist.txt` before starting
      (current: 18 `private:` + 4 `testpkg:`, verified 2026-08-06).
- [ ] All 18 `private:` rows fixed via exported methods/fields.
- [ ] All 4 gateway test files moved to `package gateway_test`.
- [ ] All 22 corresponding ledger rows deleted in the same change.
- [ ] `go test ./controlplane/...` still green, including the client-side helper packages that
      import the gateway.

Source: migrated from the retired local task tracker's task filed 2026-07-17, on 2026-08-06.

3Hren:
- Encapsulates gateway service, runner, and backend-entry access without changing registration, timestamp, ownership, or logging semantics.
- Migrates the four gateway white-box test files to the external package and preserves their lifecycle and registry contracts through public APIs.
- Adds a deterministic external regression test for the gateway drain latch; removing the latch reproduces a forbidden late return to READY.
- Removes the 16 remaining gateway private rows and four test-package rows; the two bundle rows named by the original issue were already resolved on main.

Closes #1733.

### DIFF
--- a/controlplane/gateway/gateway.go
+++ b/controlplane/gateway/gateway.go
@@ -57,8 +57,8 @@ type ClosableService interface {
 
 // serviceEntry pairs a service with its declared backend kind.
 type serviceEntry struct {
-	service Service
-	kind    BackendKind
+	Service Service
+	Kind    BackendKind
 }
 
 // defaultPerServiceMethodLimit caps the number of distinct grpc_method label
@@ -96,14 +96,14 @@ type GatewayOption func(*gatewayOptions)
 // WithService adds an in-process module or device service to the Gateway.
 func WithService(service Service) GatewayOption {
 	return func(o *gatewayOptions) {
-		o.Services = append(o.Services, serviceEntry{service: service, kind: BackendKindInProcess})
+		o.Services = append(o.Services, serviceEntry{Service: service, Kind: BackendKindInProcess})
 	}
 }
 
 // WithBuiltinService adds a framework service that shares the gateway's own gRPC server.
 func WithBuiltinService(service Service) GatewayOption {
 	return func(o *gatewayOptions) {
-		o.Services = append(o.Services, serviceEntry{service: service, kind: BackendKindBuiltin})
+		o.Services = append(o.Services, serviceEntry{Service: service, Kind: BackendKindBuiltin})
 	}
 }
 
@@ -370,26 +370,26 @@ func NewGateway(cfg *Config, options ...GatewayOption) (*Gateway, error) {
 	var serviceRunners []*ServiceRunner
 
 	for _, entry := range opts.Services {
-		if entry.service.Endpoint() == "" {
+		if entry.Service.Endpoint() == "" {
 			// Shared-server service: register on the gateway's gRPC server and
 			// point every service name at the shared loopback backend using the
 			// declared kind.
-			entry.service.RegisterService(server)
+			entry.Service.RegisterService(server)
 
-			for _, name := range entry.service.ServicesNames() {
-				registry.RegisterBackend(name, loopback, entry.kind)
+			for _, name := range entry.Service.ServicesNames() {
+				registry.RegisterBackend(name, loopback, entry.Kind)
 				log.Info("registered service in registry",
 					zap.String("service", name),
-					zap.Stringer("kind", entry.kind),
+					zap.Stringer("kind", entry.Kind),
 				)
 			}
 		} else {
 			// Out-of-process: wrap in a ServiceRunner.
-			runner := NewServiceRunner(entry.service, endpoint, cfg.Server.TLS, WithServiceRunnerLog(log))
+			runner := NewServiceRunner(entry.Service, endpoint, cfg.Server.TLS, WithServiceRunnerLog(log))
 			serviceRunners = append(serviceRunners, runner)
 		}
 
-		services = append(services, entry.service)
+		services = append(services, entry.Service)
 	}
 
 	return &Gateway{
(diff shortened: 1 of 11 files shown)


3Hren:
Addressed all GLM findings in `191d2c67`:

- Restored the concrete gRPC backend close-idempotency contract through the public `GatewayService.Register` path and verifies the real connection reaches `Shutdown` after repeated close.
- Confirmed the late readiness publication runs in `Gateway.Run`'s tracked errgroup; added a five-second Watch deadline so notification failure is bounded and diagnosable.
- Removed the exported lock-dependent `BackendEntry.RefreshIfEndpoint` mutator; renewal is now owned by a private `BackendRegistry` helper under its write lock and preserves backend, service, kind, and refcount.

Fresh local whole-candidate review is APPROVED; full `make test`, targeted race tests, build, vet, stylelint, and comment lint pass.

3Hren:
@codex review

Please re-review the current head `191d2c67` after the three addressed GLM findings.

3Hren:
Addressed the remaining GLM minors in `06d05797`:

- Renamed the preserve-mode test to state only its observable public contract: keeping a stale external registration visible.
- Consolidated all readiness-scope test values, filters, and assertions behind one test-local constant; no production API was exported.
- Kept the deterministic log-hook seam unchanged, consistent with GLM's note that its cleanup and failure paths are sound.

Fresh whole-candidate review is APPROVED and the committed fingerprint matches it exactly.

@codex review

3Hren:
Closed the final GLM coverage deltas in `dd39f0b9`:

- The preserve-mode test now registers a real gRPC backend, waits beyond TTL/sweep intervals, then successfully routes a proxied RPC into its handler. This proves the preserved connection remains open and usable, not merely listed. Immediate ticker return remains intentionally internal and is no longer claimed by the test.
- After `Gateway.Run` returns, the latch test now verifies the sole terminal transition is still `NOT_READY` with `SHUTTING_DOWN`, in addition to proving no late READY transition occurred.
- The strict `Before` cutoff remains an intentional boundary tripwire.

Fresh whole-candidate review is APPROVED; the committed fingerprint matches exactly.

@codex review

3Hren:
Addressed the final GLM medium finding in `85d37c58`:

- `TestBackendRegistry_RenewSavesFromEviction` now snapshots the pre-renew timestamp and sweeps at an asserted cutoff strictly between the old and renewed values. A mutant that keeps matching `Renew` successful but suppresses only the timestamp refresh fails 50/50 normal and 30/30 race runs; the candidate passes all 80 runs.
- The readiness log hook remains the deliberate public synchronization point that makes the drain-latch race deterministic. Its wait is bounded and its failure message names the missing publication hook, so no production test seam or additional comment is needed.
- Preserve mode intentionally pins its observable contract—stale external registration remains routable through a live proxied RPC—without coupling the external package to the private ticker short-circuit.

Fresh whole-candidate review is APPROVED, and the packaged fingerprint matches exactly.

@codex review

3Hren:
Resolved the final rebase and GLM follow-up in `b52c7bfd`:

- Rebases semantically onto `0fb8d986`; the allowlist conflict keeps main's authentication removals and the candidate's exact 16 gateway `private` plus four gateway `testpkg` removals.
- Adds failure-path cancellation and `group.Wait()` cleanup to the three migrated tests that could exit through intermediate assertions. The bounded `ZeroSweepInterval` test already joins as its only post-start assertion and remains unchanged.
- The 1 ns cutoff note is refuted: `time.Time.Add(time.Nanosecond)` performs exact arithmetic on one captured value, so no second clock read or platform clock-resolution dependency exists. The strict `Before` boundary remains intentional.

Fresh whole-candidate review is APPROVED; full local `make test` passes, including 68/68 Meson tests. The packaged fingerprint matches exactly.

@codex review

3Hren:
Final GLM low is verified and does not require a code change:

- The readiness tracker receives the same observed logger and logs every state transition. If `Set(READY)` moves before the blocking publication log, the observer records `UNKNOWN -> READY`; cancellation then records `READY -> NOT_READY`, so `require.Len(transitions, 1)` fails instead of silently passing.
- Removing only `readiness.WithDrainLatch()` was also mutation-tested: candidate passes 20/20 normal and 20/20 race; the no-latch mutant fails 20/20 normal and 5/5 race with the forbidden second `NOT_READY -> READY` transition.

Thus the test defends both assumptions without a production test hook or an explanatory comment that duplicates its assertions.

PR #2068, opened 2026-08-15
refactor(controlplane): encapsulate authentication internals

### LINKED ISSUE #1734: refactor(controlplane): encapsulate auth reach-ins and externalize its test packages
## Motivation

The gateway's SSH certificate/key authentication boundary — security-adjacent, though not
packet-path — has 10 `private:` rows (re-counted today, matching this item's original count
exactly): `sshcert/ca_store.go` (3 — `CAStore.VerifyCA` reaching `snapshot.entries`,
`NewCAStore`/`NewCAStoreFromLoader` reaching `s.snapshot`), `sshcert/authenticator.go`
(3 — `Authenticator.Authenticate` reaching `token.canonicalSignedData`/`.checkTimestamp`,
`NewAuthenticator` reaching `a.refreshLoop`), `sshcert/token.go` (1 — `parseToken` reaching
`token.validate`), `sshkey/authenticator.go` (2 — the same `Authenticate` pair), and
`sshkey/token.go` (1 — the same `parseToken` shape).

There are 15 `testpkg:` rows (one fewer than this item's original count of 16 — re-counted
directly against the ledger today), the largest single `testpkg:` cluster in the repository:
every `_test.go` file under the auth root and its `none`, `sshcert`, and `sshkey` subpackages
needs to move from its production package to an external `_test` package.

## Constraints

Extra care versus a routine DTO-export change: verify the exported surface doesn't widen what a
caller can do to a credential or token store, only what a test can observe.

## Acceptance

- [ ] Re-count both checks' rows for these packages in `lint/style/allowlist.txt` before
      starting (current: 10 `private:` + 15 `testpkg:`, verified 2026-08-06).
- [ ] All 10 `private:` rows fixed via exported methods/fields.
- [ ] All 15 test files moved to their external `_test` packages.
- [ ] All 25 corresponding ledger rows deleted in the same change.
- [ ] `go test ./controlplane/internal/auth/...` still green, with no widened
      credential-handling surface beyond what the tests need.

Source: migrated from the retired local task tracker's task filed 2026-07-17, on 2026-08-06.

3Hren:
- Encapsulates token operations behind read-only behavior while keeping credential-store mutation and refresh lifecycle private.
- Moves all 15 auth tests to external packages and preserves token bytes, loader routing, certificate diagnostics, CA reload, and concurrency coverage through public behavior.
- Removes exactly 25 style-ledger entries: 10 `private` and 15 `testpkg` rows.

Closes #1734.

### DIFF
--- a/controlplane/internal/auth/interceptor_test.go
+++ b/controlplane/internal/auth/interceptor_test.go
@@ -1,4 +1,4 @@
-package auth
+package auth_test
 
 import (
 	"context"
@@ -9,6 +9,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
 
+	"github.com/yanet-platform/yanet2/controlplane/internal/auth"
 	"github.com/yanet-platform/yanet2/controlplane/internal/auth/core"
 )
 
@@ -56,17 +57,17 @@ func TestUnaryServerInterceptor(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			manager, err := NewManager(&Config{
+			manager, err := auth.NewManager(&auth.Config{
 				Disabled: tt.disabled,
 			})
 			require.NoError(t, err)
 
-			interceptor := UnaryServerInterceptor(manager, log)
+			interceptor := auth.UnaryServerInterceptor(manager, log)
 
 			// Create context with or without token.
 			ctx := context.Background()
 			if tt.withToken {
-				md := metadata.Pairs(authMetadataKey, tt.token)
+				md := metadata.Pairs("x-yanet-authentication", tt.token)
 				ctx = metadata.NewIncomingContext(ctx, md)
 			}
 
@@ -96,12 +97,12 @@ func TestUnaryServerInterceptor(t *testing.T) {
 
 func TestStreamServerInterceptor(t *testing.T) {
 	log := zap.NewNop()
-	manager, err := NewManager(&Config{
+	manager, err := auth.NewManager(&auth.Config{
 		Disabled: true,
 	})
 	require.NoError(t, err)
 
-	interceptor := StreamServerInterceptor(manager, log)
+	interceptor := auth.StreamServerInterceptor(manager, log)
 
 	// Create context without token.
 	ctx := context.Background()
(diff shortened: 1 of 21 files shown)


3Hren:
Addressed the GLM findings in `3a269cab`:

- Removed the process-global HTTP transport mutation; HTTPS routing is now proven by the wrapped TLS verification error, with successful HTTP loading covered separately.
- Confirmed `Authenticator.log` is live state used by successful-authentication logging, so no production change was made.
- Removed both unused signing helpers.
- Added public-behavior coverage for CA reload publication, loader and parse rollback, and deterministic concurrent reload/verification with a blocked loader.

Targeted 20× race/shuffle and full SSH-auth 10× race runs pass, and the complete corrected candidate has fresh independent approval.

@codex review

3Hren:
Rebased the approved candidate onto current `main` and addressed the remaining GLM review points on head `022ce9900e3ebf64807c21f18585f8c266cf23e8`:

- Restored `CAStore.snapshot` as an embedded `atomic.Pointer`, preserving the `go vet` no-copy guard; constructors initialize it through the repository-convention `m` value, so no pointer-indirection workaround remains.
- Kept the refresh loop private and did not add a timing-based tick/shutdown test: the extraction is a direct dependency-explicit rewrite, while deterministic observation of both tick and shutdown would require a new timer/join seam solely for tests.

The full current-base build/test/lint/race suite and a fresh whole-candidate review are green.

@codex review

3Hren:
Reconciled every remaining GLM Low note against the exact current head; no further code change is warranted:

- `NewCAStore` used the caller's backing array before and after this refactor. A defensive copy would be a separate ownership-contract change, not preservation required by #1734.
- The refresh wiring is mechanically equivalent. A deterministic external test for both tick and shutdown would require a timer/join production seam solely for testing; a short-interval test would remain timing-dependent and could not prove termination.
- The literal metadata key is already pinned by the positive `TestExtractToken` cases. Exporting it would make tests echo production and widen the API contrary to this issue's constraint.
- The empty-principal case does exercise `extractPrincipal` through `CAStore.VerifyCA`; the remaining valid-flow duplication is harmless legacy redundancy, while all rejection classes stay distinct.
- gRPC `status.Errorf` formats inner errors rather than preserving their identity. Replacing the remaining class-specific literals with `Err….Error()` would be cosmetic string composition; tests already use sentinels where that is meaningful.

A fresh independent reconciliation kept the exact 21-file candidate `APPROVED`; its fingerprint and head are unchanged.

PR #2070, opened 2026-08-15
refactor(controlplane): restore auth refresh ownership

3Hren:
- Restores refresh-loop ownership to the authenticator without adding lifecycle seams.
- Removes duplicate success-path tests while preserving algorithm and failure coverage.
- Keeps the externally testable token and CA-store contracts from the original change.

### DIFF
--- a/controlplane/internal/auth/sshcert/authenticator.go
+++ b/controlplane/internal/auth/sshcert/authenticator.go
@@ -30,6 +30,7 @@ type Authenticator struct {
 	caVerifier        CAVerifier
 	revocationChecker RevocationChecker
 	timeWindow        time.Duration
+	refreshInterval   time.Duration
 	stopCh            chan struct{}
 	log               *zap.Logger
 }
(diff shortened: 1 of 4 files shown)


3Hren:
@codex review

3Hren:
GLM findings reconciled against the complete test class:

- The three `sshkey` success paths remain covered by `TestAuthenticate_Ed25519`, `TestAuthenticate_RSA`, and `TestAuthenticate_ECDSA` in `authenticator_test.go`. Those tests drive the public authenticator through the same per-algorithm signature-verification dispatch, so the removed stage-local tests were duplicate coverage.
- The valid-certificate path remains covered by `TestAuthenticator_HappyPath` in `sshcert/authenticator_test.go`. It authenticates a valid ECDSA user certificate and verifies the returned username and auth method, exercising parsing, the accepted key and certificate types, the validity window, principal extraction, CA validation, and canonical signature verification together.
- The non-positive refresh interval behavior predates this patch and is deliberately unchanged. Adding clamping here would alter the constructor contract and broaden this scoped ownership rollback.

The independent whole-candidate review audited these mappings and found no coverage loss, so no code change is needed.

Issue #2071, opened 2026-08-15
test(nat64): remove the unit test's sanitizer opt-out and fix what it was masking


3Hren:
## Motivation

Meson injects fatal sanitizer options into every test it runs — halt on error, abort on error — unless the test overrides them. Sixty-seven of the repository's sixty-eight tests run that way and are genuinely clean, measured at commit `0fb8d986`. The exception is [modules/nat64/tests/unittest/meson.build#L28-L33](https://github.com/yanet-platform/yanet2/blob/76627efc891501de314597201e0040542290f748/modules/nat64/tests/unittest/meson.build#L28-L33), which sets `UBSAN_OPTIONS=halt_on_error=0:print_stacktrace=0` and `LSAN_OPTIONS=detect_leaks=0`. It is the only opt-out in the repository, and it dates from the change that originally made sanitizer runs pass.

Behind that opt-out sit twenty undefined-behaviour reports and a leak, and **all of them are in the test file itself** ([modules/nat64/tests/unittest/nat64_test.c](https://github.com/yanet-platform/yanet2/blob/76627efc891501de314597201e0040542290f748/modules/nat64/tests/unittest/nat64_test.c)) — nothing in the module's control-plane or dataplane sources is implicated. Composition, measured:

- Eight reports of a left shift overflowing a signed int, from the DPDK address-composing macro whose first octet exceeds 127. The test file has eleven uses of that macro and every one of them has a first octet above 127, so all eleven are capable of the same undefined behaviour — only two lines report today, because the other nine are file-scope initializers that the compiler constant-folds, emitting no runtime check. Fixing only what reports would leave nine latent.
- Twelve reports of misaligned member access, where the test casts a pointer to an ICMP or ICMPv6 header onto an odd offset inside a packet buffer and reads or writes through it. Roughly three helper regions.
- With leak detection re-enabled the run reports about 29.5 kilobytes leaked across 242 allocations in 42 blocks, and every one of the 42 roots in the same test-case registration helper — the linked list of test cases is built and never freed. So the leak opt-out masks a missing teardown in the harness, not a product leak.

## Constraints

Removing the opt-out naively does not give a list of things to fix. With the fatal options restored the binary aborts on the very first report, during setup, before the suite runner starts — zero of twenty-six tests execute. The work has to be driven with the non-fatal setting until the count reaches zero, and only then should the opt-out be removed, with the removal itself acting as the assertion.

## Acceptance

[...]

PR #2073, opened 2026-08-15
ci(sanitizer): fail the Go half of the sanitizer job on a UBSan diagnostic

### LINKED ISSUE #2056: ci(sanitizer): fail the sanitizer job on a UBSan/ASan diagnostic instead of discarding it

3Hren:
A sanitizer diagnostic did not reliably fail the sanitizer job, but the gap was narrower than it looked and worth stating precisely, because it decided the shape of the fix.

Meson injects fatal sanitizer options into every test it runs unless that test overrides them, so the C half of this job already failed on a diagnostic — measured at 67 of 68 tests, the exception being one deliberate opt-out. The Go runner set none of them. An address-sanitizer error there still aborted and failed the package, so that half was never silent. What was silent is a recoverable undefined-behaviour diagnostic: it left the package reporting success, and without verbose output its text was discarded rather than merely unread. That is where fifty-three zero-length-array reports sat on a green main until #2055.

The Go invocation now carries the same option string meson injects, verified byte-identical against meson's own source rather than transcribed, so both halves of one job run one policy instead of two.

## Why fatal rather than scanning output

The issue proposed capturing every package's output and scanning it for sanitizer markers. Making the diagnostic fatal is preferred for three reasons. A failing package has its output printed by the test runner already, so visibility arrives with severity rather than needing its own mechanism. There is no marker list to keep in sync with the sanitizers' message formats. And it is what the C half already does, so the two halves stop diverging.

The one thing a scan could have covered that fatality cannot is a class that prints without killing. Enumerated, there is no such class here: recoverable undefined behaviour is the case being fixed, address errors already fail, and leaks never report at all under the Go runner, so there would have been nothing to scan for them either.

## Evidence

The Go half is clean at the base commit: zero diagnostics across the sweep, 71 packages passing, and enabling the gate changes no verdict. Instrumentation was proven present rather than inferred, so the silence is real silence.

For the second acceptance criterion, a reintroduction was verified rather than assumed. A pre-fix object from the #2055 class was rebuilt and placed into the real link path: ungated it printed six reports while its package still passed, and gated the same package failed.

## What this deliberately does not cover

Leaks remain a C-half capability only. The Go runtime exits through a raw system call, so the leak check registered at exit never runs — no flag changes that, and neither approach to this issue would have. Go code itself is never instrumented; only C reached through cgo is, which is roughly two thirds of the tested packages.

The nat64 unit test keeps its opt-out, so the C half retains exactly one hole. Removing it means fixing twenty diagnostics and a leak in that test harness, which is separable work tracked in #2071 rather than folded in here.

## Cost to expect

Any future undefined behaviour in C reachable from a Go test now fails CI, including third-party code after a submodule bump. That exposure is largely already present, since the C half applies this same policy to the same objects. When it does bite, the remedy is a targeted suppression entry or disabling the specific check for the offending translation unit — not removing the gate.

Closes #2056.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,10 @@ test-asan-only:
 		meson configure -Dbuildtype=debug -Doptimization=0 -Dfuzzing=disabled -Db_sanitize=address,undefined $(EXTRA_MODULES_FLAG) build; \
 	fi
 	meson compile -C build
-	CGO_CFLAGS="-fsanitize=address,undefined" CGO_LDFLAGS="-fsanitize=address,undefined" go test -count=1 $$(go list ./... | grep -v '^github.com/yanet-platform/yanet2/tests/functional')
+# Set as a default only, mirroring the same if-absent condition meson uses for
+# its own injection: without it, a recoverable diagnostic would go unseen and
+# the package would still pass.
+	CGO_CFLAGS="-fsanitize=address,undefined" CGO_LDFLAGS="-fsanitize=address,undefined" UBSAN_OPTIONS="$${UBSAN_OPTIONS:-halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1}" go test -count=1 $$(go list ./... | grep -v '^github.com/yanet-platform/yanet2/tests/functional')
 	meson test -C build
 
 test-tsan:


3Hren:
Both low findings taken; the medium one is already answered in the description, but it deserves the specifics rather than a pointer.

**Setting it unconditionally diverged from meson's inject-if-absent policy** — fixed, and it was the better catch of the two. The point is sharper than parity: with a caller-supplied value the old form let `meson test` honour it while the Go half in the same recipe clobbered it, which is the two-policy split this change exists to remove, reappearing exactly when someone is deliberately steering the sanitizer. It also made an inventory run impossible without editing the file, since collecting every diagnostic rather than dying on the first one requires halting turned off — which is how the measurements behind this change were taken. The assignment is now a shell default. Note the form is the substitute-on-unset-or-empty one, because meson's own condition treats an empty value as absent too; the substitute-on-unset-only form would have let an exported-but-empty variable silence the Go half while meson still injected for the C half.

**The comment described the pre-change behaviour in the present tense** — fixed, mechanism in the present and consequence in the conditional.

**On the positive control** — it exists, and predates the pull request rather than being owed by it. The concern that a clean sweep is equally consistent with an inert gate is the right one, so it was falsified through the real recipe rather than a synthetic package: an object from the defect class this follows was rebuilt from its pre-fix source using its own recorded compile arguments, placed into the actual archive the cgo link path consumes, and the module's own test package run both ways. Ungated it printed six diagnostics and the package still reported success. Gated, the same package failed. The object was then restored and verified byte-identical before the clean sweep was taken, so the two results come from the same tree. A synthetic cgo package was also checked separately and behaved the same way. That covers all three of the places the Go half differs from the proven C half — the runner buffering and discarding a passing package's output, the abort traversing the Go runtime, and the exit status reaching make.

Worth adding for the record, since it bears on how much the gate covers: leaks can never fire under the Go runner at all, because the runtime exits through a raw system call and the check registered at exit never runs. So there was no third class for a scan to have caught either, and leak coverage stays a property of the meson half.

@codex review

PR #2081, opened 2026-08-16
test(route): drop a dead module-config cleanup

3Hren:
A route functional test registers a cleanup that deletes its module config, then wires the module into a pipeline. The wiring helper registers no cleanup of its own, so the module is still referenced by a live chain when the delete runs, and the backend correctly refuses to free it rather than leave the chain dangling. The error was discarded, so the call freed nothing and reported nothing.

- Remove the cleanup. Verified by running the same delete against a wired and an unwired module: it succeeds on the unwired one and fails on the wired one, which is the only case this test ever reached, so removing it changes no observable state.
- The per-test harness teardown already releases the config and the whole shared-memory segment, and nothing after the deleted line depends on the config having been removed.

This was the last instance of the shape in the package; the two others were removed in #2077, where the mechanism was first established.

### DIFF
--- a/modules/route/tests/functional/route_test.go
+++ b/modules/route/tests/functional/route_test.go
@@ -1157,7 +1157,6 @@ func TestUpdateFIB_ShadowedNexthopCounterExcludedFromMetrics(t *testing.T) {
 		}),
 	})
 	require.NoError(t, err)
-	t.Cleanup(func() { _, _ = service.DeleteConfig(t.Context(), &routepb.DeleteConfigRequest{Name: "test"}) })
 
 	wirePipeline(t, agent, "port0", "test")
 


3Hren:
The narrative is right as written, so no correction here. There is no client-side cancellation check to fire first: the route service never reads its context parameter — it appears only in the four method signatures and in no body — and these are direct in-process method calls with no gRPC transport, so nothing consults the deadline. That was verified by running the delete with an already-cancelled context, which returned a nil error and did real work.

The operative failure really is the backend's refusal while the module is still referenced by a live chain, which is what the wired-versus-unwired comparison isolates. The same point came up on the parent change and was refuted the same way.

PR #2086, opened 2026-08-16
fix(fwstate): lock each probed fwtable layer during chain walks

GeorgyKirichenko:
The deep layers of a lookup walk were probed without any bucket lock, so a reader could scan a bucket while another thread was still working with that layer. Each probed layer now takes its own bucket read-lock for the scan and drops it before advancing, keeping the first-layer lock contract unchanged.


GeorgyKirichenko:
good catch from glm, this will be addressed in further work

Issue #1918, opened 2026-08-10
fix(controlplane): modules absent from the config default to instance 0


3Hren:
## Motivation

Modules not listed in a control-plane YAML are still constructed from their `DefaultConfig()`, whose zero-value `InstanceID` is 0 — see for example [`InstanceID` in `modules/decap/controlplane/cfg.go`](https://github.com/yanet-platform/yanet2/blob/15a5459b2ba1933f01f15f1f689775b1676c391c/modules/decap/controlplane/cfg.go#L9-L10), a shape shared by every module's `cfg.go`.

Measured in production: a control plane configured entirely for instance 1 owned 8 agents on instance 0 — route-mpls, decap, dscp, mirror, nat64, pdump, blackhole, and trafgen — none of which appeared in the operator's own configuration. Any call through those agents contends on a different instance's `cp_config_lock` than the one the operator believes is in use, which in the incident described in #1913 was one already starved to the point of being effectively poisoned.

## Scope

A module missing from the configuration should either be genuinely absent (no agent attached at all) or fail configuration loading with an explicit error, never silently default to instance 0.

## Acceptance

- [ ] Loading a control-plane configuration that omits a module either skips attaching that module's agent, or rejects the configuration with a clear error naming the module.
- [ ] A regression test covers a configuration that lists only some modules and asserts the omitted ones do not silently attach to instance 0.
- [ ] Existing single-instance deployments, where every module implicitly targets instance 0, keep working unchanged.

Source: production incident investigation, 2026-08-10. See also #1913 (incident epic).

3Hren:
Implementation notes, sharpening the acceptance criteria above (both branches of the "either/or" are being done, not one):

- A module absent from the YAML is genuinely absent — no config, no service constructed, no agent attached. This is stricter than "either skip or error": it always skips, never silently constructs from `DefaultConfig()`.
- A module that *is* listed in the YAML must state `instance_id` explicitly. Instance 0 stays legal, it just has to be written down — enforced via a new `xcfg.Required[uint32]` type. `gateway.instance_id` becomes required the same way.
- The shipped debian conffile `controlplane/etc/yanet/controlplane.d/default.yaml` is being expanded to enumerate every module it previously ran implicitly, plus the `plain` and `vlan` devices, so a package upgrade preserves today's behaviour on single-instance deployments (satisfying the third acceptance box). Deliberate exception: the `trafgen` device will no longer be started by default.

[...]

PR #2104, opened 2026-08-17
fix(xcfg): reject a config key present with a null value

### LINKED ISSUE #1947: fix(controlplane): a module key present with a null value is silently treated as absent
## Motivation

#1918 fixes the case where a module key is missing entirely from `modules:` in the control-plane YAML, but leaves an adjacent gap: a key that is *present* with an explicitly null value — `modules:\n  route:\n` — also decodes to a nil `*route.Config` via yaml.v3, and the bundle treats it identically to the key being absent: an Info-level "skipping service with no config" log, and the module is never started. An operator who comments out a module's configuration body but leaves its key in place hits exactly the silent misconfiguration #1918 exists to end. The same collapse applies one level up: an explicitly-null top-level `modules:` (or `devices:`) block starts a director with zero modules, with no error at all.

Distinguishing "key present, value null" from "key absent" needs presence detection on the underlying YAML node in `bundle.ModulesConfig`/`DevicesConfig` (for example a `yaml.Node` field or a custom `UnmarshalYAML`), which #1918 deliberately avoided — yaml.v3's nil-pointer-on-decode behavior already gives the absent-key case for free without it.

The code in question (`controlplane/bundle/bundle.go`, the "skipping service with no config" log path) is not yet in a pushed commit: it is part of #1918's in-progress branch `fix/controlplane-explicit-instance-id`, not yet opened as a PR.

## Scope

- Reject an explicitly-null module or device block with an error naming the module/device, since the key's presence signals intent to configure it.
- Reject an explicitly-null top-level `modules:`/`devices:` block the same way.

## Acceptance

- [ ] Loading a configuration with a present-but-null module or device key fails with an error naming that module/device, distinguishable from the key being absent (which #1918 already handles correctly).
- [ ] A regression test covers `modules:\n  <name>:\n` (null body) and asserts it is rejected rather than silently skipped.
- [ ] A null top-level `modules:`/`devices:` block is rejected the same way.

Source: reviewer finding during review of #1918 (branch `fix/controlplane-explicit-instance-id`, not yet opened as a PR), 2026-08-11. See also #1913 (incident epic).

3Hren:
A module key written with no value — `route:` with its body commented out — decoded identically to the key being absent from the document, so the module was never started and the only trace was an informational log line. The same collapse applied to a null top-level `modules:` or `devices:` block, which started a control plane with no modules at all and no error. This is the adjacent half of the gap #1918 closed for a genuinely absent key.

- The rejection runs during the walk over the parsed document, not inside the configuration wrappers, because the YAML library returns before consulting a type's own decoding when the node is null — so no wrapper ever gets the chance to notice, and the distinction is unrecoverable at that level. It applies to every caller of the decoder, not only to the ones that asked for unknown-key checking.
- The rule is expressed as the complement of the walk's existing stop conditions rather than as a list of known types, so a wrapper added later is classified correctly without being enumerated here. A wrapper is classified by the type it decodes into, which leaves an optional scalar exactly as silent as a plain one.
- Optional blocks, plain nested blocks and maps are rejected, naming the dotted path and the line, and telling the operator to give the key a body or remove it. An absent key stays silent, and an explicit empty mapping still reaches the block's own validation, so neither existing spelling changes meaning.
- Wrappers that carry their own decoding keep reporting through their own messages, so a required field left empty still says it must be set explicitly rather than being reported here.
- Pointer-typed blocks are deliberately excluded, which keeps the gateway block reporting through its own validation and leaves that message and its test untouched.
- Rejecting a null plain nested block is a deliberate scope choice rather than a consequence: at that position a null preserves defaults and so erases nothing in general, but the top-level modules block is exactly such a position and every field under it is optional, so there it does erase.
- The walk follows the YAML library's own merge semantics rather than approximating them, because a divergence either rejects a document that decodes or stays silent about one that erases a block. A merge key is a merge wherever it appears, including into a map, and its content is checked against the receiving mapping rather than against a field named after the merge key. A mapping's own keys are matched before anything it merges, at every nesting depth, so a key the receiving mapping or an earlier source already binds is never reported out of a source the library discards. A mapping that repeats the merge key cannot decode at all, so which of the two the walk would have applied is unobservable and left untested.
- A key written as an alias is named by the value it resolves to rather than by its anchor, which is what the library decodes it as. The two existing alias tests anchored a name identical to the value it resolved to and so could not tell the two apart; renaming the anchor gives them that power, and the fixture note that recorded the old behaviour as a constraint is gone.
- A null merge value is left to the YAML library's own merge error rather than being reported as a missing key, and a null reached through an alias is reported at the alias site rather than at the anchor.

Closes #1947.

### DIFF
--- a/common/go/xcfg/known_keys.go
+++ b/common/go/xcfg/known_keys.go
@@ -1,6 +1,7 @@
 package xcfg
 
 import (
+	"encoding"
 	"fmt"
 	"maps"
 	"reflect"
@@ -14,6 +15,8 @@ var yamlNodeType = reflect.TypeFor[yaml.Node]()
 
 var yamlUnmarshalerType = reflect.TypeFor[yaml.Unmarshaler]()
 
+var textUnmarshalerType = reflect.TypeFor[encoding.TextUnmarshaler]()
+
 // walkableType is implemented by a wrapper whose keys decode into a
 // different type than the wrapper itself, such as Optional[T], so the walk
 // can check that type's fields instead of stopping at isOpaqueToWalk.
@@ -24,11 +27,23 @@ type walkableType interface {
 // mergeTag is the resolved tag yaml.v3 assigns to a "<<" merge key.
 const mergeTag = "!!merge"
 
+const nullTag = "!!null"
+
 // unknownKey records one mapping key with no matching field, along with the
 // document line it sits on, so the reported error can point straight at it.
 type unknownKey struct {
 	Path string
 	Line int
+}
+
+type nullValue struct {
+	Path string
+	Line int
+}
+
+type findings struct {
+	Unknown []unknownKey
+	Nulls   []nullValue
 }
 
 // CheckKnownKeys reports every YAML mapping key in data that has no matching
@@ -44,20 +59,31 @@ func CheckKnownKeys[T any](data []byte) error {
 	return checkKnownKeys(data, reflect.TypeFor[T]())
 }
 
-// checkKnownKeys drives the same walk as CheckKnownKeys against a
-// reflect.Type, letting Decode reuse it for a destination value's runtime
-// type without a type parameter.
+// checkKnownKeys drives the same walk as CheckKnownKeys against the
+// reflect.Type its type parameter resolves to.
 func checkKnownKeys(data []byte, t reflect.Type) error {
-	var doc yaml.Node
-	if err := yaml.Unmarshal(data, &doc); err != nil {
+	collected, err := walkDocument(data, t)
+	if err != nil {
 		return err
 	}
+	return unknownKeysError(collected.Unknown)
+}
+
+func walkDocument(data []byte, t reflect.Type) (*findings, error) {
+	var doc yaml.Node
+	if err := yaml.Unmarshal(data, &doc); err != nil {
+		return nil, err
+	}
+	collected := &findings{}
 	if len(doc.Content) == 0 {
-		return nil
+		return collected, nil
 	}
 
-	var unknown []unknownKey
-	walkKnownKeys(doc.Content[0], t, "", &unknown, map[*yaml.Node]bool{})
+	walkKnownKeys(doc.Content[0], t, "", collected, map[*yaml.Node]bool{})
+	return collected, nil
+}
+
+func unknownKeysError(unknown []unknownKey) error {
 	if len(unknown) == 0 {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


3Hren:
Force-pushed a test-only restructure on top of the reviewed implementation: the walker's tests now live in the file named after the code they test, the "null at position X" family is one table with its config type declared beside it, and a duplicate of an existing absent-key test was dropped.

@codex review

3Hren:
Pushed one more test fix: the table runner's state-assertion hook now runs on both the error and the no-error path, so a future row that expects an error and also asserts state cannot pass without checking it.

Note for anyone reading the red checks here: all four failures on this pull request are already failing on `main` itself, introduced upstream of this branch — Go Lint, Build YANET and Build DEB Packages are red on `875fd6dc` and on `27bfeb51` before it, with an identical `objects/fwstate/controlplane` typecheck error. Nothing in this branch touches that package. This will not be merged until `main` is green again.

@codex review

3Hren:
Force-pushed a rewrite of how the walk models YAML merges, plus the two smaller points from the last pass.

- A merge key is now a merge in a map-typed destination as well as a struct-typed one, so merged content is checked against the receiving mapping instead of against a field named after the merge key, and a null merge value keeps surfacing the YAML library's own merge error there too.
- A mapping's own keys are matched before anything it merges, at every nesting depth, matching the order the library actually decodes in. Without this a merged null shadowed by an explicit key was rejected although the document decodes, and a source's own null shadowing a merged body was missed although it erases the block.
- A wrapper is classified by the type it decodes into, so a null optional scalar stays as silent as a null plain scalar.
- The decoder's doc now states that the rejection applies whether or not unknown-field checking was requested.

The four red checks on the previous head were the `objects/fwstate/controlplane` typecheck failure that was already on `main`; #2108 fixed it and this branch is rebased on top.

@codex review

3Hren:
Addressed the last pass. Two of the four findings were real and are fixed; one is declined with evidence, one is filed separately.

- Aliased mapping keys are now named by the value they resolve to. An alias node carries its anchor name in `Value`, so a key written as an alias was matched under the anchor — mis-naming the path under a map destination and, under a struct destination with unknown-key checking, rejecting a document that decodes cleanly. The two existing alias tests could not catch it because each anchored a name identical to the value it resolved to; their anchors are renamed, and the fixture note that recorded the old behaviour as a constraint is gone.
- A merged entry under a non-string map key is walked like a direct one, instead of being dropped. Such a null decoded silently to an absent block, which is the collapse this pull request exists to reject.
- The empty-name collapse for two complex keys under one map destination is declined: the library's duplicate scan compares node kind and value, every complex key carries an empty value, and the scan runs above the map and struct dispatch — so two of them make the document fatal for every destination kind, including two *different* sequence keys. No decodable document reaches the skip.
- Unbounded alias expansion in the pre-walk is filed as #2110. The class predates this branch: the walk already ran for the director and the bird-adapter, which both ask for unknown-key checking. What changed here is its reach, so the fix belongs on its own.

Rebased onto current main.

@codex review

3Hren:
One more real finding addressed: within a single mapping the entry that decides is the last one with a given resolved name, which is what the library decodes, not the first. The two rules only diverge when a duplicate escapes the library's duplicate scan — a plain key and an aliased key resolving to the same name differ in both node kind and value — and there the walk was rejecting a document that decodes to a valid body.

The companion suggestion, excluding a map entry from null checking when its key cannot decode into the map's key type, is declined. Such a document is fatal in yaml.v3 either way, so only the wording of the error changes, and keying the rule on decodability would break the case where a non-string key type accepts a complex key and this walk is the only thing that catches an erased block there.

@codex review

3Hren:
Third pass addressed; one of the three was a genuine save.

- A struct that decodes from text is no longer treated as a block. Such a type decodes from a scalar, and the library leaves a struct field untouched on a null rather than zeroing it, so a null there preserves the seeded value exactly like a null plain scalar. Without this the branch would have rejected a bird-adapter configuration that works today, whose two address fields are of that shape.
- Merge-key detection now matches the library on kind and value, not on the tag alone. The reported spelling turned out to be unreachable — the parser normalizes a non-specific tag to the merge tag before the walk sees it, and the test written for it was deleted rather than kept green for nothing. The predicate earns its place for the opposite case: an ordinary key carrying an explicit merge tag decodes as an ordinary key, and the tag-only test used to swallow a null under it, which is the erasure this pull request exists to catch.
- Deduplicating map entries by decoded key identity rather than key text is declined: the two differ only for a non-string key type, and every configured map in this repository is string-keyed, where the text is the decoded key.

One seam is known and left open, since nothing reachable hits it: a quoted merge key written inside a merge source is dropped by the library but reported by the walk. Closing it is a one-line mirror of how the library seeds its shadowing set, if a real configuration ever produces it.

@codex review

PR #2111, opened 2026-08-18
fix(operator): reject a null server block instead of crashing at startup

### LINKED ISSUE #2100: fix(operator): reject a null `server:` block instead of crashing at startup
## Motivation

An operator config with an explicitly null `server:` block (`server:` with no body) crashes the operator process at startup with a nil pointer dereference, instead of failing with a configuration error.

`xcfg` decoding leaves a pointer field nil when the YAML gives its key an explicitly null value. [common/go/operator/operator.go#L76](https://github.com/yanet-platform/yanet2/blob/e83254986f45593a13a2ed2a7fc05b6518487f93/common/go/operator/operator.go#L76) reads `opts.GRPCServer.Config.Endpoint.Unwrap()` unconditionally once `opts.GRPCServer != nil`, but nothing validates that `GRPCServer.Config` (in turn `GRPCServer.Config.Server`) is non-nil before that point. No `Validate()` on the operator config types checks `Server` for nil — see [operators/decap/internal/operator/cfg.go#L37-L60](https://github.com/yanet-platform/yanet2/blob/025b68657ced0c4a9a657f29c58fb551354322dd/operators/decap/internal/operator/cfg.go#L37-L60) and its siblings in `operators/{pipeline,route,forward}`.

Reproduced by a reviewer with a probe, not inferred.

## Scope

Applies to every operator built on `common/go/operator` (`decap`, `pipeline`, `route`, `forward`), all of which share the same config-validation gap.

## Constraints

The in-flight fix for #1947 deliberately excludes pointer positions from its new null-block rejection, to leave `yncp.Config.Validate()`'s own gateway message intact. This gap is therefore knowingly left open by that change and wants its own decision here — either a nil check in each operator config's `Validate()`, or extending #1947's null-block rejection to pointer positions.

## Acceptance

- [ ] A null `server:` block (or equivalent nil `GRPCServer.Config`) is rejected as a configuration error at startup, on all four operators, not a panic.
- [ ] Decide and document whether the fix lives in per-operator `Validate()` or in `xcfg`'s null-block rejection.

Source: surfaced by review of the #1947 fix, 2026-08-17 (reviewer probe, reproduced on `main`, not introduced by that change).

3Hren:
An operator configuration that cleared the `server:` block, written either as a bare `server:` key or as an explicit `server: null`, decoded into a nil pointer and wiped the default endpoint. The operator then died at startup on a nil dereference while reading that endpoint, rather than reporting a configuration error. All four operators built on `common/go/operator` shared the gap.

- Make the `server:` block a plain value instead of a pointer in `decap`, `pipeline`, `route` and `forward`. Nil stops being representable, so the dereference in `NewOperator` is unreachable, and a cleared block is caught by the null-block check the loader already runs — reported with the key and the line it sits on, no new validation code.
- Take the gRPC server configuration by value through `WithGRPCServer` and `NewGRPCServer` for the same reason: an operator that runs without a server says so by omitting the option, never by passing an empty one. `*GRPCServerConfig` no longer appears anywhere, so the guarantee lives in the type rather than in a rule callers must remember.
- Drop the write-only `cfg` field from `GRPCServer`, dead since it was added.

No tests are added: the rejection is `xcfg`'s, covered by its own tests, and the field's shape is now enforced by the compiler.

Rejected alternatives, both measured rather than argued: a nil check in each operator's `Validate()` leaves the dereference reachable for the next operator that forgets it, and `xcfg.Optional` seeds `new(T)` plus `Default()` on decode, so for a type without a `Default()` a partial `server:` block would silently discard the operator's defaults.

`WithGRPCServer` changing shape is a breaking change for any out-of-tree caller, which the `neighbours` operator in the private repo may be. It breaks at compile time, which is the right failure mode, but an adapter that passes `*cfg.Server` while keeping a pointer field would move the same nil dereference to its own call site.

`controlplane/yncp` keeps the pointer-plus-nil-check shape for its `gateway:` block; aligning it is filed as #2113.

Closes #2100.

### DIFF
--- a/common/go/operator/options.go
+++ b/common/go/operator/options.go
@@ -10,7 +10,7 @@ import (
 )
 
 type grpcServerOption struct {
-	Config   *GRPCServerConfig
+	Config   GRPCServerConfig
 	Services []ServiceRegistrar
 }
 
@@ -86,7 +86,7 @@ func WithGateways(register RegisterConfig, gateways ...GatewayConfig) Option {
 // metrics or operator APIs without director registration.
 //
 // When this option is not supplied, the operator does not bind a listener.
-func WithGRPCServer(cfg *GRPCServerConfig, services ...ServiceRegistrar) Option {
+func WithGRPCServer(cfg GRPCServerConfig, services ...ServiceRegistrar) Option {
 	return func(o *options) {
 		o.GRPCServer = &grpcServerOption{
 			Config:   cfg,
--- a/common/go/operator/server.go
+++ b/common/go/operator/server.go
@@ -32,7 +32,6 @@ func WithGRPCLog(log *zap.Logger) GRPCServerOption {
 
 // GRPCServer wraps a grpc.Server with the operator's service set.
 type GRPCServer struct {
-	cfg    *GRPCServerConfig
 	server *grpc.Server
 	log    *zap.Logger
 }
@@ -41,7 +40,7 @@ type GRPCServer struct {
 // registrars applied to a fresh grpc.Server and returns registered
 // service names.
 func NewGRPCServer(
-	cfg *GRPCServerConfig,
+	cfg GRPCServerConfig,
 	services []ServiceRegistrar,
 	options ...GRPCServerOption,
 ) (*GRPCServer, []string) {
@@ -57,7 +56,6 @@ func NewGRPCServer(
 	}
 
 	return &GRPCServer{
-		cfg:    cfg,
 		server: server,
 		log:    opts.Log,
 	}, serviceNames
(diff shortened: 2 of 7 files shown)


3Hren:
@codex review

3Hren:
Force-pushed a rewrite. The first attempt added a nil check for a pointer that should never have been one: it left the dereference in `NewOperator` reachable and grew an exported `ValidateNotNil` to guard a state the type could simply not have. The block is a value now, nil is unrepresentable, and the cleared-block error comes from the loader's existing null-block check.

@codex review

PR #2120, opened 2026-08-18
fix(controlplane): guard worker counter lookups against a missing counter

### LINKED ISSUE #1935: fix(controlplane): guard worker counter lookups against a missing counter
## Motivation

`controlplane/ffi/worker.go:49-58` resolves nine worker counter handles by name out of `counterByName`, and every subsequent use dereferences the result without a nil check — `dropsHandle.values` and its eight siblings at roughly [:70-120](https://github.com/yanet-platform/yanet2/blob/9f753a0040359723a599dcbb3d17edb9533c8534/controlplane/ffi/worker.go#L70-L120), plus `rxBurstsHandle.size` in the burst loop. A Go map miss yields a nil `*C.struct_counter_handle`, so a counter the dataplane did not register crashes the control plane with a nil dereference instead of surfacing a clean error.

Today the names always match, so this is latent — but it is exactly the failure mode a rename of the C-side `drops` counter would arm (tracked in #1934), and version skew between a restarted control plane and a running dataplane is the realistic trigger even without that rename: [controlplane/ffi/worker.go:41-58](https://github.com/yanet-platform/yanet2/blob/9f753a0040359723a599dcbb3d17edb9533c8534/controlplane/ffi/worker.go#L41-L58) builds the map by name from whatever the running dataplane happens to have registered.

## Scope

Nil-check each resolved handle and fail with a descriptive error naming the missing counter, in the style of the existing metadata failure right below it (`controlplane/ffi/worker.go:64-69`). This should land before any change to the C-side counter name.

## Acceptance

- [ ] Every handle resolved via `counterByName` in `WorkerCounters()` is nil-checked before use.
- [ ] A missing counter returns a descriptive error instead of crashing.

Source: found while validating the worker-counter rename (#1907, #1927, #1929) for #1934 — `controlplane/ffi/worker.go` resolves counter handles by name with no nil guard, which the rename would arm.

egnees:
Worker counter handles are resolved through a name-indexed counter set that records every counter it could not resolve, so a counter the dataplane did not register — or registered with an unexpected number of values — surfaces as a descriptive error naming it instead of crashing the control plane on a nil dereference. Version skew between a restarted control plane and a running dataplane is the realistic trigger, since the set is built from whatever the running dataplane happens to have registered.

A miss materializes as a zeroed counter of the requested shape and is remembered on the set, so the resolution stays a straight-line list of lookups checked once at the end, and one read reports every missing counter rather than only the first.

Closes #1935.

- Each lookup declares the number of values it expects, so a counter that resolves by name but carries a different shape is a miss rather than a silently wrong reading.
- The worker burst histogram is taken from the counter's own values instead of being sized from the burst-size metadata, which removes the assumption that the two agree.
- The counter surface moves out of the shared-memory file into its own file, and the worker read goes through the same decoding path as every other counter read.
- Unit tests cover the duplicate-name and instance-count rejections, and the miss bookkeeping.
- Fixed style violation on moving counters surface to the separated file, updated allowlist.txt.


3Hren (dismissed):

egnees:
@codex review

3Hren (approved):

PR #2122, opened 2026-08-18
fix(counters): handle a missing worker execution context in a counter read

### LINKED ISSUE #1926: fix(controlplane): yanet_get_counters_by_tags NULL-derefs a worker with no execution context
## Motivation

`yanet_get_counters_by_tags` collects per-worker counter-storage matches and explicitly tolerates a worker with no execution context yet:

https://github.com/yanet-platform/yanet2/blob/6057bbcaa8c95e2fbaa121f69ccd3dd7176765ea/lib/controlplane/agent/agent.c#L1763-L1769

```c
struct config_gen_ectx *ectx =
	cp_config_gen_worker_ectx(cp_config_gen, worker_idx);
if (ectx == NULL) {
	worker_matches[worker_idx] = NULL;
	continue;
}
```

But the later per-counter fill loop indexes every worker unconditionally:

https://github.com/yanet-platform/yanet2/blob/6057bbcaa8c95e2fbaa121f69ccd3dd7176765ea/lib/controlplane/agent/agent.c#L1840-L1845

```c
for (uint64_t worker_idx = 0; worker_idx < worker_count; ++worker_idx) {
	worker_storages[worker_idx] = ADDR_OF(
		&worker_matches[worker_idx][i]->storage
	);
}
```

This loop only runs when `matches0 != NULL`, i.e. worker 0 has an ectx. If any *other* worker `w` does not, `worker_matches[w]` is the NULL set at collection time above, and `worker_matches[w][i]` is a NULL-pointer dereference — crashing whichever process called this (a metrics scrape or `yanet-cli` counters read).

`cp_config_gen_worker_ectx` is documented to return NULL "when ... no execution context has been installed yet":

https://github.com/yanet-platform/yanet2/blob/6057bbcaa8c95e2fbaa121f69ccd3dd7176765ea/lib/controlplane/config/zone.h#L257-L271

so a mixed installed/not-installed state across workers is exactly what this accessor's contract represents, not a hypothetical. Whether that mixed state is reachable against a *published* config generation (as opposed to only during in-progress construction under the same `cp_config_lock` this function also takes) is not yet confirmed dynamically.

## Scope

- Guard the fill loop against a NULL `worker_matches[worker_idx]`, consistent with the collection loop's own tolerance of a worker with no ectx yet.
- Decide the right response semantics for a not-yet-installed worker (skip it, or zero-fill `dst->values[worker_idx]`) instead of crashing.

## Acceptance

- [ ] Bug-hunter confirmation that a mixed ectx/no-ectx worker state is reachable against a live, in-use config generation (e.g. mid-install or after partial worker bring-up), not just a theoretical accessor precondition.
- [ ] The fill loop in `yanet_get_counters_by_tags` no longer dereferences `worker_matches[worker_idx]` when it is NULL.
- [ ] Counter-read callers get a well-formed response (not a crash) when one or more workers have no ectx installed yet.

Source: surfaced during code review of #1914 (config-lock scrape livelock fix). Pre-existing at HEAD — #1914 preserves the identical indexing and neither introduces nor worsens it. Not observed in production; unconfirmed by dynamic analysis. Related: epic #1913.

egnees:
A counter read walked every worker's match set as soon as the first worker had one, so a worker the published generation carries no execution context for was a NULL dereference. The earlier take leaned on an all-or-none rule — a generation carries a context for every worker or for none. That rule does not hold, so the read now works from the per-worker premise: every worker owns its own execution context, and whether it has one says nothing about its neighbours.

The mixed state is reachable, not theoretical. A generation freezes its context count at install time while the read sizes itself by the live worker count, and workers register one at a time, so a worker that comes up after the last install has no context in a generation the read still walks.

- Describe the handle list from the first worker whose execution context the generation carries, so a read no longer answers empty when the first worker has none.
- Report zeros for a worker the generation carries no context for, instead of indexing its absent match set and dereferencing NULL.
- Note that the cross-worker ordering the read leans on comes from one generation building every context in a single pass, so a design that installs them per worker has to establish it another way.
- Cover a generation whose execution context is absent on a worker in the counter read tests; it segfaults against current main, and drops every counter from the answer if only the dereference is guarded.

Closes #1926.

### DIFF
--- a/lib/controlplane/agent/counters.c
+++ b/lib/controlplane/agent/counters.c
@@ -135,7 +135,9 @@ cp_counter_storage_copy_tags(const struct cp_counter_storage *storage) {
 // newly allocated memory, stashed behind the handle's opaque value array.
 //
 // The pointer array and every worker's value block live in one allocation:
-// the pointer array first, then each worker's values back to back.
+// the pointer array first, then each worker's values back to back. A
+// worker handed no storage reads back as zeros, which is what it has
+// contributed.
 //
 // The caller guarantees the per-worker storages stay alive for the
 // duration of this call.
@@ -163,6 +165,10 @@ counter_handle_copy_values(
 			continue;
 		}
 		values[w_idx] = value_blocks + w_idx * dst->size;
+		if (worker_storages[w_idx] == NULL) {
+			memset(values[w_idx], 0, dst->size * sizeof(uint64_t));
+			continue;
+		}
 		struct counter_value_handle *handle =
 			counter_get_value_handle(idx, worker_storages[w_idx]);
 		memcpy(values[w_idx],
@@ -212,7 +218,9 @@ fill_counter_handle(
 // a worker without an execution context. Every worker registry is built
 // by the same execution-context traversal, so the match order is
 // identical across workers: for match index i, every worker's array
-// names the same tag set at that index.
+// names the same tag set at that index. That alignment holds because one
+// generation builds every worker's context in a single pass; installing
+// contexts per worker would have to establish it another way.
 struct worker_counter_matches {
 	struct cp_counter_storage ***by_worker;
 	uint64_t worker_count;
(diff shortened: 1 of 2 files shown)


GeorgyKirichenko:
this was completely redesigned

PR #2124, opened 2026-08-18
fix(counters): handle counter storage allocation failures

### LINKED ISSUE #1969: fix(counters): counter_storage_spawn dereferences an unchecked memory_balloc result on arena exhaustion
## Motivation

In `counter_storage_spawn`'s per-pool block-allocation loop, the branch that allocates a brand-new block calls `memory_balloc` and writes through the result without checking it for NULL:

https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/lib/counters/counters.c#L417-L423

`refcnt` is the first field of `struct counter_storage_block`, at offset 0:

https://github.com/yanet-platform/yanet2/blob/3b8bc59994e21f7278a525d53c1f1b160825c4d1/lib/counters/counters.h#L68-L71

So an allocation failure here turns into a write to address `0x0` — a `SEGV_MAPERR` that kills the control-plane process, instead of returning a clean error.

Every sibling allocation in the same function (the `counter_storage` itself, the per-pool block array, the page, the handle array) *is* checked and produces a clean `yanet_error`. An automated scan found this to be the only unchecked `memory_balloc`/`memory_brealloc` on the control-plane config-update path.

Confirmed dynamically by a bug-hunter run with a deterministic repro: a starved `memory_context` reaches this line with a NULL return and faults with `movq $0x1,(%rax)`, `rax == 0`.

The window is narrow but structurally generated rather than coincidental: the 8-byte block-array allocation immediately preceding it splits the arena's last 16-byte block, leaving only its buddy, so the very next unchecked 16-byte request is the one that fails. Measured at roughly 0.5-2% of exhaustion trajectories per `counter_storage_spawn` call, and `config_gen_ectxs_create` calls that function from seven sites for every device, pipeline, function, chain and module x every worker, all draining the same arena.

Surfaced during investigation of a SIGSEGV observed on a production control plane, which turned out to have a different root cause; this defect was found and confirmed along the way and is not the production crash itself.

### Relationship to #1930

Mutually exclusive per call. This crash happens *before* any `goto error`, so #1930 (uninitialized `counter_value_handles` freed on the error path) is unreachable on that trajectory. On the far more common trajectory where the page allocation fails instead, control reaches `error:` and #1930 fires there. Fixing this issue alone converts every crash into a `goto error` that then runs straight into #1930, so the two should ship in one change.

## Scope

- Check the `memory_balloc` result at this call site and `goto error` on NULL, alongside #1930's fix, in the same commit.
- Consider zeroing the whole `struct counter_storage` at init as hardening, since it is `memory_balloc`'d without a `memset`.

## Acceptance

- [ ] `memory_balloc` result at the block-allocation call site is checked and a NULL return takes `goto error` instead of being dereferenced.
- [ ] Regression test: a C unit test builds a `block_allocator` over a small arena (with the counter registry kept in a separate, roomy context) and asserts `counter_storage_spawn` returns NULL rather than faulting, sweeping the starved arena size across a range in 8-byte steps — a single hand-picked size misses the second window class that appears once a pool holds more than one block.

Source: surfaced by bug-hunter dynamic confirmation during investigation of a SIGSEGV reported against a production control plane, which turned out to have a different root cause; this defect was found and confirmed along the way and is not the production crash. Related: #1930 (same function, mutually exclusive error path — ship together).

semenovad:
- Check block allocations before dereferencing them
- Initialize partially created storage for safe cleanup after OOM
- Add regression coverage for single-block and multi-block storage allocation failures

### DIFF
--- a/lib/counters/counters.c
+++ b/lib/counters/counters.c
@@ -310,26 +310,15 @@ counter_storage_new_page(struct memory_context *memory_context) {
 	return page;
 }
 
-// Initialize a counter_storage_pool to the zero-init state.
-//
-// Pool entries are allocated lazily on first counter registration, so there
-// is nothing to set up beyond zeroing the struct.
-static void
-counter_storage_pool_init(struct counter_storage_pool *self) {
-	memset(self, 0, sizeof(*self));
-}
-
 static void
 counter_storage_init(
 	struct memory_context *memory_context,
 	struct counter_storage *storage,
 	struct counter_registry *registry
 ) {
+	memset(storage, 0, sizeof(*storage));
 	SET_OFFSET_OF(&storage->memory_context, memory_context);
 	SET_OFFSET_OF(&storage->registry, registry);
-	for (uint64_t idx = 0; idx < COUNTER_POOL_SIZE; ++idx) {
-		counter_storage_pool_init(storage->pools + idx);
-	}
 }
 
 struct counter_storage *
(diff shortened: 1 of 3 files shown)


semenovad:
@codex review

GeorgyKirichenko (approved):

PR #2143, opened 2026-08-19
fix(route-mpls,encap): pass unmatched packets through and emit valid mpls-over-udp checksums

### LINKED ISSUE #2059: test(route-mpls): wire module into dataplane_ut so in-process tests can execute it

### LINKED ISSUE #2061: fix(route-mpls): invalid filter verdict drops the popped packet without output or drop

3Hren:
- Link route-mpls into the dataplane_ut cgo group so the module is reachable from in-process Go tests, and rename its static library to the sibling `_dp` convention.
- Add `modules/route-mpls/tests/functional`: the inner/outer family combinations are compared byte for byte against a gopacket-serialized reference (lengths and checksums included, even and odd inner lengths), plus pass-through and no-tunnel nexthops, weighted ECMP by hash, rule ordering, a mixed IPv4/IPv6/non-IP burst and the empty poll round.
- The empty-round test can only fail under a sanitizer build; CI runs it that way through `make test-asan-only`, so keep it even though it looks vacuous in a plain build.
- Stop losing IP packets that match no rule: the handler popped them from the input without output or drop, which leaked the mbuf; they now pass through like every other packet the module has nothing to do with.
- Beyond the two issues below, fix the MPLS-over-UDP encapsulation checksums the new byte-exact tests exposed: the outer IPv4 header checksum was the uncomplemented sum, and the UDP pseudo-header used the IP total length and misplaced the protocol byte in both families. Route-mpls is the only caller and the dataplane uses no TX checksum offload, so these were the wire values.
- The UDP encapsulation still sizes and sums the head mbuf segment only, as its existing FIXME notes; that multi-segment limitation is out of scope here.
- Closes #2059.
- Closes #2061.

### DIFF
--- a/bindings/go/dataplane_ut/dataplane_ut.go
+++ b/bindings/go/dataplane_ut/dataplane_ut.go
@@ -14,6 +14,7 @@ package dataplaneut
 #cgo LDFLAGS: -L../../../build/modules/forward/dataplane
 #cgo LDFLAGS: -L../../../build/modules/mirror/dataplane
 #cgo LDFLAGS: -L../../../build/modules/route/dataplane
+#cgo LDFLAGS: -L../../../build/modules/route-mpls/dataplane
 #cgo LDFLAGS: -L../../../build/modules/nat64/dataplane
 #cgo LDFLAGS: -L../../../build/modules/pdump/dataplane
 #cgo LDFLAGS: -L../../../build/devices/plain/dataplane
@@ -37,7 +38,7 @@ package dataplaneut
 // references between them (fwstate->acl, worker->pipeline, etc.).
 // fwstate depends on acl — acl must come first inside the group.
 #cgo LDFLAGS: -Wl,--start-group
-#cgo LDFLAGS: -lblackhole_dp -ldecap_dp -ldscp_dp -lacl_dp -lfwstate_dp -lforward_dp -lmirror_dp -lroute_dp -lnat64_dp -lpdump_dp
+#cgo LDFLAGS: -lblackhole_dp -ldecap_dp -ldscp_dp -lacl_dp -lfwstate_dp -lforward_dp -lmirror_dp -lroute_dp -lroute_mpls_dp -lnat64_dp -lpdump_dp
 #cgo LDFLAGS: -lplain_dp -lvlan_dp
 #cgo LDFLAGS: -ldataplane_ut -lpipeline -lmodule -lworker_dp -lconfig_dp -lpacket
 #cgo LDFLAGS: -llogging -lagent -lconfig_cp -lcounters -lerrors -lfilter_compiler -lfwstate -llib_utils
@@ -119,6 +120,7 @@ keep_refs(void **ptrs) {
 	extern struct module *new_module_forward(void);
 	extern struct module *new_module_mirror(void);
 	extern struct module *new_module_route(void);
+	extern struct module *new_module_route_mpls(void);
 	extern struct module *new_module_nat64(void);
 	extern struct module *new_module_pdump(void);
 
@@ -134,6 +136,7 @@ keep_refs(void **ptrs) {
 		new_module_forward,
 		new_module_mirror,
 		new_module_route,
+		new_module_route_mpls,
 		new_module_nat64,
 		new_module_pdump,
 
--- a/lib/dataplane/packet/encap.c
+++ b/lib/dataplane/packet/encap.c
@@ -241,6 +241,24 @@ packet_mpls_encap(
 	return 0;
 }
 
+// One's complement sum of the UDP pseudo-header for 4-byte (IPv4) or
+// 16-byte (IPv6) addresses and the network-order UDP length.
+//
+// The pseudo-header is the two addresses, the protocol number zero-padded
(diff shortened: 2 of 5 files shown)


3Hren:
@codex review

GeorgyKirichenko (approved):

PR #2144, opened 2026-08-19
refactor(fwstate)!: drop optional from sync_suppress_timeout

GeorgyKirichenko:
- Make `sync_suppress_timeout` a plain proto3 scalar consistent with every other `SyncConfig` timeout field; the merge in `ToCWithDefaults` now treats an omitted or zero value as keep the current window.
- Breaking: an explicit zero no longer disables suppression through `UpdateConfig`; once set, the window changes only via another non-zero value or by recreating the config. Fresh configs still default to suppression disabled.
- Wire format is unchanged (field 13, varint); old and new clients interoperate in both directions.
- The fwstate CLI assigns the plain scalar and its `--sync-suppress-timeout` doc now says omitted or zero keeps the configured window.

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -26,3 +26,9 @@ breaking:
     - PACKAGE
   ignore:
     - modules/balancer2
+  # sync_suppress_timeout intentionally dropped proto3 optional presence; the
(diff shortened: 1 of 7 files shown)


GeorgyKirichenko:
@codex review

PR #2150, opened 2026-08-19
fix(dataplane): bind worker counters by name instead of hard-coded indices

### LINKED ISSUE #2093: fix(dataplane): bind worker counters by registered index instead of hard-coded literals
## Motivation

The dataplane binds worker counter slots positionally. [dataplane/worker.c:534-555](https://github.com/yanet-platform/yanet2/blob/b19500deb7fa7edf4c86fb86db211c736e11df9e/dataplane/worker.c#L534-L555) assigns nine `dp_worker->*` pointers from `counter_get_address(N, worker_counter_storage)` with `N` written as the literals 0..8. An existing `// FIXME: do not use hard-coded counter identifiers` sits directly above it at [dataplane/worker.c:533](https://github.com/yanet-platform/yanet2/blob/b19500deb7fa7edf4c86fb86db211c736e11df9e/dataplane/worker.c#L533).

The control plane resolves the same counters by name. [controlplane/ffi/worker.go:41-60](https://github.com/yanet-platform/yanet2/blob/b19500deb7fa7edf4c86fb86db211c736e11df9e/controlplane/ffi/worker.go#L41-L60) builds a `map[string]*C.struct_counter_handle` keyed on the handle name and looks up `iterations`, `rx`, `tx`, `remote_rx`, `remote_tx`, `rx_bursts`, `local_tx_drops`, `remote_tx_drops`, `drops`.

Nothing connects the two addressing schemes. The only thing keeping them aligned is the statement order inside `worker_counters_register` ([lib/dataplane/worker/counters.c:26-54](https://github.com/yanet-platform/yanet2/blob/b19500deb7fa7edf4c86fb86db211c736e11df9e/lib/dataplane/worker/counters.c#L26-L54)), which registers exactly those nine names in exactly that order.

Inserting or removing a counter anywhere except the end of `worker_counters_register` silently shifts every later slot. The dataplane then writes `rx` into what the registry calls `tx`, while every Go name still resolves successfully. No nil handle, no error, no crash — just wrong numbers reported to the operator and to metrics, which is a harder failure mode to detect than a crash.

The fix is small and already afforded by the API: `counter_registry_register` already returns the assigned index. `register_one` at [lib/dataplane/worker/counters.c:8-22](https://github.com/yanet-platform/yanet2/blob/b19500deb7fa7edf4c86fb86db211c736e11df9e/lib/dataplane/worker/counters.c#L8-L22) captures it as `uint64_t rc`, uses it only to compare against `COUNTER_INVALID`, and then discards it. Propagating those returned indices to the `counter_get_address` call sites removes the drift entirely, with no change to the shared-memory layout or to the counter registry itself.

## Scope

- Propagate the indices returned by `worker_counters_register` (via `register_one`) through to the `counter_get_address` calls in `dataplane/worker.c:534-555`, replacing the hard-coded 0..8 literals.
- No shared-memory layout change and no change to `counter_registry_register`/`counter_get_address` themselves.

## Acceptance

- [ ] `dataplane/worker.c` binds each `dp_worker->*` counter pointer using the index returned at registration time, not a literal.
- [ ] Reordering or inserting a counter in `worker_counters_register` no longer silently mislabels a later slot (covered by a C unit test or fuzz check that asserts binding survives a reorder).
- [ ] The `// FIXME: do not use hard-coded counter identifiers` comment is removed once the fix lands.

Source: found while designing the counter-lookup API for #1935 — the dataplane binds worker counter slots by hard-coded index while the control plane resolves them by name, with nothing tying the two orders together. Related: #1935 (nil-guard for worker counter lookups, open) and #1934 (closed) — #1934's Scope explicitly flagged that "name and index can drift apart" but left it undecided and out of that issue's scope.

egnees:
fix(dataplane): bind worker counters by name instead of hard-coded indices

- The dataplane bound worker counter slots with the literal indices 0..8, while the control plane resolves the same counters by name, so inserting or removing a counter anywhere but at the end silently shifted every later slot and reported wrong numbers with no error.
- Describe the worker counters in two tables in `lib/dataplane/worker/counters.c`: one lists the name and size of every counter that is registered, the other maps each worker pointer field to a counter name and a value index inside it.
- Add `worker_counters_bind`, which resolves every field by counter name, checks that the value index fits the size the counter was registered with, and leaves the worker unchanged if any field fails to resolve.
- `dataplane_worker_start` and `dataplane_ut_new` now call `worker_counters_bind` instead of repeating the index arithmetic, and both the `FIXME` in `dataplane/worker.c` and the hard-coded index enum in `dataplane_ut` are gone.
- Declare `counter_registry_lookup_index` in the public header and add `counter_registry_lookup_size`, so the size of a counter can be checked at bind time. No shared-memory layout change.
- Log which worker failed to start in `dataplane_device_start`.
- Add `lib/tests/dataplane/worker_counters_test.c`: it checks that every field lands on the address its own name and value index resolve to, and that a renamed or undersized counter makes binding fail without half-wiring the worker.
- Reordering or inserting a counter in the specification table no longer mislabels a later slot.

Closes #2093.

### DIFF
--- a/dataplane/device.c
+++ b/dataplane/device.c
@@ -89,6 +89,7 @@ dataplane_device_start(
 	for (uint32_t wrk_idx = 0; wrk_idx < device->worker_count; ++wrk_idx) {
 		struct dataplane_worker *worker = device->workers + wrk_idx;
 		if (dataplane_worker_start(worker)) {
+			LOG(ERROR, "failed to start worker %u", wrk_idx);
 			return -1;
 		}
 	}
--- a/dataplane/worker.c
+++ b/dataplane/worker.c
@@ -42,6 +42,7 @@
 #include "dataplane/dataplane.h"
 #include "dataplane/device.h"
 
+#include "lib/counters/counters.h"
 #include "lib/dataplane/config/zone.h"
 #include "lib/dataplane/module/packet_front.h"
 #include "lib/dataplane/packet/data.h"
@@ -61,6 +62,7 @@
 
 #include <rte_ethdev.h>
 
+#include <stddef.h>
 #include <string.h>
 
 static void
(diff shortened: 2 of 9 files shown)


egnees:
Closed as overcomplicated. See another approach to this issue in #2163

PR #2154, opened 2026-08-19
fix(dataplane): bound packet recirculation

### LINKED ISSUE #2051: fix(dataplane): a module routing a packet into its own device entry loses the packet and leaks the mbuf
## Motivation

In [`lib/dataplane/worker/pipeline_round.c`](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/lib/dataplane/worker/pipeline_round.c#L63-L119), `worker_pipeline_round` processes each device entry point and then merges it into the round front with `packet_front_merge(packet_front, schedule)`, for both the input pass and the output pass. [`packet_front_merge`](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/lib/dataplane/module/packet_front.h#L81-L96) ends with `packet_front_init(src)`, which resets all three of `src`'s lists, including `input`.

`src` here is the device entry's own schedule, and `schedule.input` is exactly where [`module_ectx_route_input`/`module_ectx_route_output`](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/lib/controlplane/config/econtext.h#L37-L88) deposit a redirected packet — they call `packet_front_input(&device_ectx->{input,output}_pipelines->schedule, packet)` on the *target* device's schedule. When a module routes a packet to the device entry that is currently executing, that write lands on the very `schedule` object the round loop is about to merge and reset. The packet is silently unlinked with no `packet_free`/`packet_drop` call, so in production this is a permanent mempool leak that ends in pool exhaustion and total traffic loss on the affected worker.

The `packet_front_merge` doc comment asserts src never holds pending lists at merge time — that premise stopped being true for device entry schedules once routing started writing straight into them.

This is a regression from `refactor(dataplane): route packets straight to device entry schedules` (695652a8, 2026-08-12). Before it, modules appended to `pending_input`/`pending_output` lists that the merge concatenated upward, and the next iteration demuxed them into the target entries, so resetting `src` was harmless.

## Evidence

Reproduced with a scratch self-transmitting module emitting one frame and routing it to a chosen device index. The frame is lost if and only if the target entry is the entry currently being processed — perfectly symmetric, and independent of worker count and device count: module on device 0 targeting device 0 loses it, on device 0 targeting device 1 survives, on device 1 targeting device 0 survives, on device 1 targeting device 1 loses it; the same four results hold for the output entry.

Confirmed by an independent bug-hunter pass with a reproducible A/B table, and the mechanism above was verified directly against the pushed source at 695652a8.

## Blast radius

Reachable in production whenever a rule's target device equals the device whose entry the module runs on, not just in a test harness:

- `forward` — [`modules/forward/dataplane/dataplane.c`](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/modules/forward/dataplane/dataplane.c#L136-L142) `route_input`/`route_output` call sites.
- `mirror` — [`modules/mirror/dataplane/dataplane.c`](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/modules/mirror/dataplane/dataplane.c#L155-L164).
- `route`, when placed on a device output entry resolving to that same device — [`modules/route/dataplane/dataplane.c`](https://github.com/yanet-platform/yanet2/blob/695652a8f193144ae541e780b1cd6237df665ddf/modules/route/dataplane/dataplane.c#L240-L247).

Not gated on worker or device count: a single-worker single-device deployment with a self-targeting `forward` or `mirror` rule leaks every matched packet. No public-tree Go test covers this shape.

## Open design question (blocks implementation)

The obvious fix — stop the merge from wiping `src->input` — restores the pre-refactor recirculation behaviour, but every in-tree module re-decides identically on the second pass, so it converts the leak into a non-terminating round: a self-targeting `forward` `MODE_IN` rule matches again every pass, and a self-targeting `mirror` `MODE_IN` clones again every pass (unbounded loop plus exponential packet growth). The round loop has no iteration bound. The remedy is a real choice between preserving recirculation, dropping with accounting, or preserving with a recirculation budget, and it needs a decision before implementation.

## Acceptance

- [ ] Bug-hunter confirmation of the mechanism above is on record (this issue carries it; treat as confirmed pending any refutation).
- [ ] The mbuf is no longer leaked when a module routes a packet into the entry currently being processed.
- [ ] The chosen behaviour (drop-with-accounting, budgeted recirculation, or unbounded recirculation restored) is deliberate and observable to an operator, not a silent packet loss.
- [ ] A permanent Go regression test in the dataplane_ut harness pins the behaviour, with the cross-device case as a passing control.

Source: surfaced during an autopilot run on #1724 while investigating that issue's dataplane_ut corruption symptom. This is a distinct defect from #1724, not a duplicate — #1724 is data corruption in the harness capture/copy path for a worker>=2/device>=2 topology, this is a packet/mbuf loss on self-targeting device routing, reproducible with a single worker and independent of #1724's topology precondition.

moonug:
- Bounds packet recirculation with a configurable lineage limit and an independent stall limit.
- Count route attempts and target-device drops while keeping counters and packet metadata consistent across decapsulation and cloning.
- Preserve fixed packet sets, payload snapshots, and ownership when dataplane UT rounds are replayed.
- Add C, dataplane UT, CGO binding, clone, and functional regression coverage.
- Closes #2051.

### DIFF
--- a/lib/controlplane/config/econtext.h
+++ b/lib/controlplane/config/econtext.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "lib/counters/counters.h"
+#include "lib/dataplane/packet/data.h"
 #include "lib/dataplane/pipeline/econtext.h"
 
 #include "lib/errors/errors.h"
@@ -27,13 +29,32 @@ config_gen_ectx_get_object(
 	return ADDR_OF(objects + index);
 }
 
+// Count a packet dropped after exhausting its device recirculation budget.
+//
+// Counter slot 0 stores packets. Slot 1 stores the full packet length across
+// all chained mbufs, unlike the first-segment byte tallies on packet fronts.
+static inline void
+device_entry_ectx_count_recirc_drop(
+	struct device_entry_ectx *entry_ectx, const struct packet *packet
+) {
+	uint64_t *counter = counter_handle_get_value(
+		ADDR_OF_NONNULL(&entry_ectx->counter_packet_recirc_drop)
+	);
+	counter[0] += 1;
+	counter[1] += rte_pktmbuf_pkt_len(packet_to_mbuf(packet));
+}
+
 // Route a packet to its target device's input entry, counting it as
 // pending_input on the originating schedule.
 //
 // The packet lands on the target device entry's schedule input so the next
 // round picks it up. When the target device is absent from this generation
 // the packet is dropped on the originating schedule instead of being
-// stranded. packet->tx_device_id must already name the destination.
+// stranded. Input and output routes share the packet lineage budget; an
+// exhausted budget drops on the originating schedule and increments the target
+// entry's input_recirc_drop counter. packet->tx_device_id names the
+// destination. The pending counters count every route attempt, including
+// dropped packets.
 static inline void
 module_ectx_route_input(
 	struct module_ectx *module_ectx,
@@ -52,9 +73,16 @@ module_ectx_route_input(
 		packet_front_drop(packet_front, packet);
 		return;
 	}
-	packet_front_input(
-		&ADDR_OF(&device_ectx->input_pipelines)->schedule, packet
-	);
+	struct device_entry_ectx *entry_ectx =
+		ADDR_OF(&device_ectx->input_pipelines);
+	if (!packet_recirc_try_redirect(
+		    packet, module_ectx->packet_recirc_limit
+	    )) {
+		device_entry_ectx_count_recirc_drop(entry_ectx, packet);
+		packet_front_drop(packet_front, packet);
+		return;
+	}
+	packet_front_input(&entry_ectx->schedule, packet);
 }
 
 // Route a packet to its target device's output entry, counting it as
@@ -63,7 +91,10 @@ module_ectx_route_input(
 // Symmetric to module_ectx_route_input: the packet is placed on the target
 // device's output-pipelines schedule, or dropped on the originating schedule
 // when the device is gone. packet->tx_device_id must already name the
-// destination.
+// destination. Input and output routes share the packet lineage budget; an
+// exhausted budget drops on the originating schedule and increments the target
+// entry's output_recirc_drop counter. The pending counters count every route
+// attempt, including dropped packets.
 static inline void
 module_ectx_route_output(
 	struct module_ectx *module_ectx,
@@ -82,9 +113,16 @@ module_ectx_route_output(
 		packet_front_drop(packet_front, packet);
 		return;
 	}
-	packet_front_input(
-		&ADDR_OF(&device_ectx->output_pipelines)->schedule, packet
-	);
+	struct device_entry_ectx *entry_ectx =
+		ADDR_OF(&device_ectx->output_pipelines);
+	if (!packet_recirc_try_redirect(
+		    packet, module_ectx->packet_recirc_limit
+	    )) {
+		device_entry_ectx_count_recirc_drop(entry_ectx, packet);
+		packet_front_drop(packet_front, packet);
+		return;
+	}
+	packet_front_input(&entry_ectx->schedule, packet);
 }
 
 // Build one execution context per worker.
--- a/lib/dataplane/worker/worker.c
+++ b/lib/dataplane/worker/worker.c
@@ -35,9 +35,9 @@ worker_clone_packet(
 	struct packet *packet_clone = mbuf_to_packet(mbuf);
 	// Order is load-bearing: init the source budget first (no-op when the
 	// lineage is already initialized, so an already-spent source splits its
-	// remainder), snapshot into the clone, then partition floor/ceil. A swap
-	// of init and memcpy would copy uninitialized fields and re-init the
-	// clone to the full limit on its first redirect.
+	// remainder), snapshot into the clone, then partition floor/ceil. A
+	// swap of init and memcpy would copy uninitialized fields and re-init
+	// the clone to the full limit on its first redirect.
 	packet_recirc_init(packet, packet_recirc_limit);
 	rte_memcpy(packet_clone, packet, sizeof(struct packet));
 	packet_clone->mbuf = mbuf;
--- a/lib/dataplane/worker/worker.c
+++ b/lib/dataplane/worker/worker.c
@@ -32,9 +32,8 @@ worker_clone_packet(
 		return NULL;
 	}
 
-	packet_recirc_init(packet, packet_recirc_limit);
-
 	struct packet *packet_clone = mbuf_to_packet(mbuf);
+	packet_recirc_init(packet, packet_recirc_limit);
 	rte_memcpy(packet_clone, packet, sizeof(struct packet));
 	packet_clone->mbuf = mbuf;
 	packet_clone->next = NULL;
--- a/lib/dataplane/worker/worker.c
+++ b/lib/dataplane/worker/worker.c
@@ -3,6 +3,8 @@
 #include "lib/dataplane/packet/data.h"
 #include "lib/dataplane/packet/packet.h"
 
+#include <string.h>
+
 struct packet *
 worker_packet_alloc(struct dp_worker *dp_worker) {
 	struct rte_mbuf *mbuf = rte_pktmbuf_alloc(dp_worker->rx_mempool);
@@ -11,6 +13,7 @@ worker_packet_alloc(struct dp_worker *dp_worker) {
 	}
 
 	struct packet *packet = mbuf_to_packet(mbuf);
+	memset(packet, 0, sizeof(*packet));
 	packet->mbuf = mbuf;
 
 	return packet;
--- a/lib/dataplane/worker/worker.c
+++ b/lib/dataplane/worker/worker.c
@@ -20,7 +20,11 @@ worker_packet_alloc(struct dp_worker *dp_worker) {
 }
 
 struct packet *
-worker_clone_packet(struct dp_worker *dp_worker, struct packet *packet) {
+worker_clone_packet(
+	struct dp_worker *dp_worker,
+	struct packet *packet,
+	uint16_t packet_recirc_limit
+) {
 	struct rte_mbuf *mbuf = rte_pktmbuf_copy(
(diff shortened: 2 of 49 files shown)


moonug:
## Review: PR #2154 — Bound packet recirculation (fix #2051)

### Summary

The PR fixes #2051 (self-loop packet loss + mbuf leak) by:
1. Detaching the device schedule batch into a local `active` front before processing, so packets routed back into the schedule survive the `packet_front_merge` reset.
2. Adding a per-packet-lineage redirect budget (`recirc_remaining` / `recirc_initialized`) with a configurable limit (4–256, default 64).
3. Counting budget-exhausted drops on the target device (`input_recirc_drop` / `output_recirc_drop`).
4. Partitioning the budget on clone (source keeps odd credit, clone gets floor).
5. Bumping `YANET_MODULE_ABI_VERSION` 17→18.
6. Threading `packet_recirc_limit` through `module_ectx` to all clone sites (mirror, unrdup).
7. Defensive `bytes.Clone` in the Go test harness to prevent use-after-free on recirculated packet data.

### Core fix (pipeline_round.c)

The root-cause fix is correct and elegant: copying `schedule` into `active`, then `packet_front_init(schedule)` before processing means any packets routed back into the device entry during processing land in the fresh `schedule` and survive the next iteration. Previously, `packet_front_merge` would reset the shared schedule, silently losing recirculated packets and leaking their mbufs.

### Budget mechanics (packet.h / packet.c)

- `packet_recirc_try_redirect` decrements `recirc_remaining` and returns false when exhausted — correct.
- `packet_recirc_init` sets `recirc_remaining` to the configured limit and marks `recirc_initialized = 1` on first redirect. Lazy init — packets that never redirect pay no cost.
- Clone partitioning: source keeps `(remaining + 1) / 2` (odd credit), clone gets `remaining / 2`. The invariant `src + clone == original` is tested. The load-bearing initialization order (init source first, then memcpy, then partition) is documented in `worker.c` — good.

### Config parsing (config/zone.c)

- YAML key `packet_recirc_limit` is parsed, clamped to [4, 256], default 64 when unset or 0.
- The clamp rejects degenerate values below the minimum internal header delta — follows the "zero config limits mean unset/no clamp" convention.

### Tests

**C unit tests**: alloc reset, multi-segment clone partitioning, jumbo clone, recirc drop byte counting, poison pattern verification — thorough.

**Go dataplane_ut tests**: decap nested IPIP (5 layers, limit table-driven, miss-no-renew), forward self-loop/cross-device/mixed-mode, mirror bounded population with mbuf reclamation, route self-loop — comprehensive.

**Functional (QEMU)**: boots VM with `packet_recirc_limit: 5` in YAML, configures forward self-loop, verifies counters via `yanet-cli-counters` — end-to-end validation.

### Observations (non-blocking)

1. **QEMU pgrep refactor**: `[q]emu-system-x86_64` self-match prevention + `([[:space:]]|$)` boundary — well-tested with 6 unit cases. Clean testable abstraction.
2. **`bytes.Clone` in Go harness**: prevents use-after-free when mbuf is recycled but Go-side Payload still references its buffer. Good defensive fix for recirculation.
3. **ABI version bump**: 17→18 correct since `struct packet` layout changed.
4. **`fill_packet_net4`/`net6` memset**: more correct than previous partial init that left recirc fields uninitialized.

### Verdict

Well-structured, addresses root cause correctly, thorough test coverage at every level. Iterative review history (12 commits with review-fix commits) shows refinement. No blocking issues found.

moonug:
## Code Review Summary

**Verdict: Comment** (no blocking issues; 0 critical, 0 warnings, 3 suggestions)

This PR bounds packet recirculation (self-redirects) via a per-packet-lineage credit budget. The design is clean, well-tested at unit and functional levels, and the ABI version bump is handled correctly. Below are non-blocking observations.

---

### Architecture

The recirculation budget is implemented as a pair of fields in `struct packet`:
- `recirc_remaining` (uint16_t): credits left, consumed by `packet_recirc_try_redirect`
- `recirc_initialized` (uint8_t): lazy-init flag so packets entering the pipeline before the module context is available default to "full budget"

The budget flows: `dataplane.yaml` (`packet_recirc_limit`) -> `dp_config->packet_recirc_limit` -> `module_ectx->packet_recirc_limit` -> consumed in `module_ectx_route_input/output` and partitioned in `worker_clone_packet`.

The pipeline round (`worker_pipeline_round`) now **detaches** each device entry's schedule before processing (`active = *schedule; packet_front_init(schedule)`), so packets routed back during processing land in the reusable schedule inbox for the next iteration. This is the core mechanism that enables recirculation.

### ABI / Struct Sizes

- `struct packet`: 48 -> 56 bytes (added `recirc_remaining` + `recirc_initialized` + padding)
- `struct module_ectx`: 184 -> 192 bytes (added `packet_recirc_limit`)
- `struct device_entry_ectx`: new `counter_packet_recirc_drop` handle (now 216 bytes)
- `YANET_MODULE_ABI_VERSION` bumped 17 -> 18
- All `_Static_assert` size tripwires in `plugin_abi_assert.h` updated
- `dp_config` stays 1184 bytes (`uint16_t packet_recirc_limit` fits in existing padding after `uint32_t numa_idx` before the 8-byte `storage_size`)

### Clone Budget Partition

`worker_clone_packet` does init-then-partition: source is lazily initialized to the limit, then the clone gets `remaining / 2` (floor) and source keeps `remaining - floor` (ceil). The comment in `worker.c` explaining the load-bearing ordering (init source -> memcpy -> partition) is excellent. Verified:
- Source uninitialized + limit=5: source gets 3, clone gets 2
- Source exhausted (0 remaining): clone gets 0 — both blocked from further redirects

### Test Coverage

- **Unit**: `worker_clone_test.c` — clone budget partition, jumbo frames, recirc drop counter, uninitialized source. `packet_recirc_test.c` — `packet_recirc_init` and `packet_recirc_try_redirect` edge cases. `parse_frag_test.c` — memset prevents stale recirc fields.
- **Functional**: `forward_test.go::Test_PacketRecircLimit_FromDataplaneConfig` — creates a self-loop (device output -> forward -> same device), sends one packet, verifies `output_rx` increments by exactly `testPacketRecircLimit` (5) and `output_recirc_drop` increments by 1 packet + packetSize bytes. This is end-to-end proof the bound works.
- **Smoke**: `smoke_test.c` — recirculation round geometry rejection test.

### Convention Compliance

- `memory_balloc` does not zero -> `worker_packet_alloc` now does `memset(packet, 0, sizeof(*packet))`. Correct.
- `fill_packet_net4/net6` now `memset` to prevent stale recirc fields from stack-allocated `struct packet`. Correct.
- CGO boundary: `lib/controlplane/config/econtext.h` now includes `data.h` (pulls `<rte_mbuf.h>`), but this header is NOT on the CGO path (only included by `.c` source files, not by CGO-facing headers like `api/agent.h`). Not a violation, but worth monitoring if future code adds it to a CGO path.
- Multi-segment mbufs: `device_entry_ectx_count_recirc_drop` correctly uses `rte_pktmbuf_pkt_len` (whole-packet length) for the byte counter, not `data_len` (head-only). Correct.

---

### Suggestions (non-blocking)

1. **`lib/controlplane/config/econtext.h:3`** — The new `#include "lib/dataplane/packet/data.h"` pulls `<rte_mbuf.h>` into a control-plane header. While not on the CGO path today, this header is one `#include` away from a CGO violation. Consider whether `device_entry_ectx_count_recirc_drop` could live in a `.c` file or a dataplane-only header to keep the control-plane headers DPDK-free.

2. **`lib/dataplane/pipeline/econtext.h:47`** — `module_ectx->packet_recirc_limit` is `uint16_t` and the limit is clamped at config ingestion, but there is no runtime assert that `module_ectx->packet_recirc_limit >= PACKET_RECIRC_LIMIT_MIN` (4). A misconfigured build or future code path could set 0, which would drop all routed packets immediately. Consider a debug assert or documenting the invariant.

3. **`lib/dataplane/worker/worker.c:36-48`** — The clone budget partition comment explains the init-memcpy-partition ordering, but the partition itself (`clone = remaining / 2; source -= clone`) silently floors. When `remaining == 1`, the clone gets 0 and the source keeps 1 — meaning the clone inherits an exhausted budget and will be dropped on its first redirect. This is correct behavior (the lineage's total budget is preserved), but it might surprise a future reader. A one-line comment on the floor semantics would help.

---

*Reviewed by Hermes Agent (reviewer-c profile)*

lib/controlplane/config/econtext.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "lib/counters/counters.h"
```
moonug: **Suggestion:** The new `#include "lib/dataplane/packet/data.h"` pulls `<rte_mbuf.h>` into a control-plane header. While not on the CGO path today, this header is one `#include` away from a CGO violation. Consider whether `device_entry_ectx_count_recirc_drop` could live in a `.c` file or dataplane-only header to keep the control-plane headers DPDK-free.

lib/dataplane/worker/worker.c
```diff
@@ -26,9 +33,17 @@ worker_clone_packet(struct dp_worker *dp_worker, struct packet *packet) {
...
+	// lineage is already initialized, so an already-spent source splits its
+	// remainder), snapshot into the clone, then partition floor/ceil. A
+	// swap of init and memcpy would copy uninitialized fields and re-init
+	// the clone to the full limit on its first redirect.
+	packet_recirc_init(packet, packet_recirc_limit);
 	rte_memcpy(packet_clone, packet, sizeof(struct packet));
 	packet_clone->mbuf = mbuf;
 	packet_clone->next = NULL;
+	packet_clone->recirc_remaining = packet->recirc_remaining / 2;
+	packet->recirc_remaining -= packet_clone->recirc_remaining;
```
moonug: **Suggestion:** The partition `clone = remaining / 2; source -= clone` silently floors. When `remaining == 1`, the clone gets 0 and will be dropped on its first redirect. This is correct (total budget is preserved), but a one-line comment on the floor semantics would help future readers.

moonug (commented):
.

PR #2155, opened 2026-08-20
feat(dataplane): accept a unit suffix in instance memory sizes

kriomant:
- Instance dp_memory and cp_memory in dataplane.yaml accept a KiB, MiB, GiB or TiB suffix, a bare number stays bytes.
- The parser is one static function next to parse_unsigned(), the suffix set is exact and binary, fractions and other spellings are rejected.
- The dpdk_memory value keeps its bare mebibyte meaning, it goes straight to EAL -m.
- The example configuration and the functional test harness use the unit form.

### DIFF
--- a/dataplane.yaml
+++ b/dataplane.yaml
@@ -1,13 +1,16 @@
 dataplane:
   storage: /dev/hugepages/yanet
-  dpdk_memory: 1024
+  # Memory sizes accept a unit suffix: b, k/kb, m/mb, g/gb, t/tb, binary and
+  # case-insensitive. Without a suffix dpdk_memory is MiB, which is also its
+  # smallest unit, while the instance sizes are bytes.
+  dpdk_memory: 1g
   # Maximum self-redirects (recirculations) per packet lineage, shared by the
   # input and output routes; 4..256, default 64. Packets exceeding the budget
   # are dropped and counted in input_recirc_drop / output_recirc_drop.
   packet_recirc_limit: 64
   instances:
-    - dp_memory: 1073741824
-      cp_memory: 2147483648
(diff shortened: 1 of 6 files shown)


3Hren:
This is a breaking change. If you want backward-compatible - add number-or-string support, where strings can be "1234 MiB" etc.

3Hren:
Thanks for the PR. I reworked it on top of main and force-pushed to your branch, so here is what changed and why.

- The first version switched `dp_memory` and `cp_memory` to bare mebibytes. That reads well, but it silently changes the meaning of every existing file, hence the earlier request for a number-or-string form.
- The last revision then grew a grammar that is more than we want the dataplane to carry: case-insensitive short spellings, a per-field granularity table so that `dpdk_memory` could take units as well, and the diagnostics around it. The spellings also made `MB` mean 2^20 bytes, and we do not want to break SI: `MB` is 10^6 and `MiB` is 2^20, and a config that says `1 MB` should not quietly mean something else. Hugepage sizing is binary by nature, so we settled on exactly `KiB`, `MiB`, `GiB` and `TiB` and nothing else, with a bare number still meaning bytes. Not accepting `MB` at all keeps the file honest instead of picking one reading of an ambiguous unit.
- `dpdk_memory` stays a bare mebibyte count: it goes straight to EAL `-m`, and a unit form for it means teaching the parser to reject sub-MiB values for one field.
- The standalone tests were dropped together with the grammar they covered. The existing config test already drives both instance fields through the new function, including the invalid-value sweep.

What is left is one 40-line static function next to `parse_unsigned()`, plus the example `dataplane.yaml` and the functional harness in the unit form.

Issue #2151, opened 2026-08-19
Allow overriding listen endpoints and `instance_id` via CLI flags or environment variables


borislitv:
## Title

Allow overriding listen endpoints and `instance_id` via CLI flags or environment variables

### Summary

Every YANET 2 process currently takes its listen endpoints and its
`instance_id` **exclusively** from the YAML config file. There is no way to
override them at launch time. This makes it impossible for an external
orchestrator to own the port layout, which is a hard requirement for running
YANET 2 under Kubernetes.

We would like a supported way to override at least the listen endpoints and
`instance_id` from outside the config file. **CLI flags or environment
variables — either is fine for us**, whichever fits the project's style better.
What matters is that the override exists and takes precedence over the file.

### Current behaviour

`yanet-controlplane` accepts only `--config/-c`:

```go
// controlplane/cmd/yncp-director/main.go
rootCmd.Flags().StringVarP(&cmd.ConfigPath, "config", "c", "", "Path to the configuration file (required)")
rootCmd.MarkFlagRequired("config")
```

All operators share the same single-flag command builder:

```go
// common/go/operator/cmd.go
root.Flags().StringVarP(... "config" ...)
root.MarkFlagRequired("config")
```

And the config loader is a plain read + unmarshal, with no environment
expansion or overlay:

```go
// common/go/xcfg/load.go
buf, err := os.ReadFile(path)
...
if err := Decode(buf, cfg, options...); err != nil { ... }
```

`grep -rn "os.Getenv" controlplane operators common cli` returns nothing, so
environment variables are not consulted anywhere in the config path.

### Why this blocks us

We are moving YANET 2 firewall hosts from hand-managed systemd units to a
Kubernetes operator. The operator generates one controlplane Deployment per
NUMA domain and is responsible for the Kubernetes-side contract: container
ports, Services and endpoints.

Right now the operator has to *guess* which ports a process will actually bind,
because the real values live in a host YAML file it deliberately does not parse
or rewrite (the config file remains the site's source of truth, mounted
read-only). If the file and the operator disagree, the Services point at the
wrong ports and nothing works — the process binds what the file says, while
Kubernetes advertises something else.

Concretely, on one of our multi-NUMA hosts the layout is:

| instance | gRPC | HTTP |
|---|---|---|
| numa0 | `[::1]:8080` | `[::1]:8081` |
| numa1 | `[::1]:8082` | `[::1]:8083` |

[...]

PR #2157, opened 2026-08-20
feat(xcfg): override config values from the environment

borislitv:
A config value could only come from the file, so deploying the same image to several hosts meant shipping a different file to each one just to change an endpoint or an instance id.

Add WithEnv, which lets an environment variable override the value a key holds in the document. A variable is named after the path its key sits at, prefixed with YANET_, upper-cased, with every separator replaced by an underscore: server.endpoint reads YANET_SERVER_ENDPOINT and gateways[0].endpoint reads YANET_GATEWAYS_0_ENDPOINT. WithEnvPrefix serves a binary that must not read the shared namespace.

The override is applied to the YAML document before it is decoded rather than written onto the struct by reflection, so every value still arrives through the destination type's own UnmarshalYAML: an empty NonEmptyString is rejected whether it came from the file or the environment, and a Required set this way still counts as explicitly set. The wrappers report the type they configure through the new envTyped interface, which also decides whether a value is emitted as a quoted scalar -- without it an endpoint such as "[::1]:8080" would be read as a flow sequence.

The variable name is generated from the destination type and never parsed out of the environment, so a key that itself contains an underscore (http_endpoint) cannot be split at the wrong place. Two paths folding onto one name are rejected rather than left silently ambiguous. A map stays file-only, having no fixed key set to generate names from, and a sequence element past the end of the file's list is appended only when every index up to it is supplied, so a gap cannot decode as a null element.

The walk descends into a field only where a variable names a leaf the type actually generates, matched against those names rather than against a string prefix. Two sibling keys where one extends the other are the reason: YANET_GATEWAY_DEVICES_A belongs to gateway_devices, yet it opens with every character of the name gateway generates. Descending on that resemblance would materialise gateway as an empty mapping no leaf ever fills, which decodes as null and wipes the subtree's defaults -- and with it the config, since a null key is rejected. A variable naming nothing in the type, or an index whose element holds no named leaf, is likewise left alone.

### DIFF
--- a/common/go/xcfg/env.go
+++ b/common/go/xcfg/env.go
@@ -1,0 +1,1099 @@
+package xcfg
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"slices"
+	"strconv"
+	"strings"
+	"time"
+	"unicode"
+
+	"gopkg.in/yaml.v3"
+)
+
+// DefaultEnvPrefix is the prefix every environment override carries.
+const DefaultEnvPrefix = "YANET_"
+
+// maxEnvDepth bounds the type walk so a self-referential config type cannot
+// recurse forever while names are generated.
+const maxEnvDepth = 32
+
+var durationType = reflect.TypeFor[time.Duration]()
+
+// envTyped is implemented by a wrapper that stores its value in an unexported
+// field, such as NonEmptyString or NonZero[T], so the env walk can see the
+// type actually being configured instead of stopping at the wrapper.
+//
+// The wrapped type decides two things the wrapper itself cannot answer:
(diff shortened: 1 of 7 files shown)


borislitv:
https://github.com/yanet-platform/yanet2/issues/2151

borislitv:
@codex review

borislitv:
@codex review

borislitv:
@codex review

borislitv:
@codex review

borislitv:
@codex review

borislitv:
@codex review

borislitv:
@codex review

3Hren (approved):

PR #2158, opened 2026-08-20
fix(controlplane,operator): default backends to ephemeral ports

### LINKED ISSUE #2156: Default every self-registering listener to port `0`
### Summary

Components that already advertise their real listen address to the gateway
still ship with hardcoded ports in their default configs. Since the port they
announce is taken from the listener rather than from the file, the number in
the file buys nothing — it only creates a collision to manage.

We would like `:0` to become the shipped default everywhere self-registration
makes the actual port irrelevant.

### What already works this way

The controlplane modules are the precedent. `controlplane.d/default.yaml`
gives every one of the 12 modules `endpoint: "[::1]:0"`, and the service
runner registers whatever the kernel handed out:

```go
// controlplane/gateway/runner.go:133
if err = m.register(ctx, listener.Addr()); err != nil {
```

We verified this on a production host: the control plane holds a dozen
kernel-assigned ports while the gateway stays on its configured one, and
everything resolves through the registry.

### What does not, but could

The standalone operators use exactly the same mechanism — they bind, then
register the listener's own address:

```go
// common/go/operator/operator.go:126,145-148
listener, err := m.makeListener()
...
runner := NewGatewayRegRunner(
    ...
    listener.Addr(),
```

```go
// common/go/operator/operator.go:174
return net.Listen("tcp", m.endpoint)
```

Yet each one ships a fixed port:

| operator | shipped YAML | Go default |
|---|---|---|
| pipeline | `[::1]:50001` | `localhost:50001` |
| route | `[::1]:50002` | `localhost:50002` |
| forward | `[::1]:50003` | `[::1]:50003` |
| decap | `[::1]:50004` | `[::1]:50004` |

Nothing dials these numbers directly — clients reach the operators through the
gateway registry, which knows the address from `listener.Addr()`. As far as we
can tell the ports are pure convention, and `50001..50004` is a range someone
has to keep track of when several instances share a host.

Two smaller inconsistencies visible in the same table: the Go default and the
packaged YAML disagree on the host for pipeline and route (`localhost` vs
`[::1]`), which resolve differently on a dual-stack host.

### What we propose

Make `[::1]:0` the default for the four operators, in both
`operators/*/internal/operator/cfg.go` and the packaged
`operators/*/etc/yanet/*-default.yaml`, matching what the modules already do.
An installation that wants a fixed port keeps setting one in its file; nothing
is removed.

If you would rather keep the packaged YAML explicit and only change the Go
defaults (or the reverse), that works for us too — the important part is that
the out-of-the-box configuration does not pin a port that nothing needs.

### What must stay pinned

For completeness, the places where `:0` is *not* applicable, because the
address is the entry point and no self-registration exists:

- the gateway itself (`gateway.server.endpoint` / `http_endpoint`) — every
  client and every module's `gateway_endpoint` is configured with it;
  `DefaultConfig()` pins `[::1]:8080` at `controlplane/gateway/cfg.go:86`;
- `bird-adapter`'s own `listen_addr` (`localhost:50051`) — BIRD connects to it;
  its `route_operator_endpoint: "[::1]:8080"` is a dial target, not a listener,
  and would follow whatever the gateway ends up on.

### Why this matters to us

We are moving YANET 2 firewall hosts to a Kubernetes operator that owns the
port layout. Every listener that picks its own port is one less number the
orchestrator has to reserve, reconcile and keep in sync with a file it does not
own. With the operators on `:0`, the gateway pair is the only fixed layout
left — which is exactly what the companion request about gateway overrides is
for.

### Notes

- Happy to send a PR for the default change if the direction is acceptable.
- If any of the four operator ports is in fact dialled directly somewhere we
  have not found, pointing us at it would be useful — that is the one thing
  that would rule this out.

3Hren:
- Default self-registering operator listeners to kernel-assigned loopback ports.
- Remove redundant backend endpoint keys from shipped control-plane and operator configurations.
- Verify omitted module, device, and operator endpoints inherit `[::1]:0`.

Closes #2156.

### DIFF
--- a/controlplane/etc/yanet/controlplane.d/default.yaml
+++ b/controlplane/etc/yanet/controlplane.d/default.yaml
@@ -64,60 +64,48 @@ modules:
     memory_path: *memory_path
     # Memory requirements for a single FIB-push transaction.
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   route-mpls:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   decap:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   dscp:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   forward:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   mirror:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   nat64:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   pdump:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 16MB
-    endpoint: "[::1]:0"
   acl:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 8GB
-    endpoint: "[::1]:0"
   blackhole:
     instance_id: 0
     memory_path: *memory_path
     memory_requirements: 4MB
-    endpoint: "[::1]:0"
 devices:
   plain:
     instance_id: 0
     memory_path: *memory_path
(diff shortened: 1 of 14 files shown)


3Hren:
@codex review

PR #2164, opened 2026-08-20
fix(cli,common): silence clippy 1.98 drift

GeorgyKirichenko:
- The CLI error type carries its payload behind a boxed inner struct, shrinking the value from 176 bytes to one pointer, so clippy 1.98's result_large_err stops failing every workspace build that returns the error by value.
- Public constructors and methods keep their exact signatures; field-style access keeps compiling through Deref, and a unit test pins the two-word size so the lint cannot silently return.


3Hren:
Pinned 1.97 instead.

Issue #2083, opened 2026-08-16
Decouple the route operator from controlplane/ffi


OrIX219:
**Motivation**

`yanet-route-operator` talks to the gateway over gRPC and never touches shared
memory, yet it links the whole cgo shared-memory stack: `libagent`,
`libconfig_cp`, `libconfig_dp`, `libcounters`, every module's `*_cp` archive,
and after https://github.com/yanet-platform/yanet2/pull/2074 `libcounter_pattern` and `librure`.

The entire cause is one line in
`operators/route/internal/discovery/neigh/entry.go`:

```go
type HardwareRoute = route.HardwareRoute
```

That imports modules/route/controlplane, which imports controlplane/ffi.
The operator references nothing else from it — grep 'ffi\.' across
operators/route/ returns nothing. #cgo LDFLAGS is package-level metadata,
so the flags reach the binary whether or not a symbol is used, and there is no
"import for types only".

Consequences:

- A cargo build sits on the operator's critical path, for an engine it never calls.
- The link needs hand-maintained depends: entries in meson, because a cgo link
line is an opaque string to it. Missing one is silent until ninja happens to
schedule the Go link first, which is how build-release-artifacts failed with
cannot find -lrure during #2074.

Possible solution

Move HardwareRoute — the struct plus String() and Compare(), all
stdlib-only — out of modules/route/controlplane/backend.go into a leaf
package such as modules/route/controlplane/hwroute. Leave
type HardwareRoute = hwroute.HardwareRoute behind so the module's public API
and its internal uses are untouched, and point neigh/entry.go at the leaf
package. newHardwareRoute stays put, since it takes a routepb.FIBNexthop.

Three files and one new package. Prototyped and verified.

Definition of done

- go list -deps ./operators/route/cmd/yanet-route-operator contains neither
controlplane/ffi nor modules/route/bindings/go/croute.
- CGO_ENABLED=0 go build ./operators/route/cmd/yanet-route-operator succeeds.
- The lib_counter_pattern / librure_targets entry in
operators/route/meson.build is removed along with its comment.
- go test ./modules/route/... ./operators/route/... passes.

Alternatives

- Keep mirroring cgo flags into meson depends: by hand. This is the status
quo, and it breaks silently every time someone adds a flag.
- Duplicate the struct in the operator. Cheap, but two definitions to drift.

PR #2167, opened 2026-08-21
build(operators): disable cgo for the grpc-only operators

3Hren:
- Follow-up to #2165 (#2083): the route operator is pure Go again, but nothing asserted it, so a stray import of a module control plane or of the FFI bindings would silently restore the cgo link that the removed meson `depends:` used to paper over, and surface only as the ordering-dependent `cannot find -lrure` failure seen in #2074.
- Build the gRPC-only operators (`decap`, `pipeline`, `route`) with `CGO_ENABLED=0` through a shared `yanet_go_nocgo_env` in the root `meson.build`, so every regular build (release, deb, sanitizer, functional) is the guard and no extra CI job is needed.
- `forward` and `bird-adapter` are untouched: they still reach `bindings/go/filter` through `common/filterpb/v1`.
- The three binaries become statically linked pure-Go executables (`go version -m` reports `CGO_ENABLED=0`); the Go resolver replaces the libc one, which is the usual shape for Go daemons.
- Verified that the guard trips: re-importing `modules/route/controlplane` into the route operator makes `meson compile yanet-route-operator` fail with `controlplane/ffi: build constraints exclude all Go files`.
- The fuzzing-skip comments of the route and pipeline operator targets are refreshed: with cgo disabled the sanitizer flags no longer reach these builds, so the skip is kept only because the fuzzing build has no use for the binaries.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -134,6 +134,16 @@ if get_option('fuzzing').enabled()
   yanet_cgo_ldflags += cgo_flags
 endif
 
+# gRPC-only operators never touch shared memory and link no cgo.
+#
+# Building them with cgo disabled turns a stray import of a module control
+# plane or of the FFI bindings into a loud build failure; with cgo on it
+# would silently restore the cgo link, which meson cannot see and which
+# then fails only as an ordering-dependent "cannot find -l<lib>" error.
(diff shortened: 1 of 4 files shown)


3Hren:
@codex review

PR #2190, opened 2026-08-21
perf(pipeline): speed up chain and pipeline hash demux

GeorgyKirichenko:
- Pad the weighted chain and pipeline routing tables to a power of two so per-packet selection masks the flow hash instead of dividing it; the padding is spread by largest-remainder allocation so configured weight ratios survive the rounding and zero-weight destinations stay unroutable.
- Build each destination's packet list through locally cached tails during the demux pass and splice it once afterwards, instead of reloading the shared list state and tallies for every packet; destinations past a small cache append directly.
- Prefetch the next packet descriptor while walking the demux source list.
- The chain table now stores destination indices, matching the pipeline table.
- Extend the dispatch regression suite with deterministic tests for the padded weights, the zero-weight isolation under padding, and the direct-append fallback beyond the tail cache, on both demux levels.

### DIFF
--- a/lib/controlplane/config/econtext.c
+++ b/lib/controlplane/config/econtext.c
@@ -2,9 +2,79 @@
 
 #include <string.h>
 
+#include "common/numutils.h"
+
 // cp_config and cp_config_gen
 #include "lib/controlplane/config/zone.h"
 #include "lib/errors/errors.h"
+
+// Minimum power-of-two routing table size once any destination is enabled.
+//
+// A table barely larger than the weight sum rounds small ratios coarsely
+// (three equal chains need four slots), while the floor keeps configured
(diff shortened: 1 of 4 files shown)


GeorgyKirichenko:
@chatgpt-codex-connector review

Review findings addressed in deb3b590:

- Ratio distortion (both P1s): the round-robin padding is replaced by largest-remainder allocation (`routing_table_pad`), which spreads the power-of-two padding proportionally to the configured weights, plus a 64-slot minimum table size so small weight sums round negligibly (1:2 now splits 32.8/67.2 instead of 50/50; 100:1 splits 99.2/0.8 instead of 89/11). Zero-weight destinations still receive no slots. Products are computed at 128 bits so large weights cannot overflow the share arithmetic.

PR #2196, opened 2026-08-24
refactor(commonpb): colocate metric json encoding

3Hren:
- Keep histogram JSON serialization next to the metric conversion helpers.
- Consolidate the JSON regression coverage with the metric tests.
- Remove the stale test-package allowlist entry after deleting the standalone test file.


3Hren:
@codex review

PR #2213, opened 2026-08-24
fix(acl): restore generic metrics

### LINKED ISSUE #1746: fix(acl): stop exporting generic per-module counters as acl_rule_* metrics
## Motivation

Every module gets `rx`, `tx`, `drop`, `pending_input`, `pending_output`, and `hist_0`..`hist_5` registered for it generically by the C control-plane module init (`lib/controlplane/config/cp_function.c`/`cp_pipeline.c` register `drop`/`pending_input`/`pending_output` the same way for every module).

`modules/acl/controlplane/metrics.go`'s `collectDataplaneMetrics` switches on `counter.Name` against a fixed `aclStructuralCounters` list (`acl_no_match`, `acl_action_*`, `acl_state_miss`, `acl_sync_sent`, confirmed at lines 47-51) with a `case` per name (lines 139-183) but no case for any generic counter, so all of them fall into the `default` arm (lines 184-191) and get exported as `acl_rule_packets`/`acl_rule_bytes` labelled `counter=<name>` — a scrape reports `rx`/`tx` as if they were ACL rules. The `hist_0`..`hist_5` counters are worse: they are multi-bucket, not a scalar packet/byte pair, so exporting them through that arm reports a bucket value as a packet count. Confirmed unchanged on current `main`.

`modules/fwstate/controlplane`'s `emitCounterMetrics` already handles both cases correctly — dedicated `fwstate_rx_packets`/`fwstate_tx_packets`/etc. names for the generic counters, with the histogram counters skipped — confirming the correct shape already exists in-repo one module over.

This is live by default, not just under a specific query: with no `counter` tag on the request, ACL's `metrics.Query` call resolves through the unrestricted branch, and `DPConfig.ModuleCounters` treats a nil/empty counter query as "return all counters" (`controlplane/ffi/shm.go:611-613`) — so every generic and histogram counter is already read and mis-exported on the module's default, untagged Prometheus scrape today.

## Acceptance

- [ ] ACL's `collectDataplaneMetrics` switch has explicit arms for `rx`/`tx`/`drop`/`pending_input`/`pending_output`, giving them dedicated metric names, mirroring fwstate's `emitCounterMetrics`, instead of falling into the `acl_rule_*` default.
- [ ] The `hist_0`..`hist_5` histogram counters are either skipped (mirroring fwstate) or exported through a shape that reflects that they are multi-bucket — never through the `acl_rule_*` packet/byte pair.
- [ ] A regression test pins both: the generic counters get dedicated names, and a histogram counter never surfaces as `acl_rule_*`.

Source: local task tracker migration, item dated 2026-08-06, verified against current `main` on 2026-08-06.

semenovad:
ACL metrics previously exposed only ACL-specific structural counters such as `acl_no_match` and `acl_action_*`. This change adds dedicated packet/byte metrics for `rx`, `tx`, `drop`, `pending_input`, and `pending_output`, while keeping histogram and per-rule counters excluded

### DIFF
--- a/modules/acl/controlplane/metrics.go
+++ b/modules/acl/controlplane/metrics.go
@@ -14,6 +14,10 @@ type metricsSource interface {
 	Metrics(tags ...*commonpb.MetricTag) ([]*commonpb.Metric, error)
 }
 
+type moduleCounterReader interface {
+	ModuleCounters(dataplaneConfig *ffi.DPConfig, position ffi.ModuleReference, counterNames []string) []ffi.CounterInfo
+}
+
 // MetricsService exposes ACL module metrics over its own gRPC service.
 type MetricsService struct {
 	aclpb.UnimplementedMetricsServiceServer
@@ -46,7 +50,8 @@ func (m *MetricsService) GetMetrics(ctx context.Context, req *commonpb.GetMetric
 var aclStructuralCounters = []string{
 	"acl_no_match", "acl_action_allow", "acl_action_deny", "acl_action_count",
 	"acl_action_check_state", "acl_action_create_state", "acl_action_unknown",
-	"acl_state_miss", "acl_sync_sent",
+	"acl_state_miss", "acl_sync_sent", "rx", "tx", "drop", "pending_input",
+	"pending_output",
 }
 
 // Metrics returns ACL module metrics matching tags: per-pipeline packet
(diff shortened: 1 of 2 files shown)


semenovad:
@codex review

GeorgyKirichenko (approved):

Issue #1743, opened 2026-08-06
fix(acl): stop grpcmetrics double-counting fwstate RPCs on the shared server


3Hren:
## Motivation

`controlplane/gateway/runner.go` builds one `*grpc.Server` per module process and chains all of that module's `UnaryServerInterceptors()` onto it. `modules/acl/controlplane/mod.go` registers three services on that single server — `ACLService`, `MetricsService`, and `FWStateService` (`fwstate.NewFWStateService` is constructed and hosted inside the ACL module process).

`ACLModule.UnaryServerInterceptors()` returns exactly one interceptor, built from `grpcmetrics.NewFactory(grpcmetrics.WithLabeler(labeler))` (`modules/acl/controlplane/mod.go:81-83`), with no `grpcmetrics.WithServiceFilter`. Since the interceptor runs for every RPC on the shared server regardless of which service it targets, ACL's own `grpcmetrics.ServerMetrics` records `FWStateService` RPCs as if they were ACL's.

`common/go/grpcmetrics/options.go`'s `WithServiceFilter` exists for exactly this scenario (its doc comment calls out transparent-proxy/multi-service servers), and `controlplane/gateway/gateway.go` already uses it correctly at the gateway layer. Fwstate's own side of this gap is already fixed: `fwstate.NewMetricsFactory()` (`modules/fwstate/controlplane/service.go:65`) scopes its collector with `WithServiceFilter` to the two fwstate service FQNs. ACL's own factory call remains unscoped — confirmed still present on current `main` at `modules/acl/controlplane/mod.go:81-83`.

## Scope

- Only ACL's own `grpcmetrics.NewFactory` call in `modules/acl/controlplane/mod.go`. Fwstate's half of this class of bug is already fixed and out of scope.

## Acceptance

- [ ] ACL's `grpcmetrics.NewFactory(...)` call is scoped with `WithServiceFilter` to only the ACL/Metrics service FQNs it actually owns, mirroring the pattern already used by `fwstate.NewMetricsFactory()`.
- [ ] A regression test exercises RPCs across both services sharing the one server and shows ACL's `ServerMetrics` no longer records `FWStateService`/fwstate `MetricsService` RPCs.

Source: local task tracker migration, item dated 2026-07-09 (updated 2026-07-14), verified against current `main` on 2026-08-06.

saushew:
[...]

PR #2215, opened 2026-08-24
fix(acl): scope grpc metrics to own services

saushew:
- Introduce acl.NewMetricsFactory() pre-binding the module's labeler and a service filter admitting only the ACL and ACL Metrics service names, mirroring the fwstate module's factory pattern.
- Derive the service-name constants from the generated descriptors and use them in the module wiring, replacing the hand-written FQN string.
- Cover the scoping with a regression test asserting foreign services sharing the module server record no series.

Refs #1743.


saushew:
@codex review

PR #2217, opened 2026-08-24
feat(commonpb): add a fixed-width ipv6 address message

### LINKED ISSUE #2199: feat(commonpb): add a fixed-width ipv6 address message
## Motivation

`commonpb.IPAddress` represents IPv6 as a 16-byte sub-slice whose length is also the family discriminator ([ipaddr.proto#L13-L21](https://github.com/yanet-platform/yanet2/blob/4a40b2325ab3f4d0bb2ef8f327a198588fed4ae9/common/commonpb/v1/ipaddr.proto#L13-L21)). IPv6-only consumer fields cannot state their family structurally, every Go decode allocates the byte slice, and the bi-contiguous network work (#2183) has no natural per-half representation to build on. A fixed-width two-half message makes every representable value a valid IPv6 address.

## Outcome

A `commonpb.IPv6Address` message with `fixed64 hi` (address bytes 0-7 as a big-endian u64) and `fixed64 lo` (bytes 8-15), plus Go/Rust helpers and string-formatted JSON. Wire size matches today's 16-byte payload (18 bytes either way); `fixed64` rather than `uint64` because global-unicast `hi` values have high bits set and would cost 10 varint bytes each.

## Scope

- Add the message to `common/commonpb/v1/ipaddr.proto` with the byte-order convention documented on the fields.
- Go helpers: `netip.Addr` conversions accepting only IPv6, `AsLogValue`, JSON string form. Zone-bearing addresses are rejected explicitly: a zone is host-local and belongs as a sibling field on the owning message (see `NeighbourEntry.device`), and the current `NewIPAddressFromAddr` silently strips it.
- Rust: conversions against `std::net::Ipv6Addr` and `u128`, `Display`, serde string form.
- Cross-language shared test vectors with asymmetric halves (for example `2a02:6b8:0:1::100`) so an hi/lo swap or endianness defect cannot pass in either language.

## Acceptance

- [ ] Every (hi, lo) pair decodes as a valid IPv6 address; the message has no malformed-length state.
- [ ] `::`, `::1`, an IPv4-mapped address, and the asymmetric vectors round-trip wire and JSON identically in Go and Rust.
- [ ] A deliberately half-swapped fixture fails tests in both languages.
- [ ] Zoned `netip.Addr` input returns an error rather than being silently stripped.
- [ ] `buf lint` and generated-artifact checks pass; no existing message or consumer changes.

3Hren:
- Adds an IPv6Address message carrying the address as two fixed64 halves in network byte order, so every value is a valid address with no malformed-length state.
- Adds Go helpers mirroring the IPv4 sibling: netip conversions that accept the IPv4-mapped form but reject native IPv4 and zoned input, a total conversion back, compact log rendering, and bare-string JSON.
- Adds Rust conversions, Display that keeps the IPv4-mapped rendering, FromStr, and hand-written string-form serde excluded from the derive list.
- Pins the half-split and byte-order contract with shared cross-language fixtures, including a golden wire-bytes test.

Closes #2199.

### DIFF
--- a/common/commonpb/v1/ipaddr.proto
+++ b/common/commonpb/v1/ipaddr.proto
@@ -28,3 +28,20 @@ message IPv4Address {
   // The address as a big-endian u32 value: 10.1.2.3 is 0x0A010203.
   fixed32 addr = 1;
 }
+
+// IPv6Address represents an IPv6 host address.
+//
+// Unlike IPAddress, the family is fixed by the type and every value is a
+// valid address: there is no malformed state to validate after decoding.
+// IPv4-mapped addresses belong to this family, distinct from the same
+// address in its IPv4Address form.
+message IPv6Address {
+  // Address bytes 0-7 as a big-endian u64.
+  //
+  // For 2a02:6b8:0:1::100 this is 0x2a0206b800000001.
+  fixed64 hi = 1;
+  // Address bytes 8-15 as a big-endian u64.
+  //
+  // For 2a02:6b8:0:1::100 this is 0x0000000000000100.
+  fixed64 lo = 2;
+}
--- a/common/commonpb/v1/ipv6addr.go
+++ b/common/commonpb/v1/ipv6addr.go
@@ -1,0 +1,81 @@
+package commonpb
+
+import (
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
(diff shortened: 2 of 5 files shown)


3Hren:
@codex review

PR #2227, opened 2026-08-24
refactor(filterpbconv): remove unused prefix converters

3Hren:
- Remove unused protobuf prefix conversion helpers from the filter bindings bridge.
- Keep the live standard-library prefix converters used by route MPLS unchanged.


3Hren:
@codex review

PR #2228, opened 2026-08-24
fix(controlplane): derive execution context release length from its allocation

### LINKED ISSUE #1973: fix(controlplane): config_gen_ectx error path frees a different size class than it allocated
## Motivation

[`config_gen_ectx_create`](https://github.com/yanet-platform/yanet2/blob/2436e2b6effe62722a7e9a4a2f61b881d22ae47a/lib/controlplane/config/econtext.c#L1710-L1865) sizes its allocation by the device registry's capacity:

```
size_t ectx_size = sizeof(struct config_gen_ectx) +
                    sizeof(struct device_ectx *) *
                        cp_device_registry_capacity(&cp_config_gen->device_registry);
```

[`config_gen_ectx_free`](https://github.com/yanet-platform/yanet2/blob/2436e2b6effe62722a7e9a4a2f61b881d22ae47a/lib/controlplane/config/econtext.c#L1356-L1411) instead sizes the matching `memory_bfree` by `config_gen_ectx->device_count`:

```
size_t ectx_size =
        sizeof(struct config_gen_ectx) +
        sizeof(struct device_ectx *) * config_gen_ectx->device_count;
```

Those two expressions agree only after `device_count` has been assigned, which happens well into the constructor. Two fallible steps run before that assignment: the counter-storage registry [allocation](https://github.com/yanet-platform/yanet2/blob/2436e2b6effe62722a7e9a4a2f61b881d22ae47a/lib/controlplane/config/econtext.c#L1741-L1750) from `cp_config->memory_context`, and its [init](https://github.com/yanet-platform/yanet2/blob/2436e2b6effe62722a7e9a4a2f61b881d22ae47a/lib/controlplane/config/econtext.c#L1752-L1762). Either failing jumps to `goto error`, which calls `config_gen_ectx_free` while `device_count` is still zero (the block was `memset` to zero right after allocation), so the block is freed as a bare `sizeof(struct config_gen_ectx)` although it was allocated with room for the flexible array.

The [comment](https://github.com/yanet-platform/yanet2/blob/2436e2b6effe62722a7e9a4a2f61b881d22ae47a/lib/controlplane/config/econtext.c#L1765-L1766) sitting immediately above the `device_count` assignment claims it is set "before any fallible work so the error path's config_gen_ectx_free computes the correct flexible-array size". That claim is false as written — the registry allocation and its init both precede the assignment — and it is exactly what makes the defect easy to miss on review.

Consequence: the block allocator (`common/memory*.h`) is a size-class free-list allocator, so returning a block to a smaller size class than it was taken from corrupts the arena's free-list accounting. A measured instance during unrelated dynamic analysis showed 272 bytes allocated against 208 bytes freed for this block.

This is pre-existing on the current default branch and is not introduced by any in-flight change. It was surfaced by an allocation ledger while validating a separate control-plane lifetime fix, and is the same allocation-hygiene family as #1969 and #1930 (both `counter_storage_spawn` error-path defects) — independent of them and can land on its own.

## Scope

- Free with the same size expression the allocation used, or assign `device_count` (and `object_count`, which has the same ordering relative to the object array size, though its error path frees via a separately-tracked pointer/count pair rather than baking the count into the block size) before the first fallible step so the existing `config_gen_ectx_free` expression is always true.
- Correct the comment above the `device_count` assignment, whichever way the fix goes.

## Acceptance

- [ ] The allocation and the matching free derive their size from the same expression on every path through `config_gen_ectx_create`, including every `goto error`.
- [ ] A regression test drives the constructor's early failure paths (registry allocation failure, registry init failure) and asserts the arena's free accounting returns to its pre-call state. Note this repository's rule that behaviour reachable from Go through cgo is tested in Go, not C.

Source: surfaced by an allocation-ledger run during dynamic validation of an unrelated control-plane module-lifetime fix; confirmed pre-existing by reading the current default branch at [lib/controlplane/config/econtext.c](https://github.com/yanet-platform/yanet2/blob/2436e2b6effe62722a7e9a4a2f61b881d22ae47a/lib/controlplane/config/econtext.c#L1710-L1865). Same allocation-hygiene family as #1969 and #1930.

OrIX219:
- Read the device registry capacity once and record it before the first fallible step, so a construction that fails early releases the block against the same length it was allocated with instead of a shorter one, which filed it under a smaller block size class and corrupted the arena's free-list accounting.
- Record the object count alongside it and bound the device loop by the same value the trailing array was sized with.

### DIFF
--- a/lib/controlplane/config/econtext.c
+++ b/lib/controlplane/config/econtext.c
@@ -1833,11 +1833,11 @@ config_gen_ectx_create(
 	struct cp_config *cp_config = ADDR_OF(&cp_config_gen->cp_config);
 	struct memory_context *memory_context = &cp_config->ectx_memory_context;
 
+	uint64_t device_count =
+		cp_device_registry_capacity(&cp_config_gen->device_registry);
+
 	size_t ectx_size = sizeof(struct config_gen_ectx) +
-			   sizeof(struct device_ectx *) *
-				   cp_device_registry_capacity(
-					   &cp_config_gen->device_registry
-				   );
+			   sizeof(struct device_ectx *) * device_count;
 
 	struct config_gen_ectx *config_gen_ectx = (struct config_gen_ectx *)
 		memory_balloc(memory_context, ectx_size);
@@ -1853,6 +1853,10 @@ config_gen_ectx_create(
 	packet_front_init(&config_gen_ectx->packet_front);
 
 	SET_OFFSET_OF(&config_gen_ectx->cp_config_gen, cp_config_gen);
+
+	config_gen_ectx->device_count = device_count;
+	config_gen_ectx->object_count =
+		cp_object_registry_capacity(&cp_config_gen->object_registry);
 
 	struct cp_config_counter_storage_registry *registry =
 		(struct cp_config_counter_storage_registry *)memory_balloc(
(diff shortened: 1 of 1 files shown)


OrIX219:
@yanet-platform/yanet-core ship pls

GeorgyKirichenko (approved):

Issue #2197, opened 2026-08-24
feat(commonpb,filterpb)!: family-typed ip address and network wire model


3Hren:
## Goal

Replace the length-discriminated `bytes` IP representation on the public protobuf surface with family-typed messages: fixed-width `IPv4Address`/`IPv6Address`, family-typed contiguous network messages, a bi-contiguous IPv6 network message, and `oneof` containers for the genuinely mixed-family fields. Family and mask-class contracts become structural at the wire boundary instead of comment-enforced and re-validated by every consumer.

## Scope

- `common/commonpb/v1` gains the typed address and network messages; `IPAddress` and `ContiguousIPNetwork` become family oneofs. The zone reservation is dropped: a scope/zone is host-local and belongs as a sibling field on the owning message, as `NeighbourEntry.device` already models ([neighbour.proto#L57-L66](https://github.com/yanet-platform/yanet2/blob/4a40b2325ab3f4d0bb2ef8f327a198588fed4ae9/operators/route/operatorpb/v1/neighbour.proto#L57-L66)).
- Family-fixed module fields (nat64, fwstate/acl state sync, bird-adapter MPLS sources) move to the typed address messages; filter network lists (acl, forward, mirror) and prefix lists (dscp, decap) split per family with the mask class each consumer actually accepts; `filterpb.IPNet` and `IPPrefix` are removed once unconsumed.
- JSON: the typed messages serialize as bare strings identically in Go and Rust (decided at #2198: `IPv4Address` renders `"10.1.2.3"` with no object wrapper), ending the per-language object-versus-string split the legacy messages carry. Legacy messages keep their current forms until removed; a field retype from a legacy message therefore changes the JSON shape its web consumer sees, and that TS update rides the same migration PR.
- Wire breaks are accepted for the migration: each breaking PR adds its proto file to `buf.yaml` `breaking.ignore` in the same change, following the existing acl/fwstate cutover precedent ([buf.yaml#L24-L34](https://github.com/yanet-platform/yanet2/blob/4a40b2325ab3f4d0bb2ef8f327a198588fed4ae9/buf.yaml#L24-L34)), and the entries are removed at the end of the series.

[...]

PR #2229, opened 2026-08-24
refactor(route-mpls,filterpb)!: migrate rule prefixes to the contiguous network container

### LINKED ISSUE #1949: refactor(common): migrate filterpb.IPPrefix to commonpb.ContiguousIPNetwork
## Motivation

[common/filterpb/v1/filter.proto:14-17](https://github.com/yanet-platform/yanet2/blob/aef29727a60898786459bb10e7fabca4c915485d/common/filterpb/v1/filter.proto#L14-L17) defines:

```proto
message IPPrefix {
  bytes addr = 1;
  uint32 length = 2;
}
```

This is a structural duplicate of `common.commonpb.v1.ContiguousIPNetwork`, added in #1908 (merged via #1939) precisely to give callers that need a CIDR prefix a canonical commonpb wire type. `IPPrefix` has exactly one consumer: `Rule.prefix` in [modules/route-mpls/controlplane/routemplspb/v1/routempls.proto:53](https://github.com/yanet-platform/yanet2/blob/aef29727a60898786459bb10e7fabca4c915485d/modules/route-mpls/controlplane/routemplspb/v1/routempls.proto#L53), plus the route-mpls Rust CLI. Of the deferred migration candidates, this is the strongest: it removes a real duplicate type rather than reshaping a field.

## Scope

- Replace `IPPrefix` with `common.commonpb.v1.ContiguousIPNetwork` in `Rule.prefix`.
- Update the route-mpls Rust CLI wherever an `IPPrefix` value is parsed or rendered.
- Remove `IPPrefix` from `filter.proto` once no consumer remains (re-grep before deleting — do not assume the population above is still current).

## Constraints

**This is a WIRE-BREAKING protobuf change**: changing a field's message type is exactly what `buf breaking`'s `PACKAGE` rule (run in this repo's CI) exists to catch. This needs a deliberate compatibility decision before implementation — either a new field number carrying `ContiguousIPNetwork` with the old `prefix` field deprecated, or a coordinated breaking change with every route-mpls deployment. A blind field-type swap will fail CI and, if forced through, silently corrupt wire compatibility with any client still on the old schema.

## Acceptance

- [ ] `Rule.prefix` (or a new field) carries `common.commonpb.v1.ContiguousIPNetwork`, with the compatibility strategy documented in the PR.
- [ ] `buf breaking` passes.
- [ ] The route-mpls Rust CLI is updated to match.
- [ ] `IPPrefix` is removed from `filter.proto` if no other consumer remains.

Source: deferred from #1908's work (merged as PR #1939), which added `ContiguousIPNetwork` but deliberately left existing module-specific prefix messages unmigrated.

3Hren:
- Retypes `Rule.prefix` in place to `common.commonpb.v1.ContiguousIPNetwork`, keeping the field number, and removes the now-unconsumed `filterpb.IPPrefix` message.
- Decodes prefixes through the container's `ToPrefix`, which rejects a missing address and an out-of-range prefix length and masks host bits, with tests for both invalid shapes.
- Makes `ToPBMPLSRoute` return an error for a prefix the structured message cannot represent, logged and skipped by the BIRD import loop the same way as `ToPBRoute`.
- Replaces the route-mpls CLI's hand-rolled prefix conversion with the commonpb `From` impl, so structured output renders the prefix as a bare CIDR string.
- The wire break is accepted under #2197 and follows the #2220/#2221 precedent: `buf breaking` flags the field retype and the message removal, and no `breaking.ignore` entries are added.
- Closes #1949.

### DIFF
--- a/common/filterpb/v1/filter.proto
+++ b/common/filterpb/v1/filter.proto
@@ -11,11 +11,6 @@ message VlanRange {
   uint32 to = 2;
 }
 
-message IPPrefix {
-  bytes addr = 1;
-  uint32 length = 2;
-}
-
 message IPNet {
   bytes addr = 1;
   bytes mask = 2;
--- a/common/rust/filterpb/build.rs
+++ b/common/rust/filterpb/build.rs
@@ -1,13 +1,11 @@
 use core::error::Error;
 
 pub fn main() -> Result<(), Box<dyn Error>> {
-    let serialize = "#[derive(serde::Serialize)]";
     let serialize_deserialize = "#[derive(serde::Serialize, serde::Deserialize)]";
 
     tonic_build::configure()
         .emit_rerun_if_changed(false)
         .build_server(false)
-        .message_attribute(".common.filterpb.v1.IPPrefix", serialize)
         .message_attribute(".common.filterpb.v1.Device", serialize_deserialize)
         .message_attribute(".common.filterpb.v1.PortRange", serialize_deserialize)
         .message_attribute(".common.filterpb.v1.ProtoRange", serialize_deserialize)
--- a/modules/route-mpls/cli/build.rs
+++ b/modules/route-mpls/cli/build.rs
@@ -1,7 +1,7 @@
 use core::error::Error;
 
 pub fn main() -> Result<(), Box<dyn Error>> {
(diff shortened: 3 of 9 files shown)


3Hren:
@codex review

3Hren:
@codex review

PR #2231, opened 2026-08-24
fix(pdump): store the prepared config only once publishing it succeeds

### LINKED ISSUE #2230: fix(pdump): failed publish leaves the stored config pointing at freed shared memory
## Motivation

A pdump config update stores the new config in the service map *before* publishing it, and never restores the previous entry when the publish fails. By the time the publish is attempted, the ring inside that stored config has been rewritten to address the shared memory of the module config being published — memory that the failure path then frees. The map is left holding a config whose ring points into freed shared memory, and a later read of the same config dereferences it.

The window is currently narrow: a publish only fails when the shared-memory write itself fails. It widens as soon as a publish can fail for an ordinary reason such as a cancelled request, which is what #1680 introduces.

## Evidence

Code-proven by reading the update path at 06a17efb5583150792660c9f03803d1cbbef9b79; not yet reproduced at runtime.

- The new config is committed to the map, then published, with no restore on the failure path: [service.go#L403-L410](https://github.com/yanet-platform/yanet2/blob/06a17efb5583150792660c9f03803d1cbbef9b79/modules/pdump/controlplane/service.go#L403-L410).
- The publish reads that same entry back out of the map and hands its ring to the parameter transfer: [service.go#L279-L291](https://github.com/yanet-platform/yanet2/blob/06a17efb5583150792660c9f03803d1cbbef9b79/modules/pdump/controlplane/service.go#L279-L291).
- The ring setup rewrites the caller's ring in place, filling it with addresses inside the shared memory owned by the module config under construction: [ffi.go#L232-L265](https://github.com/yanet-platform/yanet2/blob/06a17efb5583150792660c9f03803d1cbbef9b79/modules/pdump/controlplane/ffi.go#L232-L265).
- Both failure paths free that shared memory and return, leaving the rewritten ring in the map: [service.go#L286-L296](https://github.com/yanet-platform/yanet2/blob/06a17efb5583150792660c9f03803d1cbbef9b79/modules/pdump/controlplane/service.go#L286-L296).
- A later read passes the initialization guard, because the ring does have worker areas, and copies those addresses for the reader: [service.go#L379-L391](https://github.com/yanet-platform/yanet2/blob/06a17efb5583150792660c9f03803d1cbbef9b79/modules/pdump/controlplane/service.go#L379-L391).

The previously stored config is a distinct object and stays intact: the prepared config always carries a freshly allocated ring rather than the one already in the map, so restoring the map entry is sufficient to undo the damage ([service.go#L414-L427](https://github.com/yanet-platform/yanet2/blob/06a17efb5583150792660c9f03803d1cbbef9b79/modules/pdump/controlplane/service.go#L414-L427)).

Every other module publishes first and stores afterwards, so this ordering is unique to pdump.

## Scope

- Restore the previously stored config when a publish fails, and remove the entry entirely when there was none.

## Acceptance

- [ ] A failed publish leaves the previously stored config in the map, with its ring unchanged.
- [ ] A failed publish for a name that had no config leaves no entry behind.
- [ ] A test covers both, driving the real update path with a failing publish.

egnees:
- An update stored the prepared config before publishing it and left it there when the publish failed.
- Publishing addresses the prepared ring at the shared memory it allocates and releases that memory on failure, so the stored entry was left holding pointers into freed shared memory that a later read follows.
- Hand the prepared config to the publish rather than routing it through the map, and store it only once that succeeds, so a failed update does not take effect and no rollback is needed.
- The existing reader-drain test covers this: it now expects the previously published config to survive an update whose publish fails.

Closes #2230.

### DIFF
--- a/modules/pdump/controlplane/service.go
+++ b/modules/pdump/controlplane/service.go
@@ -186,8 +186,8 @@ func (m *PdumpService) SetConfig(
 	err := m.updateConfig(
 		name,
 		request,
-		func() error {
-			return m.updateModuleConfig(name)
+		func(config *pdumpConfig) error {
+			return m.updateModuleConfig(name, config)
 		},
 	)
 	if err != nil {
@@ -282,14 +282,13 @@ func (m *PdumpService) transferConfigParameters(
 // the previous ring have stopped and the state lock has been reacquired.
 func (m *PdumpService) updateModuleConfig(
 	name string,
+	modConfig *pdumpConfig,
 ) error {
 	if m.agent == nil {
 		return fmt.Errorf("pdump agent is required")
 	}
 
 	m.log.Debug("update config", zap.String("module", name))
-
-	modConfig := m.configs[name]
 
 	ffiConfig, err := NewModuleConfig(m.agent, name, WithModuleConfigLog(m.log))
 	if err != nil {
@@ -416,7 +415,7 @@ func (m *PdumpService) registerRingReaders(
 func (m *PdumpService) updateConfig(
 	name string,
 	request *pdumppb.SetConfigRequest,
-	publish func() error,
+	publish func(config *pdumpConfig) error,
 ) error {
 	newConfig, err := m.prepareConfig(name, request)
 	if err != nil {
@@ -427,8 +426,12 @@ func (m *PdumpService) updateConfig(
 		name,
 		fmt.Errorf("terminated by config update"),
 		func() error {
+			if err := publish(newConfig); err != nil {
+				return err
+			}
 			m.configs[name] = newConfig
-			return publish()
+
+			return nil
 		},
 	)
 }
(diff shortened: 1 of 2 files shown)


modules/pdump/controlplane/service.go
```diff
@@ -427,8 +427,19 @@ func (m *PdumpService) updateConfig(
 		name,
 		fmt.Errorf("terminated by config update"),
 		func() error {
+			previous, existed := m.configs[name]
 			m.configs[name] = newConfig
```
GeorgyKirichenko: I think map should be updated only in case if publish succeeds whereas rollback logic here is complicated

GeorgyKirichenko (changes requested):

egnees:
@codex review

GeorgyKirichenko (approved):

Issue #2232, opened 2026-08-24
chore(acl)!: drop the legacy mixed-family network lists after the operator migration


3Hren:
## Motivation

#2205 splits ACL rule networks into family-typed lists (`sources4`/`sources6` as `commonpb.ContiguousIPv4Network`, `destinations4`/`destinations6` with `commonpb.BiContiguousIPv6Network` on the IPv6 side) and migrates every in-tree client — the Rust CLI ruleset schema, the web UI write path, and the functional fixtures. The legacy mixed-family `Rule.srcs`/`Rule.dsts` (`filterpb.IPNet`) stay on the wire marked `deprecated`, accepted and merged server-side solely for the external ACL operator, which still produces the legacy schema. As long as those fields exist, acl keeps `filterpb.IPNet` alive, which #2212 wants removed from the schema entirely.

## Closing outcome

The external ACL operator produces only the typed lists in production, and the deprecated `srcs`/`dsts` fields plus their merge path are removed: `aclpb.Rule` carries networks exclusively as family-typed, class-typed lists.

## Scope

- Track the operator-side migration to `sources4`/`sources6`/`destinations4`/`destinations6` (external repository; a precondition for the removal, not deliverable here).
- Remove `srcs`/`dsts` from `aclpb/v1/acl.proto`, the legacy merge in the service decode path, the web read-side merge, and the legacy-schema compatibility tests.
- The removal is wire-breaking for any remaining legacy writer; it follows the #2197 breaking posture, and `acl.proto` already sits in `buf.yaml` `breaking.ignore`.

## Acceptance

- [ ] The ACL operator writes only the typed lists against a production ACL service.
- [ ] `srcs`/`dsts` no longer exist in `aclpb.Rule`; the service decodes only the typed lists.
- [ ] acl no longer references `filterpb.IPNet`, unblocking its removal in #2212.
- [ ] `buf lint` passes.

## Source

Deferred from #2205 by the author's review decision: the legacy fields were kept wire-compatible for the external operator while all in-tree clients moved to the typed lists.

3Hren:
Recording the current plan after today's revert: #2248 jumped ahead of this issue's precondition and removed `filterpb.IPNet`/the legacy `srcs`/`dsts` fields before the external ACL operator had migrated off them, breaking the deployed operator's wire compatibility. It was reverted in #2252, restoring `filterpb.IPNet` and the legacy mixed-family fields.

[...]

PR #2233, opened 2026-08-24
feat(acl): split filter network lists by address family on the wire

### LINKED ISSUE #2205: feat(acl)!: split filter network lists by address family on the wire
## Motivation

ACL rules carry source and destination networks as mixed-family `filterpb.IPNet` lists of unconstrained address/mask byte pairs ([acl.proto#L82-L83](https://github.com/yanet-platform/yanet2/blob/4a40b2325ab3f4d0bb2ef8f327a198588fed4ae9/modules/acl/controlplane/aclpb/v1/acl.proto#L82-L83)). The compiler's accepted classes are narrower — contiguous IPv4 and bi-contiguous IPv6 ([helper.c#L19-L32](https://github.com/yanet-platform/yanet2/blob/4a40b2325ab3f4d0bb2ef8f327a198588fed4ae9/lib/filter/compiler/helper.c#L19-L32)) — so family is rediscovered from byte length and mask shape is revalidated on every decode, with invalid rules rejected only server-side.

## Outcome

ACL rule networks arrive family-split and class-typed: `srcs_v4`/`dsts_v4` as `repeated commonpb.ContiguousIPv4Network` and `srcs_v6`/`dsts_v6` as `repeated commonpb.BiContiguousIPv6Network`. Invalid mask shapes become unrepresentable on the wire; within-family order is preserved. This is a deliberate wire break under the parent epic.

## Scope

- Replace the two `IPNet` list fields in `aclpb/v1/acl.proto` with the four typed lists; rule semantics treat each list as a match set, so no cross-family ordering is lost.
- Decode through family-typed conversion into the filter binding values (coordinating with the typed slices from #2182/#2184), converting per-half prefix lengths to C masks at the FFI boundary.
- Move mask-class validation into the Rust CLI rule parser so a bad rule fails at parse time with a precise error.
- The proto file already sits in `buf.yaml` `breaking.ignore`; extend its TODO to mention the epic.
- Keep show/inspect rendering and functional coverage behaviorally equivalent.

## Acceptance

- [ ] A non-contiguous IPv4 mask or an IPv6 mask with a hole inside either half is unrepresentable on the wire.
- [ ] A valid IPv6 mask with a hole exactly at the `/64` boundary is expressible and reaches the compiler.
- [ ] Rules preserve within-family network order end to end.
- [ ] The CLI rejects invalid masks at parse time; the service no longer length-sniffs family.
- [ ] `buf lint` passes; the breaking gate is handled by the existing ignore entry.

## Depends on

- #2200 — `ContiguousIPv4Network` message.
- #2183 — `BiContiguousIPv6Network` message.

3Hren:
- Adds family-typed, class-typed rule network lists: `sources4`/`destinations4` carry `commonpb.ContiguousIPv4Network` and `sources6`/`destinations6` carry `commonpb.BiContiguousIPv6Network`, so a non-contiguous IPv4 mask or an IPv6 mask with a hole inside a half is unrepresentable on the wire, while a hole exactly at the /64 boundary remains expressible.
- Keeps the legacy `srcs`/`dsts` accepted and merged server-side for external clients, now marked deprecated; #2232 tracks their removal after the ACL operator migrates, so no wire break happens here.
- Migrates every in-tree client to the typed lists: the CLI ruleset YAML schema gains parse-time mask validation, the web UI writes and reads the typed lists, and the functional fixtures move to the new keys.
- Adds generic bulk prefix converters to commonpb and drops the redundant CIDR-parsing constructors in favor of netip conversions, inlining the string parse into the JSON unmarshal path.
- Closes #2205.

### DIFF
--- a/bindings/go/filterpbconv/v1/convert.go
+++ b/bindings/go/filterpbconv/v1/convert.go
@@ -12,6 +12,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/yanet-platform/yanet2/bindings/go/filter"
+	commonpb "github.com/yanet-platform/yanet2/common/commonpb/v1"
 	filterpb "github.com/yanet-platform/yanet2/common/filterpb/v1"
 )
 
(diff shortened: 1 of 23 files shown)


3Hren:
@codex review

3Hren:
@codex review

3Hren:
@codex review

Issue #2234, opened 2026-08-24
feat(commonpb,acl)!: carry filter networks as general mask-typed messages


3Hren:
## Motivation

#2205 (merged as #2233) typed the acl filter lists with class-typed messages: `sources4`/`destinations4` carry `commonpb.ContiguousIPv4Network` and `sources6`/`destinations6` carry `commonpb.BiContiguousIPv6Network`. That bakes the current filter-compiler mask classes ([helper.c#L19-L32](https://github.com/yanet-platform/yanet2/blob/4a40b2325ab3f4d0bb2ef8f327a198588fed4ae9/lib/filter/compiler/helper.c#L19-L32)) into the wire format. The classes are an engine limitation, not a domain property: the legacy `filterpb.IPNet` carried arbitrary masks and the compiler rejected unsupported shapes at decode. With the class in the type, an engine that later accepts more mask shapes needs another wire migration of every filter field; with a general message it is a server-side validation relaxation and the wire does not move.

Decided with the epic author (2026-08-24): prefix-shaped fields keep `addr + prefix_len` messages; filter mask fields move to general `addr + mask` messages whose mask class is validated by the consuming API.

## Outcome

commonpb gains `IPv4Network` (`IPv4Address addr` + `fixed32 mask`) and `IPv6Network` (`IPv6Address addr` + `IPv6Address mask`), serialized as bare strings — the CIDR form when the mask is contiguous, the address/mask form otherwise, following the `BiContiguousIPv6Network` precedent. The four acl typed lists retype to them in place. `BiContiguousIPv6Network` is removed once unconsumed.

## Scope

- Add the two messages with Go/Rust helpers, bare-string JSON identical in both languages, and conversions from `netip`/`xnetip` values.
- Retype `sources4`/`sources6`/`destinations4`/`destinations6` in place (same field numbers): the fields shipped in #2233 with no external consumers — the external operator still writes the deprecated legacy lists (#2232).
- Enforce the mask classes in the acl service decode, converging with the legacy path's `filter.IPNet.MaskIsValid` check; the CLI stays a thin pass-through (decided in review: server-side validation is sufficient).
- Remove `BiContiguousIPv6Network` (message, helpers, tests) once acl is off it.
- #2206 and #2207 adopt the new messages directly.

## Acceptance

- [ ] Any mask is representable on the wire; the current classes are enforced by the acl API with precise errors, and a hole exactly at the `/64` boundary still reaches the compiler.
[...]

PR #2236, opened 2026-08-24
refactor(commonpb)!: rename the contiguous network messages to prefix messages

### LINKED ISSUE #2235: refactor(commonpb)!: rename the contiguous network messages to prefix messages
## Motivation

`ContiguousIPv4Network` and `ContiguousIPv6Network` carry `addr + prefix_len` — the domain concept is a CIDR prefix, and the current name explains a mask-class property instead of naming the thing. With #2234 giving the filter surfaces general `IPv4Network`/`IPv6Network` messages, the prefix-shaped messages read best as `IPv4Prefix`/`IPv6Prefix`, matching the `netip.Prefix` terminology the Go side already speaks.

## Closing outcome

The messages are renamed to `IPv4Prefix`/`IPv6Prefix`, with per-message proto/Go/Rust files and helpers renamed accordingly. A message rename is invisible on the binary wire and in the bare-string JSON, so this is a generated-API break only; `buf breaking` reports it as a delete plus an add.

## Scope

- Rename the two messages and their files (`ipv4network.proto` frees up for #2234's general message), regenerate, and rename the Go/Rust helpers to the prefix vocabulary.
- Best ordered after #2234: the remaining consumers are then commonpb itself and the #2210 container work.
- #2210's family oneof container over these messages naturally takes the name `IPPrefix` (recorded in the epic; the `filterpb.IPPrefix` name was freed by #2229).

## Acceptance

- [ ] `IPv4Prefix`/`IPv6Prefix` replace the contiguous network messages everywhere; no functional change.
- [ ] `buf lint` passes; the breaking gate follows the epic posture.

3Hren:
- Renames `ContiguousIPv4Network`/`ContiguousIPv6Network` to `IPv4Prefix`/`IPv6Prefix`: the messages carry `addr + prefix_len`, so the name states the domain concept instead of a mask-class property, matching the netip vocabulary.
- Renames the per-message proto/Go files and the Go/Rust helpers accordingly; the freed `ipv4network`/`ipv6network` names go to the general mask-typed messages of #2234.
- A message rename is invisible on the binary wire and in the bare-string JSON, so this is a generated-API break only; `buf breaking` reports the expected delete-plus-add pair for the two messages, following the #2197 posture of no ignore entries.
- `BiContiguousIPv6Network` is deliberately untouched: it is removed under #2234 once acl leaves it.
- Closes #2235.

### DIFF
--- a/bindings/go/filterpbconv/v1/convert.go
+++ b/bindings/go/filterpbconv/v1/convert.go
@@ -115,7 +115,7 @@ func ToIPNet(pb *filterpb.IPNet) (filter.IPNet, error) {
 
 // ToNet4sFromContiguous converts family-typed contiguous IPv4 network
 // messages to filter IPNets.
-func ToNet4sFromContiguous(pb []*commonpb.ContiguousIPv4Network) (filter.IPNets, error) {
+func ToNet4sFromContiguous(pb []*commonpb.IPv4Prefix) (filter.IPNets, error) {
 	prefixes, err := commonpb.PrefixesFromNetworks(pb)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "invalid IPv4 network: %v", err)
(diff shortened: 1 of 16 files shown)


3Hren:
@codex review

PR #2238, opened 2026-08-24
feat(commonpb): add a general ipv6 network message

3Hren:
- Adds `IPv6Network`: an IPv6 base address plus an arbitrary bitmask carried as an `IPv6Address`, taking the file name freed by the #2235 rename.
- Unlike `IPv6Prefix` and the bi-contiguous class, any mask bit pattern is representable — a hole inside a 64-bit half included; the consuming API decides which mask classes it accepts.
- Go and Rust conversions go through xnetip/netip network values, normalizing address bits outside the mask; an absent mask is the zero mask, mirroring `IPv4Network`; JSON is a bare string — the CIDR form for a contiguous mask, the address/mask form otherwise.
- Part of #2234; stacked on #2237.


3Hren:
@codex review

PR #2239, opened 2026-08-24
feat(acl)!: carry filter networks as general mask-typed messages

### LINKED ISSUE #2234: feat(commonpb,acl)!: carry filter networks as general mask-typed messages

3Hren:
- Retypes the acl typed lists in place: `sources4`/`destinations4` carry `commonpb.IPv4Network` and `sources6`/`destinations6` carry `commonpb.IPv6Network`; the fields shipped in #2233 with no external consumers, so the retype is coordinated in-tree only.
- The compiler mask classes (contiguous IPv4, bi-contiguous IPv6) are enforced in the service decode with indexed errors, converging with the legacy path's mask validation; a class that the engine later learns to accept becomes a validation relaxation, not a wire migration.
- Removes `BiContiguousIPv6Network` (message, helpers, tests) — the acl lists were its only consumer; `buf breaking` reports exactly that deletion, per the epic posture.
- The CLI stays a thin pass-through: any parseable mask reaches the wire and the server rejects out-of-class shapes, so no client-side class validation is duplicated.
- Closes #2234; stacked on #2238.


3Hren:
@codex review

PR #2251, opened 2026-08-25
feat(readiness): give scopes a per-scope observation interval

### LINKED ISSUE #1747: feat(controlplane): give readiness scopes a per-scope freshness interval
## Motivation

`common/go/readiness/readiness.go`'s `Tracker` maintains `observedAt` per scope and exposes `Touch()` explicitly so freshness consumers can distinguish a live scope from one whose last event was long ago. But `readinesspb.Scope` (`common/readinesspb/*/readiness.proto`, `message Scope` at line 23) carries only `name`/`state`/`reasons`/`observed_at` (line 29)/`last_transition_time` (line 31) — no field for how often that scope is expected to heartbeat. Confirmed unchanged on current `main`.

Each readiness source ticks on its own cadence with nothing publishing it: the route operator's neighbour monitor is a 5-minute ticker (`operators/route/internal/discovery/neigh/neigh.go`, `UpdateInterval: 5 * time.Minute`), gateway/operator reconcile actuators are ~30s, the bird RIB readiness helper is ~1s. Any staleness consumer (CLI badge, future Prometheus alert, future web badge) has to guess a single global threshold instead of reading a per-scope contract. Verified live: a 60s threshold flagged the healthy `neighbours` scope (5-minute cadence) as stale; the in-progress CLI redesign ships a deliberately generous 15-minute default (`--stale-after`, 0 disables) purely as a safety margin, a guess rather than a contract.

## Scope

- The freshness contract for scopes that already exist today: gateway plus the four operators (route, forward, decap, pipeline), and bird-adapter if it registers scopes. Not a module-side readiness tracker.

## Acceptance

- [ ] `readinesspb.Scope` gains an `expected_freshness` (or equivalent) duration field, set from the source's own ticker/heartbeat interval when the scope is registered.
- [ ] Every `NewTracker`/scope-registration call site (gateway plus the four operators; bird-adapter if applicable) supplies its real interval.
- [ ] Scopes with no natural heartbeat (e.g. gateway's set-once `gateway` scope) declare none and are never judged stale.
- [ ] Consumers compute staleness as `now - observed_at > k * expected_freshness` (k tunable), not a single global constant.
- [ ] `cli/modules/ready/src/render/age.rs`'s `is_stale` uses the per-scope value; `--stale-after` is dropped or repurposed as the multiplier `k` rather than an absolute duration.

Source: local task tracker migration, item dated 2026-07-14 (updated 2026-08-05), verified against current `main` on 2026-08-06.

saushew:
- Adds `expected_observation_interval` to readiness scopes so consumers can derive a threshold from each source's nominal cadence.
- Publishes reconcile, RIB, BIRD, and neighbour observation intervals from the gateway and operator scope registrations while leaving set-once scopes without a freshness contract.
- Replaces the CLI's absolute `--stale-after` threshold with `--stale-multiple`, using `age > multiplier * expected_observation_interval`; `0` disables stale tags.
- Keeps `Watch` transition-only: its initial message is a full snapshot and subsequent messages contain only state or reason changes; freshness checks use current `Ready` snapshots.
- Serializes the observation interval in CLI JSON output and documents the per-scope freshness contract for each operator.
- Covers distinct source cadences, scopes without heartbeats, CLI boundaries and arguments, and JSON serialization.

Closes #1747.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3653,6 +3653,7 @@ dependencies = [
  "prost",
  "prost-types",
  "serde",
+ "serde_json",
  "tonic",
  "tonic-build",
 ]
--- a/cli/core/src/output.rs
+++ b/cli/core/src/output.rs
@@ -435,6 +435,17 @@ pub fn paint_dim(text: &str) -> String {
     text.truecolor(127, 127, 127).to_string()
 }
 
+/// Returns `true` if the installed backend serializes rather than renders.
+///
+/// Answers the inverse of [`data`]'s split: code that renders outside a
+/// `data` closure — a periodic timer in a long-running watch loop, say —
+/// can check whether free-form output is wanted at all instead of relying
+/// on a side effect such as [`is_colored`], which folds in `NO_COLOR` and
+/// stderr's TTY-ness and would misclassify a piped human run.
+pub fn serializes() -> bool {
+    current().serializes()
+}
+
 /// Returns `true` if the current locale advertises UTF-8 encoding.
 fn is_utf8_locale() -> bool {
     for var in ["LC_ALL", "LC_CTYPE", "LANG"] {
--- a/cli/core/src/output.rs
+++ b/cli/core/src/output.rs
@@ -435,17 +435,6 @@ pub fn paint_dim(text: &str) -> String {
     text.truecolor(127, 127, 127).to_string()
 }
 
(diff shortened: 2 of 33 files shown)


saushew:
@codex review

saushew:
@codex review

3Hren (approved):

PR #2259, opened 2026-08-25
feat(acl): serve per-rule counter metrics from their own method

OrIX219:
GetMetricsRules returns acl_rule_packets and acl_rule_bytes for every rule counter, labelled with the config, the position and the counter name. It reads the runtime-kind storages GetMetrics never touches, so GetMetrics keeps returning the predefined counters alone.

The request names the config, device, pipeline, function and chain to report on, with an empty field matching every value and the module type implied. A position the request excludes is skipped before its counters are read rather than filtered out afterwards.

The acl CLI gains a metrics-rules command with a flag per selector. Rule metrics need a renderer of their own, because the existing metrics table keeps only counters that carry no counter label. GetRulesCounters keeps serving the same counters in its own shape.

### DIFF
--- a/modules/acl/cli/src/args.rs
+++ b/modules/acl/cli/src/args.rs
@@ -16,6 +16,8 @@ pub enum ModeCmd {
     Show(ShowCmd),
     /// Show ACL metrics
     Metrics(MetricsCmd),
+    /// Show per-rule ACL counter metrics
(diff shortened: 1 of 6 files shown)


OrIX219:
@codex review

OrIX219:
@codex review

GeorgyKirichenko (approved):

PR #2302, opened 2026-08-27
fix(functional): require kvm device access

sakateka:
Verify that `/dev/kvm` can be opened read-write before enabling QEMU hardware acceleration.
Fall back to TCG when the KVM device exists but is inaccessible.
Cover accessible and inaccessible device paths with regression tests.


sakateka:
@codex review

sakateka:
@codex review

3Hren (approved):

PR #2303, opened 2026-08-27
refactor(counters)!: move the cross-worker counter merge into the go bindings

GeorgyKirichenko:
- The merged by-tags read and the six name-parameter wrappers leave the agent C library; every counter read now goes through the per-worker API and the cross-worker union lives solely in the Go bindings.
- The Go name wrappers keep their signatures and nil/empty contracts, building their tag sets in Go and flattening the merged groups.
- The lock, race, leak and snapshot C tests read through the per-worker API with identical pin and lock windows; merged-list tests are replaced by per-worker shared-counter assertions plus the existing Go merge tests, and a new Go test pins the name reads end to end — every counter spans one value set per worker.
- Entity names longer than the tag field now fail the read instead of prefix-matching; every existing caller sources names from fixed-size configuration fields that cannot exceed the limit.


GeorgyKirichenko:
@codex review

PR #2305, opened 2026-08-27
refactor(counters)!: chunk the registry names array

### LINKED ISSUE #1742: refactor(common): chunk counter_registry's names and links arrays to cut peak arena demand
## Motivation

`counter_registry_expand` (`lib/counters/counters.c`) grows the registry's names array by allocating a brand-new, larger contiguous block sized to the new capacity, copying the old contents into it, and only then freeing the old block — so during every growth step it holds both the old and new arrays live simultaneously, 1.5x the final size. Growth is doubling, so this is the single largest contiguous allocation demand made against the shared-memory agent arena by any subsystem profiled. A companion investigation into the arena's inability to reclaim capacity from freed blocks identified this specific growth pattern as the first thing to starve as the allocator's largest formable block shrinks over time: it is both the biggest single ask and, being a doubling contiguous allocation, disproportionately likely to be the one request a fragmented arena can no longer satisfy even when plenty of free bytes remain in aggregate.

The function already carries an existing FIXME calling out the copy as inefficient for a different reason — the copy cost itself. This work is that FIXME, scoped and prioritized by the additional finding that the same pattern is also the arena's primary starvation source, not just a per-growth copy-cost inefficiency.

This is debt, not a defect: `counter_registry_expand` behaves correctly today. It is costly — the largest contiguous demand in the arena, a 1.5x transient overhead during growth, an already-acknowledged inefficiency — not wrong. Fixing it mitigates the exposure of a separate, more fundamental allocator defect (its inability to coalesce freed memory back into larger blocks) by shrinking that defect's most exposed starving class, but does not itself fix the allocator's inability to coalesce — that remains open as its own, larger, architect-level decision.

## Scope

- Replace `names` (and `links`, covered by the same FIXME) with chunked storage, so growth allocates one new fixed-size chunk instead of a new full-size copy of everything.
- Eliminate the transient 1.5x (old-plus-new) contiguous demand that today's doubling-copy growth path creates.

## Acceptance

- [ ] `counter_registry`'s `names` and `links` arrays are backed by chunked storage instead of a single contiguous doubling allocation.
- [ ] No transient old-plus-new contiguous demand remains during growth.
- [ ] Existing counter-registry tests pass unchanged; a new test asserts growth beyond several chunk boundaries never requires a single allocation larger than one chunk.
- [ ] If a compile-time arena/throughput regression benchmark is ever built, the largest formable block after a sequence of registry growths should be measurably improved relative to before this change — not a blocking condition of this work.

Source: packet-path safety backlog migration; originally extracted from a broader arena-fragmentation investigation as its cheapest, highest-leverage, and most contained remedy, needing no architect-level decision unlike the underlying allocator defect it mitigates.

GeorgyKirichenko:
Replace the counter registry's doubling-copy names array with fixed-size chunks, cutting the arena's largest single contiguous allocation demand and its 1.5x growth transient.

- Growth allocates exactly one chunk of entries plus a tiny re-housed pointer array: the registry never asks the arena for a block larger than one chunk, byte-tied to the 32 KiB allocator class, and never holds old and new storage live at once.
- The chunk pointer array grows with EQUATE_OFFSET rather than realloc-copy, because the slots are self-relative pointers that would otherwise resolve against their old addresses.
- The registry struct grows by eight bytes, so the dataplane plugin size tripwires move and YANET_MODULE_ABI_VERSION bumps to 20.
- A probe-shimmed build of the registry sources backs a new acceptance test: registering counters across several chunk boundaries never allocates more than one chunk, lookups and linking keep resolving, and fini returns every arena byte.
- Tight-arena tests recalibrate their limits to the new one-chunk floor, with comments explaining the sizing.

The dataplane packet path is unaffected: workers increment counters through handles resolved at config activation, and the value pages were already chunked.

Closes #1742.

### DIFF
--- a/lib/controlplane/agent/counters.c
+++ b/lib/controlplane/agent/counters.c
@@ -335,19 +335,17 @@ worker_counter_list_build(
 		struct cp_counter_storage *cp_storage = matches[i];
 		struct counter_storage *storage = ADDR_OF(&cp_storage->storage);
 		struct counter_registry *registry = ADDR_OF(&storage->registry);
-		struct counter *counters = ADDR_OF(&registry->names);
-
 		size_t storage_matches = 0;
 		size_t region = 0;
 		for (uint64_t idx = 0; idx < registry->count; ++idx) {
-			if (!counter_pattern_set_match(
-				    names, counters[idx].name
-			    )) {
+			struct counter *counter =
+				counter_registry_entry(registry, idx);
+			if (!counter_pattern_set_match(names, counter->name)) {
 				continue;
 			}
 			size_t counter_region;
 			if (!counter_value_region_size(
-				    1, counters[idx].size, &counter_region
+				    1, counter->size, &counter_region
 			    ) ||
 			    !counter_region_add(&region, counter_region)) {
 				yanet_error_add(
@@ -385,12 +383,11 @@ worker_counter_list_build(
 		struct cp_counter_storage *cp_storage = matches[i];
 		struct counter_storage *storage = ADDR_OF(&cp_storage->storage);
 		struct counter_registry *registry = ADDR_OF(&storage->registry);
-		struct counter *counters = ADDR_OF(&registry->names);
 		struct counter_tag *storage_tags = NULL;
 		for (uint64_t idx = 0; idx < registry->count; ++idx) {
-			if (!counter_pattern_set_match(
-				    names, counters[idx].name
-			    )) {
+			struct counter *counter =
+				counter_registry_entry(registry, idx);
+			if (!counter_pattern_set_match(names, counter->name)) {
 				continue;
 			}
 			if (storage_tags == NULL) {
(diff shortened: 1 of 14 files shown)


GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

GeorgyKirichenko:
@codex review

GeorgyKirichenko:
Addressed the latest review round:

- The chunk-pointer array is gone: chunk pointers now live in fixed 8 KiB directory pages (1,024 slots each), so growth past 65,536 counters adds one directory page — also a single 8 KiB class block — instead of an unbounded flat array. The pages array itself is re-housed with EQUATE_OFFSET and stays a handful of pointers (one page covers 65,536 counters). The acceptance test now registers 65,546 counters, crossing a directory-page boundary, and asserts no allocation exceeds one chunk.
- The route web editor now validates the reduced 103-byte counter-name limit (constant comment cites the C-side derivation), with the boundary tests moved from 127/128 to 103/104; a client-accepted name can no longer be rejected server-side. All 484 web tests pass.

GeorgyKirichenko:
@codex review

GeorgyKirichenko:
the result is bilge so closing

PR #2306, opened 2026-08-27
refactor(fwstate,dataplane,pdump): make fixed-width byte reads alias-safe

### LINKED ISSUE #2281: refactor(fwstate,dataplane,pdump): drop the remaining fixed-width pointer reads over byte arrays
## Motivation

#2279 makes the `*(uint32_t *)` / `*(uint64_t *)` reads of network byte arrays in the filter compiler safe under strict aliasing and at any alignment: the plain idiom is undefined behaviour under strict aliasing and a misaligned access whenever the bytes sit at an odd address, and it surfaced on aarch64. A tree-wide grep for the same shape over yanet sources (excluding `subprojects/`) leaves three sites. None has a reproduced symptom; each is listed below with what the code proves, so the outcome is a verdict per site rather than a blind rewrite.

## Scope

1. [`lib/fwstate/fwmap.h:249-257`](https://github.com/yanet-platform/yanet2/blob/0ba13d696f5f0d02b03d71f38bbb103cce32ba42/lib/fwstate/fwmap.h#L249-L257) — the default key comparison reads 4- and 8-byte keys through `uint32_t` / `uint64_t` lvalues over `const void *`. Code-proven: the production maps are built with `sizeof(struct fw4_state_key)` / `sizeof(struct fw6_state_key)` keys ([`fwmap_typed.h:77,109`](https://github.com/yanet-platform/yanet2/blob/0ba13d696f5f0d02b03d71f38bbb103cce32ba42/lib/fwstate/fwmap_typed.h#L77)), which take the `memcmp` branch, so the fixed-width branches are reachable only for other key sizes. If they are ever hit with a non-integer key object the read violates the effective-type rule. A constant-size `memcmp` folds to the same single compare at `-O2`.
2. [`lib/dataplane/packet/mss.c:194-215`](https://github.com/yanet-platform/yanet2/blob/0ba13d696f5f0d02b03d71f38bbb103cce32ba42/lib/dataplane/packet/mss.c#L194-L215) — the MSS option (a packed `kind`, `len`, `data[]` structure inside mbuf data) is written and read through `uint16_t *` casts, including reading the two `uint8_t` members `kind` and `len` as one `uint16_t` for the incremental checksum. Alignment holds in practice because the option follows the fixed TCP header after even L2/L3 lengths, but the aliasing violation is code-proven. The routine has no caller in this repository besides `lib/tests/dataplane/mss_test.c`.
3. [`modules/pdump/dataplane/ring.h:76-84`](https://github.com/yanet-platform/yanet2/blob/0ba13d696f5f0d02b03d71f38bbb103cce32ba42/modules/pdump/dataplane/ring.h#L76-L84) — reads the message header's first field through `uint32_t *` at a ring position that every writer keeps 4-byte aligned, and the header is copied into the ring with `memcpy` from a `struct ring_msg_hdr`, so both alignment and effective type are established by construction. Listed for the audit verdict; keeping the cast with the existing comment is an acceptable outcome.

## Closing outcome

Every listed site either accesses the bytes through `memcpy` (or an equivalent aliasing-safe, unaligned-safe helper) with the constant-size copy folding to the same instruction at `-O2`, or keeps the cast with a code-proven justification recorded in its comment. A tree-wide grep for `\*\((const )?uint(16|32|64)_t \*\)` over yanet sources (excluding `subprojects/`) then returns only justified sites.

## Acceptance

- [ ] `fwmap.h` compares keys without a fixed-width lvalue over `void *`, or the 4/8-byte branches are shown unreachable and removed.
- [ ] `mss.c` writes and reads the option bytes without `uint16_t *` casts, and `lib/tests/dataplane/mss_test.c` passes in the ASan/UBSan build.
- [ ] `ring.h` either uses `memcpy` or keeps the cast with the alignment comment extended to the effective-type argument.
- [ ] `-O2` code generation of the touched hot paths inspected once (constant-size `memcmp` / `memcpy` fold to single loads and compares).
- [ ] `make test-asan` green and `aarch64-linux-gnu-gcc -O2 -Werror` clean for the touched files.

## Source

Follow-up to #2279. Precedent for the class: #150 (strict aliasing in `ADDR_OF`).

sakateka:
- Replace the remaining fixed-width pointer dereferences over byte-oriented storage in firewall-state key comparison, TCP MSS handling, and the packet-dump ring with constant-size `memcmp` or `memcpy` operations.
- Preserve byte-for-byte comparison, network byte order, incremental checksum updates, and ring reclamation while removing alignment and effective-type assumptions.
- Closes #2281.

### DIFF
--- a/lib/dataplane/packet/mss.c
+++ b/lib/dataplane/packet/mss.c
@@ -159,17 +159,18 @@ static void
 clamp_mss_option(
 	struct rte_tcp_hdr *tcp, struct tcp_option *opt, uint16_t clamp_mss
 ) {
-	uint16_t *mss_ptr = (uint16_t *)opt->data;
-	uint16_t old_mss = rte_be_to_cpu_16(*mss_ptr);
-	if (old_mss <= clamp_mss) {
+	uint16_t old_mss_be;
+	memcpy(&old_mss_be, opt->data, sizeof(old_mss_be));
+	if (rte_be_to_cpu_16(old_mss_be) <= clamp_mss) {
 		return;
 	}
 
 	/* Incremental TCP checksum update per RFC 1624. */
 	uint16_t cksum = ~tcp->cksum;
-	cksum = csum_minus(cksum, *mss_ptr);
-	*mss_ptr = rte_cpu_to_be_16(clamp_mss);
-	cksum = csum_plus(cksum, *mss_ptr);
+	cksum = csum_minus(cksum, old_mss_be);
+	uint16_t new_mss_be = rte_cpu_to_be_16(clamp_mss);
+	memcpy(opt->data, &new_mss_be, sizeof(new_mss_be));
+	cksum = csum_plus(cksum, new_mss_be);
 	/* preserve all-ones checksum (RFC 1624). */
 	tcp->cksum = (cksum == 0xffff) ? cksum : ~cksum;
 }
@@ -196,7 +197,8 @@ write_into_mss_gap(struct rte_mbuf *mbuf, uint16_t offset, uint16_t mss) {
 		rte_pktmbuf_mtod_offset(mbuf, struct tcp_option *, offset);
 	opt->kind = TCP_OPTION_KIND_MSS;
 	opt->len = TCP_OPTION_MSS_LEN;
-	*(uint16_t *)opt->data = rte_cpu_to_be_16(mss);
+	uint16_t mss_be = rte_cpu_to_be_16(mss);
+	memcpy(opt->data, &mss_be, sizeof(mss_be));
 	return opt;
 }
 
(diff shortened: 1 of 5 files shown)


sakateka:
@codex review

sakateka:
@codex review

sakateka:
Performance and codegen were verified with GCC as the primary compiler and Clang as an additional check. The assembly below comes from real release artifacts rather than only from a standalone codegen harness.

- x86-64: GCC 13.3, `-O2 -march=haswell`; the complete Meson build finished successfully (1019/1019 targets), as did the release Rust workspace build.
- AArch64: GCC 15.3, `-O2 -march=armv8-a+crc`; the complete dataplane cross-build finished successfully (694/694 targets). The final `yanet-dataplane` binary and the MSS test suite were executed under QEMU 10.1.5. The MSS suite passed all 16 cases.

The AArch64 build excludes only the mlx5 PMDs because the cross sysroot has no RDMA/ibverbs libraries; this does not affect the code reviewed here.

In the final dataplane, the alias-safe constant-size `memcpy` loads in the default comparator are fully eliminated by GCC. The 4-byte path is:

```asm
# x86-64
mov  eax, DWORD PTR [rsi]
cmp  DWORD PTR [rdi], eax
sete al
ret
```

```asm
// AArch64
ldr  w0, [x1]
ldr  w1, [x5]
cmp  w1, w0
cset w0, eq
ret
```

The 8-byte path has the same shape with 64-bit loads. Only the generic-size fallback calls `memcmp`; the optimized 4/8-byte paths contain no libc calls and match the original typed-load codegen.

GCC also keeps the IPv6 comparison inline. On Haswell it emits two 16-byte SIMD comparisons, one per address:

```asm
vmovdqu xmm1, XMMWORD PTR [rdi + 8]
vpxor   xmm0, xmm1, XMMWORD PTR [rsi + 8]
vptest  xmm0, xmm0
jne     .Lnot_equal
vmovdqu xmm0, XMMWORD PTR [rdi + 24]
vpxor   xmm0, xmm0, XMMWORD PTR [rsi + 24]
vptest  xmm0, xmm0
sete    al
ret
```

The real AArch64 build emits the corresponding two 128-bit SIMD comparisons:

```asm
ldur  q31, [x0, #8]
ldur  q30, [x1, #8]
eor   v30.16b, v31.16b, v30.16b
umaxp v30.4s, v30.4s, v30.4s
fmov  x3, d30
cbnz  x3, .Lnot_equal
```

The second address uses the same sequence. There is no `memcmp` call in this path. Clang 18 remains a useful additional check and folds both addresses into one 32-byte comparison on x86-64, but GCC is the primary compiler and its actual project codegen is still efficient.

The MSS and pdump fixed-width accesses also compile to direct narrow loads/stores on both architectures. Their object files contain no `memcpy` introduced by these changes. The remaining `memmove` in MSS inserts TCP options, and the remaining `memcpy` calls in pdump copy real packet metadata/payload.

I do not see a codegen or build regression in the final version on either architecture.

sakateka:
@codex review

3Hren (approved):

PR #2310, opened 2026-08-28
Support for proper testing of chains with several outputs

kriomant:
Currently tests can only use single output device. Which means that chains with multiple outputs cannot be tested properly.

Say, tests for 'mirror' module check that two packets arrived, but they can't verify that second packet went from another device, so if 'mirror' module has bug in routing, it won't be caught.

This change adds support for multi-device tests both to unit and function tests. Separate VM with two devices is used to not touch existing tests which doesn't require it.


GeorgyKirichenko:
@codex review

PR #2314, opened 2026-08-28
fix(test-functional): use guest-local cli paths in snapshots

sakateka:
- Ensure snapshot-backed VMs use the CLI binaries stored with the snapshot instead of potentially newer binaries from the host-mounted build directory.
- Keep booted single-VM guests on the shared target mount.
- Verify that the generic CLI can dispatch to sibling module binaries from guest-local storage.


sakateka:
@codex review

moonug (approved):

3Hren (approved):

PR #2395, opened 2026-08-30
fix(devices): drop the vlan short flag and bound the vlan id

### LINKED ISSUE #2337: fix(devices): vlan cli -v means both --vlan and verbosity and accepts any vlan id
## Motivation

In `yanet-cli device vlan update` the short flag `-v` is declared twice: as the global verbosity counter and as the short form of `--vlan`. After the subcommand, `-v` resolves to `--vlan`, so verbosity cannot be raised there and `-vv` is a parse error. Clap's debug assertions would flag the duplicate on any debug build that constructs the subcommand.

The same command accepts any `u16` as the VLAN id. Nothing between the CLI and the dataplane masks it: the Go service casts the value, the C side stores it raw and the dataplane writes it into the TCI unmasked, so `--vlan 5000` sets PCP/DEI bits on egress instead of failing.

Expected: one meaning per short flag; a VLAN id outside `0..=4094` is rejected by the client with a usage error (the server-side check is tracked separately).

## Evidence

Reproduced on a release build of `620c1316`:

```
$ yanet-cli-device-vlan update -n x --vlan 5 -vv --endpoint grpc://127.0.0.1:1
error: the argument '--vlan <VLAN>' cannot be used multiple times
[exit 2]
$ yanet-cli-device-vlan update --help | grep -A1 -- ' -v'
  -v, --vlan <VLAN>
          Vlan tag
  -v...
          Log verbosity level
$ yanet-cli-device-vlan update -n x --vlan 5000 --endpoint grpc://127.0.0.1:1
[ERR] connect failed: transport error          # accepted by the client, would be sent
```

Code: global `-v` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/devices/vlan/cli/src/main.rs#L32-L33; `--vlan` with `short` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/devices/vlan/cli/src/main.rs#L52-L54; unmasked TCI write — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/devices/vlan/dataplane/dataplane.c#L111-L112; the `0` value meaning untagged is documented nowhere in `--help` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/devices/vlan/dataplane/dataplane.c#L57-L91.

## Scope

`yanet-cli-device-vlan update`. Server-side validation is part of the service-side range validation issue.

## Acceptance

- [ ] `-v` after `update` raises verbosity; `--vlan` has no colliding short form.
- [ ] `Cmd::command().debug_assert()` runs in the crate's tests.
- [ ] A VLAN id outside `0..=4094` is a usage error before any connection; `--help` states the range and that `0` means untagged.

3Hren:
- `-v` after `update` raises verbosity again: the vlan id no longer has a short form that shadowed the global counter.
- The vlan id is validated on the client as `0..=4094` before any connection, and the help states that `0` emits untagged frames.
- Closes #2337.

### DIFF
--- a/devices/vlan/cli/src/main.rs
+++ b/devices/vlan/cli/src/main.rs
@@ -1,4 +1,4 @@
-use clap::{ArgAction, CommandFactory, Parser};
+use clap::{ArgAction, CommandFactory, Parser, value_parser};
 use clap_complete::CompleteEnv;
 use code::{UpdateDeviceVlanRequest, device_vlan_service_client::DeviceVlanServiceClient};
 use commonpb::pb::Device;
@@ -49,8 +49,8 @@ pub struct UpdateCmd {
     /// Pipeline assignments in format "pipeline_name:weight"
     #[arg(short, long)]
     pub output: Vec<String>,
-    /// Vlan tag
-    #[arg(short, long)]
+    /// VLAN id in 0..=4094, where 0 makes the device emit untagged frames.
+    #[arg(long, value_parser = value_parser!(u16).range(0..=4094))]
     pub vlan: u16,
 }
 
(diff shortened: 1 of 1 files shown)


3Hren:
@codex review

PR #2405, opened 2026-08-31
fix(cli): validate module update arguments

### LINKED ISSUE #2341: fix(cli): payload flags are optional and address values untyped in several module clis
## Motivation

Several module CLIs let a whole-set `update` go out with an empty payload because the payload flag is optional, and several parse addresses or numeric limits from strings or over-wide integer types after connecting. The service-side invariants (tracked separately) are the real guard; the client's job is to reject a malformed or incomplete command line before it reaches the network, with a usage error the user can act on.

Observed today (code-proven on `620c1316`):

- `yanet-cli pipeline update --name X` without `--functions` sends `functions: []` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/cli/modules/pipeline/src/main.rs#L76-L83. The twin `function update` already requires `--chains` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/cli/modules/function/src/main.rs#L69-L71.
- `yanet-cli decap update --name X` without `--prefixes` sends an empty set — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/decap/cli/src/main.rs#L58-L66.
- `yanet-cli route-mpls update --label` accepts any `u32`; MPLS labels are 20 bits — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/route-mpls/cli/src/main.rs#L80-L100.
- `yanet-cli fwstate update --src-addr` / `--dst-addr-multicast` are `Option<String>` parsed after connecting, `--port-multicast` is `u32` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/fwstate/cli/src/args.rs#L56-L66, parsing at https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/fwstate/cli/src/main.rs#L198-L205.
- `yanet-cli acl update --port-multicast` / `--port-unicast` are `u32` with no `0..=65535` check — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/acl/cli/src/args.rs#L32-L66.

Expected: required payload flags, typed clap values (`IpAddr`/`Ipv6Addr`, `u16` ports, bounded labels) so that every malformed value is a usage error before any connection.

## Scope

Argument definitions of `pipeline`, `decap`, `route-mpls`, `fwstate`, `acl` CLIs. Not in scope: server-side validation (service-side epic) and the semantics of an empty update.

## Acceptance

- [ ] `pipeline update` and `decap update` require their payload flag.
- [ ] Labels, ports and addresses are typed and bounded at parse time; a bad value exits 2 before connecting.
- [ ] No parse test is added: the constraints are clap declaratives (owner decision 2026-08-30, see the `code-cli` skill).

3Hren:
- Reject incomplete whole-set updates and malformed labels, ports, and addresses before establishing a gRPC connection.
- Preserve the existing protobuf shapes by widening typed `u16` ports losslessly to their `uint32` wire fields.
- Treat ACL synchronization IPv6 and MAC flags as part of the issue address-validation scope because they used the same post-connection parsing path.
- Apply the 20-bit label bound to both route updates and withdrawals because both populate the same MPLS wire field, while leaving label semantics to the service.
- Treat payload-flag presence, rather than the semantics of an explicitly empty set, as the CLI boundary described by the issue.
- Add no clap parsing tests, following the issue explicit owner decision for declarative constraints.

Closes #2341.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3115,6 +3115,7 @@ dependencies = [
  "clap",
  "clap_complete",
  "log",
+ "netip",
  "prost",
  "prost-types",
  "ptree",
--- a/cli/modules/pipeline/src/main.rs
+++ b/cli/modules/pipeline/src/main.rs
@@ -78,7 +78,7 @@ pub struct UpdateCmd {
     #[arg(short, long, add = ArgValueCandidates::new(pipeline_candidates))]
     pub name: String,
     /// Pipeline functions.
-    #[arg(long, value_delimiter = ',')]
+    #[arg(long, required = true, num_args = 0.., value_delimiter = ',')]
     pub functions: Vec<String>,
 }
 
--- a/modules/acl/cli/Cargo.toml
+++ b/modules/acl/cli/Cargo.toml
@@ -13,6 +13,7 @@ workspace = true
 ync = { path = "../../../cli/core", version = "0.1", package = "yanet-cli" }
 commonpb = { path = "../../../common/rust/commonpb", version = "0.1", package = "yanet-commonpb" }
 filterpb = { path = "../../../common/rust/filterpb", version = "0.1", package = "yanet-filterpb" }
+netip = "0.3"
 log = "0.4"
 clap = { version = "4.5", features = ["derive", "wrap_help"] }
 clap_complete = { version = "4.5", features = ["unstable-dynamic"] }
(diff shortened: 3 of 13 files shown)


3Hren:
@codex review

3Hren:
@codex review

PR #2408, opened 2026-08-31
fix(nat64): validate and render cli prefixes

### LINKED ISSUE #2336: fix(nat64): cli prints prefixes as byte arrays and accepts any prefix length
## Motivation

`yanet-cli nat64 show` renders every configured prefix as a Rust debug dump of raw bytes instead of an IPv6 network, and `--format json` emits the same numbers. An operator cannot read back what is configured.

Separately, `nat64 prefix add --prefix` accepts any IPv6 network and silently drops the prefix length: `2001:db8::/64` is sent — and stored — as the first 12 bytes, i.e. as a `/96`, without any diagnostic.

Expected: `show` prints `2001:db8::/96`; a prefix whose length is not `/96` is rejected before the request is sent, with a message naming the requirement.

## Evidence

Code-proven.

- Human rendering uses `{:?}` on `Vec<u8>` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/nat64/cli/src/main.rs#L437-L439; the JSON payload is the raw proto with the same bytes.
- Only the first 12 address bytes are taken and the prefix length is discarded — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/nat64/cli/src/main.rs#L293-L296 and https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/nat64/cli/src/main.rs#L313-L316; the help text says "IPv6 prefix (12 bytes)" — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/nat64/cli/src/main.rs#L100-L113.
- The server checks the byte count only — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/nat64/controlplane/service.go#L196-L217.

Related: #2197 (family-typed address wire model) will change the JSON representation of the prefix bytes; the human rendering and the client-side length check are independent of it.

## Scope

`yanet-cli-nat64 show`, `prefix add`, `prefix remove`. Server-side length validation is tracked with the service-side range validation work.

## Acceptance

- [ ] `show` renders each prefix as an IPv6 network in both human and JSON output.
- [ ] `prefix add`/`prefix remove` reject a prefix length other than `/96` with a usage error (exit 2) before connecting.
- [ ] A unit test covers the rendering and the length check.

3Hren:
- Render stored 12-byte NAT64 prefixes as canonical IPv6 /96 networks in human and JSON CLI output.
- Reject add and remove inputs whose prefix length is not /96 before opening a connection.
- Treat malformed server prefix lengths as invalid instead of panicking.
- Keep the existing 12-byte wire model and server-side validation unchanged until #2197.
- Preserve the legacy NAT64 tree layout and change only the prefix leaf representation.
- Assume existing JSON consumers tolerate the intentional nested prefix value change from byte arrays to CIDR strings.
- Cover the formatter and fixed-length parser with focused unit tests.

Closes #2336.

### DIFF
--- a/modules/nat64/cli/build.rs
+++ b/modules/nat64/cli/build.rs
@@ -8,6 +8,10 @@ pub fn main() -> Result<(), Box<dyn Error>> {
         .build_server(false)
         .extern_path(".common.commonpb.v1", "::commonpb::pb")
         .message_attribute(".", "#[derive(Serialize)]")
+        .field_attribute(
+            ".modules.nat64.controlplane.nat64pb.v1.Prefix.prefix",
+            "#[serde(serialize_with = \"crate::serialize_prefix\")]",
+        )
         .compile_protos(&["nat64pb/v1/nat64.proto"], &["../controlplane", "../../.."])?;
 
     Ok(())
--- a/modules/nat64/cli/src/main.rs
+++ b/modules/nat64/cli/src/main.rs
@@ -99,8 +99,8 @@ pub struct AddPrefixCmd {
     /// The name of the config to operate on.
     #[arg(long = "name", short = 'n', add = ArgValueCandidates::new(config_candidates))]
     pub config_name: String,
-    /// IPv6 prefix (12 bytes) to be added.
-    #[arg(long)]
+    /// IPv6 `/96` prefix (12 bytes) to be added.
+    #[arg(long, value_parser = parse_prefix)]
     pub prefix: Contiguous<Ipv6Network>,
 }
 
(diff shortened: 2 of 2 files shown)


3Hren:
@codex review

PR #2419, opened 2026-08-31
feat(nat64)!: carry typed prefixes on the wire

3Hren:
- Replace all NAT64 prefix byte fields with the shared family-typed IPv6 prefix message in one wire cutover.
- Preserve strict /96 validation while keeping the dataplane's 12-byte representation behind the control-plane adapter.
- Reuse commonpb CIDR rendering and serialization in the Rust CLI.
- Exclude only the NAT64 protobuf file from Buf breaking until the new baseline reaches main.

Refs #2197.


3Hren:
@codex review

3Hren:
@codex review

3Hren:
@codex review

PR #2439, opened 2026-08-31
feat(l3b): add the l3b load-balancer module

GeorgyKirichenko:
- Adds the l3b IP-in-IP load-balancer module: module-level destination filters classify initial-fragment TCP/UDP traffic into virtual services, per-service source filters gate it, and a hash-masked weighted round-robin ring selects a real server to encapsulate the packet towards with a derived outer source address.
- Virtual services are standalone named shared objects under objects/l3b, linked into module configurations by the service names their destination rules carry through cp_module_link_object and resolved per worker at execution-context build time; real-server weight (clamped to 1000) and state updates mutate the published object in place behind acquire/release barriers and a seqlock-guarded ring.
- Every service owns a session table — a layered statemap table in a second cp_object under the l3b_session_table type — pinning each client flow (source address and port) to its real server for 300 s, with first-write-wins promotion so flows survive layer rotations and ring retargets.
- Extracts the generic layered hash map core into lib/statemap with a startup callback registry, and builds lib/l3state for the l3b flow-state primitives, leaving fwstate with only its firewall-specific pieces.
- Ships the full control plane: L3bService gRPC API with protos, Go bindings, a Rust CLI on the current ync API whose services and module configs populate from YAML documents, and end-to-end functional tests covering the filter, session, ring and encapsulation paths through the live dataplane.


GeorgyKirichenko:
All 11 Codex findings addressed in a2b41081:

- Fragment exclusion (P1): non-initial fragments are forwarded untouched — classified only when `packet->fragment_offset == 0`, with a functional test.
- CLI service/module documents (P1): `create/update-service --file` now carries real servers and source filter rules, `update-module-config --file` carries destination filter rules (YAML schema over the shared filterpb network types).
- Real-server family mismatch (P1): rejected in `buildVirtualServiceConfig` before the union members are read.
- Ring replacement races (P2): the ring is now a seqlock — the writer marks it unstable, rewrites entries, publishes the count and re-stabilizes; the dataplane retries the selection on a version change.
- Zero-length VLAs (P2): early return on an empty front, mirroring forward.
- NotFound/AlreadyExists statuses (P2): the backend reports proper gRPC codes so the CLI exits 3 on missing services.
- Deferred-handle reclamation (P2): retried at the start of every mutating call, not only on delete.
- CString leak (P2): service-name strings freed after the C update copies them.
- Virtual-service index validation (P2): rejected on the Go side before crossing into C.
- Empty-list markers (P2): printed only for empty lists with an early return.

@codexb[bot] please re-review: a2b41081

GeorgyKirichenko:
Destination filter rules now carry the virtual service **name** instead of an index: `l3b_destination_filter_rule.virtual_service` names the object, `l3b_module_config_update` links each rule's service through `cp_module_link_object` (rules naming the same service share one link), and the redundant `ModuleConfig.services` array is gone from the proto and CLI — the linked service set is exactly what the rules name (6c48a677).

GeorgyKirichenko:
Session stickiness added (3ffc28ae):

- Every virtual service now owns a session table — a layered fwtable published as a second cp_object under type `l3b_session_table` with the service's name, created, published and destroyed together with the service object (`session_index_size` on the proto/config sizes its hash index; 300 s default record TTL).
- The dataplane handler builds a session key from the client's source address and port, looks the flow up in the table first and reuses the pinned real server; only flows without a live record go through the hash-masked ring, and the choice is written back as a session record. A pinned real that became disabled or vanished falls through to a fresh pick.
- fwmap gained a generic first-write-wins `FWMAP_PROMOTE_VALUE_KEEP_OLD` value function so promoted records keep the real the session was already pinned to across layer rotations.
- New functional test proves stickiness: an existing flow keeps its real across a ring retarget while a new flow follows the updated ring.

GeorgyKirichenko:
Session state moved off fwstate onto its own library stack (360731f1):

- `lib/statemap` — the generic layered hash map core (`fwmap`/`fwtable`) extracted from fwstate; the callback registry is now defined once in the library with startup registration, so the id space stays stable without cross-library header coupling.
- `lib/l3state` — l3b's flow-state library: the `l3s` key/value types, layer config (FNV1A + overwrite-on-update + first-write-wins promotion), TTL/index defaults, lookup/insert helpers and the packet key builder. l3b now depends only on statemap and l3state.
- fwstate keeps its firewall-specific pieces (fw4/fw6 keys, state values, sync, cursors) and registers its callbacks at startup plus on every table creation path; `objects/l3b` swaps `lib_fwstate_dep` for `lib_l3state_dep`.

All local gates green (meson 80/80 including the fwstate suites, full Go suite incl. acl/fwstate/l3b functional tests).

GeorgyKirichenko:
All follow-up findings addressed in 866b7405 (the earlier batch on 3a07be9e was already fixed by b7b66bff):

- Disabled servers leave the scheduler ring (P1): the backend tracks per-server enabled state and rebuilds the ring over enabled servers only, so a disabled backend's share shifts to the healthy ones instead of blackholing; the dataplane-side readiness check remains as a safety net.
- Ring and state accesses are now formally race-free (P2): ring entries are written and read with relaxed atomic stores/loads under the seqlock, and real-server state flips with a release store paired against the dataplane's acquire load.
- Service and module-config names are validated (P2): names longer than the C registry's 79-byte payload, empty, or containing unprintable bytes are rejected before crossing into C.
- The unknown-service path in update-module-config reports NotFound (P2) — landed with the name-based routing refactor in b8553022.
- CLI documents reject protocols the dataplane never classifies (P2): only tcp and udp can appear in proto ranges.
- list-services and list-module-configs sort their responses (P3), update-service document failures attribute to update-service (P3), and the stale source-filter comment above build_destination_filters is gone (P3).

@codex please re-review: 866b7405

GeorgyKirichenko:
Service updates now preserve sessions correctly (7d9aa0c7):

- The replacement service object **adopts** the published session table instead of allocating an empty one, so every pinned flow keeps its backend across an update. The table is destroyed only when the service is deleted.
- Session records now store the real server's **destination address**, not a positional index: a session survives list reorders, and a session whose backend left the list (or was disabled) re-schedules and re-pins instead of silently landing on whatever now occupies its old index.
- The dataplane UT harness gained `ResetWorkerGenerations()` — harness workers only advance their generation inside rounds, so a mid-test configuration publish would otherwise block forever in its generation wait.
- Functional coverage: an update keeps a pinned flow on its backend while new flows follow the new ring, and a list replacement re-pins the orphaned flow onto the new ring.

GeorgyKirichenko:
Session observability added (4295779e):

- `ListSessions(service, cursor, limit)` RPC pages through a service's session records — client source ip:port, pinned real destination, absolute deadline — plus the dataplane time the deadlines are relative to; continuation token 0 marks a complete listing.
- lib/l3state gained a table cursor (`l3s_table_read`, layer<<32|slot token) and records now carry `expires_at` stamped at insert, so readers never touch the map's bucket metadata. `objects/l3b` exposes the service's table through `l3b_virtual_service_session_table`.
- CLI: `yanet-cli-l3b list-sessions --name <service> [--cursor N] [--limit N]` renders a SOURCE / REAL / EXPIRES-IN table and hints the continuation token when truncated.
- Functional coverage: the update-survival test now also asserts both pinned flows are visible with their correct backends through the listing.

GeorgyKirichenko:
Closing in favor of a fresh PR: GitHub stopped delivering workflow events to this PR after 866b7405 (three subsequent pushes received zero runs while other branches ran normally). The replacement PR carries the same branch and full history.

PR #2449, opened 2026-09-01
perf(dataplane): resolve pipeline stage counters at publish time

GeorgyKirichenko:
Stacked on #2441 — merge that first.

- Pipeline, function, chain and device-entry stages carry only absolute counter pointers, resolved by the publishing process from the counter registry ids and the stage counter storages when a generation is assigned to workers; the relative counter fields are gone.
- Derivation recomputes from controlplane-owned data on every assignment, staying correct across dataplane restarts and re-publications.
- Links the pipeline library into the five cgo consumers of the dataplane config library.


GeorgyKirichenko:
absorbed by #2494

PR #2451, opened 2026-09-01
feat(controlplane): expose the dataplane current time to attached processes

### LINKED ISSUE #2450: feat(controlplane): expose the dataplane current time to attached processes
## Motivation

Timestamps written into shared memory by the dataplane live in a clock domain
that only the dataplane can read, so everything above it substitutes the host
wall clock. Each worker anchors its clock to `CLOCK_REALTIME` once at startup
and derives every later timestamp from the cycle counter, never re-reading
realtime ([dataplane/worker.c#L369](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/dataplane/worker.c#L369),
[lib/dataplane/time/clock.c#L8-L29](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/lib/dataplane/time/clock.c#L8-L29)).
Nothing outside that process can obtain a value in the same domain.

The substitution is already visible at three levels above shared memory:

- The control plane decides whether a firewall-state layer has expired by
  comparing dataplane-stamped deadlines against the host's realtime clock
  ([map_service.go#L531](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/objects/fwstate/controlplane/map_service.go#L531)),
  and filters expired entries the same way
  ([#L565](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/objects/fwstate/controlplane/map_service.go#L565)).
- It exports a remaining time-to-live by subtracting a realtime timestamp from a
  dataplane deadline, with a comment on the line above stating the deadline is on
  the dataplane's clock
  ([#L266-L273](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/objects/fwstate/controlplane/map_service.go#L266-L273)).
- Raw dataplane timestamps reach clients unconverted
  ([convert.go#L25](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/objects/fwstate/controlplane/fwstatemappb/v1/convert.go#L25)),
  and the web page computes an entry's age as browser-wall-clock minus that value
  ([FWStatePage.tsx#L357-L363](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/modules/fwstate/web/FWStatePage.tsx#L357-L363))
  and renders it as an absolute UTC time
  ([#L473](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/modules/fwstate/web/FWStatePage.tsx#L473)) — by then the comparison
  happens in a browser on another machine.

The domains agree while the host's realtime clock is left alone, and diverge the
moment it is corrected — an NTP step, an administrator setting the date, a
virtual machine resuming — because the worker clocks do not follow such a
correction and everything above them does. Reported ages and lifetimes are then
wrong, and an expiry judgement made in the substituted domain can discard live
state. Every consumer that has to reason about dataplane time either repeats the
substitution or invents its own workaround.

## Outcome

A process attached to a shared-memory zone can read that instance's current time
in the dataplane's own clock domain, and can tell when no such time is available
yet, so a deadline or a stamped time read out of shared memory is judged and
displayed without crossing clock domains.

## Evidence

Established while filing, at `9d92fa95`.

The time is already published, on every round:

- A worker writes its current time as the first statement of its round, before
  any packet work and before the early return taken while no configuration is
  installed ([worker.c#L259-L266](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/dataplane/worker.c#L259-L266),
  [round.h#L24](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/lib/dataplane/worker/round.h#L24)), so idle rounds refresh it
  like any other.
- The round loop is unconditional and never sleeps
  ([worker.c#L309](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/dataplane/worker.c#L309)), so the published value trails
  the true time by one round.
- The store is a plain assignment today ([round.h#L24](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/lib/dataplane/worker/round.h#L24)),
  written for a single-writer, no-reader field.

Reading it from an attached process needs no new mechanism:

- Control-plane code already walks the worker array across the mapping and reads
  `dp_worker` fields, relative pointers and all
  ([counters.c#L643-L671](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/lib/controlplane/agent/counters.c#L643-L671)), and
  the instance's worker count is already exposed this way
  ([agent.c#L108](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/lib/controlplane/agent/agent.c#L108),
  [api/agent.h#L143](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/api/agent.h#L143)).
- fwstate's control plane already reaches an instance's configuration through
  that path ([map_service.go#L703-L708](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/objects/fwstate/controlplane/map_service.go#L703-L708)).

The per-worker clocks are one domain in practice, and the dataplane already
relies on that:

- Workers are created only during startup — worker init is reached solely from
  device init ([device.c#L148](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/dataplane/device.c#L148)), itself reached only
  from the device loop in dataplane init
  ([dataplane.c#L338](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/dataplane/dataplane.c#L338)) — so every anchor is taken
  inside one startup window and no worker can anchor after a later realtime step.
- Entries stamped by one worker are expired by another against that other
  worker's clock, so any cross-worker divergence is already load-bearing inside
  the dataplane.

Converting cycles to nanoseconds outside the dataplane is the part that does not
travel:

- Establishing an anchor reads the counter frequency from the environment
  abstraction layer ([clock.c#L14](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/lib/dataplane/time/clock.c#L14)), which
  only the dataplane process initialises.
- On 64-bit ARM the counter the dataplane reads is a build option: the generic
  timer by default, the performance-monitor cycle counter under
  `RTE_ARM_EAL_RDTSC_USE_PMU`
  (`subprojects/dpdk/lib/eal/arm/include/rte_cycles_64.h`), so a counter read
  reproduced elsewhere can silently disagree with the anchor it is applied to.
- The control plane's CGO build links the control-plane, dataplane-config,
  counters, logging and device-API archives and no part of the environment
  abstraction layer
  ([controlplane/ffi/agent.go#L3-L10](https://github.com/yanet-platform/yanet2/blob/9d92fa95ac35ec5fd87195c380cbeded89b267e8/controlplane/ffi/agent.go#L3-L10)).

## Scope

The read path from an attached process to the time the dataplane already
publishes, and the control-plane surface that exposes it. Consumers are
converted separately; this issue delivers the capability.

Agreed direction: publish what the workers already compute rather than
reproducing the cycle-counter conversion outside the dataplane. The instance's
time is the latest value any of its workers has published — that is robust to a
worker stuck in a driver call, never moves backwards, and is as correct as the
dataplane's own cross-worker expiry checks. An instance whose workers have
published nothing reports that no time is available, rather than resolving to a
host-clock value.

Out of scope: whether the dataplane clock should be re-anchored or slewed
against realtime, and whether the per-worker anchors should be collapsed into
one, both of which are #1754. This capability holds under either outcome.

## Acceptance

- [ ] A process attached to a zone reads that instance's current time in the
      dataplane's domain without initialising the environment abstraction layer.
- [ ] The value agrees with what a worker of that instance is using for its own
      expiry decisions, to within the length of a round.
- [ ] An instance that has published no time is reported as such, distinguishably
      from a valid time, and never as a host-clock value.
- [ ] The value does not jump when the host's realtime clock is stepped after the
      dataplane has started.
- [ ] The value never moves backwards across reads, including when one worker of
      the instance stops advancing.
- [ ] The publication and the read of the shared field are defined against each
      other rather than relying on the width of the store.

OrIX219:
Timestamps and deadlines in an instance's shared memory are on the dataplane's clock, which is anchored to real time once at startup and advanced from the cycle counter afterwards. Nothing outside that process could read a value in the same domain, so every consumer substituted the host clock and diverged from the dataplane whenever the host clock was corrected.

- Report an instance's current time as the latest its workers have published: that trails processing by at most one round, cannot be held back by a worker stopped inside a driver call, and never moves backwards.
- Report the absence of a time distinguishably, so a caller with nothing to compare against says so instead of reaching for the host clock.
- Publish the per-round worker time atomically, now that it is read from outside the process that writes it.

Closes #2450

### DIFF
--- a/api/agent.h
+++ b/api/agent.h
@@ -142,6 +142,15 @@ dataplane_instance_numa_idx(struct dp_config *dp_config);
 uint32_t
 dataplane_instance_worker_count(struct dp_config *dp_config);
 
+// Returns the current time of the dataplane instance, in nanoseconds.
+//
+// @param dp_config Handle to the dataplane instance.
+// @param time_ns Set to the current time in nanoseconds on success.
+//
+// @return 0 on success, -1 when the instance has published no time yet.
+int
+dataplane_instance_current_time(struct dp_config *dp_config, uint64_t *time_ns);
+
 // Detaches a module agent from shared memory, releasing associated resources.
 //
 // @param agent Handle to the module agent to detach
--- a/controlplane/ffi/shm.go
+++ b/controlplane/ffi/shm.go
@@ -131,6 +131,15 @@ func (m *DPConfig) WorkerCount() uint32 {
 	return uint32(C.dataplane_instance_worker_count(m.ptr))
 }
 
+func (m *DPConfig) CurrentTime() (uint64, bool) {
+	var timeNS C.uint64_t
+	if C.dataplane_instance_current_time(m.ptr, &timeNS) != 0 {
+		return 0, false
+	}
+
+	return uint64(timeNS), true
+}
+
 // Modules returns a list of dataplane modules available.
 func (m *DPConfig) Modules() []DPModule {
 	ptr := C.yanet_get_dp_module_list_info(m.ptr)
--- a/lib/controlplane/agent/agent.c
+++ b/lib/controlplane/agent/agent.c
@@ -109,6 +109,43 @@ dataplane_instance_worker_count(struct dp_config *dp_config) {
 	return (uint32_t)dp_config->worker_count;
 }
 
+int
+dataplane_instance_current_time(
+	struct dp_config *dp_config, uint64_t *time_ns
+) {
+	if (dp_config == NULL || time_ns == NULL) {
+		return -1;
+	}
+
+	uint64_t worker_count = dp_config->worker_count;
+	struct dp_worker **workers = ADDR_OF(&dp_config->workers);
+	if (workers == NULL) {
+		return -1;
+	}
+
+	uint64_t latest = 0;
+	for (uint64_t idx = 0; idx < worker_count; ++idx) {
+		struct dp_worker *worker = ADDR_OF(&workers[idx]);
+		if (worker == NULL) {
+			continue;
+		}
+
+		uint64_t published = __atomic_load_n(
+			&worker->current_time, __ATOMIC_RELAXED
+		);
+		if (published > latest) {
+			latest = published;
+		}
+	}
+
+	if (latest == 0) {
+		return -1;
+	}
+
+	*time_ns = latest;
+	return 0;
+}
+
 // Body of agent_free_unused_agents for a caller that already holds
 // cp_config_lock. See the definition further down for details.
 static void
--- a/lib/controlplane/agent/agent_test.c
+++ b/lib/controlplane/agent/agent_test.c
@@ -23,6 +23,195 @@
 #define TEST_CP_MEMORY (1 << 20)
 #define TEST_STORAGE_SIZE (TEST_DP_MEMORY + TEST_CP_MEMORY)
 
+// Publish a worker array on the instance, mirroring the registration the
+// dataplane performs at startup.
+//
+// The array and the workers are allocated from the instance's own memory so
+// the relative pointers a reader follows resolve the same way they do in a
+// live segment.
+static int
+publish_workers(
+	struct dp_config *dp_config, const uint64_t *times, uint64_t count
+) {
+	struct dp_worker **workers = (struct dp_worker **)memory_balloc(
+		&dp_config->memory_context, sizeof(struct dp_worker *) * count
+	);
+	if (workers == NULL) {
+		return -1;
+	}
+
+	for (uint64_t idx = 0; idx < count; ++idx) {
+		struct dp_worker *worker = (struct dp_worker *)memory_balloc(
+			&dp_config->memory_context, sizeof(struct dp_worker)
+		);
+		if (worker == NULL) {
+			return -1;
+		}
+
+		memset(worker, 0, sizeof(struct dp_worker));
+		worker->idx = idx;
+		worker->current_time = times[idx];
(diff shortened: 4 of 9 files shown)


OrIX219:
@codex review

GeorgyKirichenko (approved):

GeorgyKirichenko (approved):

PR #2484, opened 2026-09-02
perf(dataplane): resolve pipeline stage hops at publish time

GeorgyKirichenko:
Stacked on #2449 — merge that first (and #2441 beneath it).

- Chains, functions and the weighted chain and pipeline maps expose absolute addresses to the packet hot path: creation writes relative ones and the publishing process turns them absolute when a generation is assigned to workers.
- Chain and pipeline tails keep a controlplane-owned relative array for the free and link paths, while the weighted maps are recoded in place from the relative arrays and the controlplane weights.
- Recomputation on every assignment re-derives everything from controlplane-owned data, staying correct across dataplane restarts and re-publications.


GeorgyKirichenko:
absorbed be #2449

PR #2486, opened 2026-09-02
fix(operators): restore default config filenames

3Hren:
- Rename packaged generic operator role and forward payload examples to default-suffixed paths.
- Keep unsuffixed paths available for deployment-managed configurations.
- Cover default naming and template instance installation in packaging tests.


3Hren:
@codex review

PR #2490, opened 2026-09-02
perf(filter): store line attributes in a single memory chunk

GeorgyKirichenko:
Stacked on #2484 — merge that first (and #2449, #2441 beneath it).

- Vlan, device, port and proto-range attributes build and query a vline, a single-dimensional value table whose values live in one contiguous memory chunk instead of the chunked pointer array.
- The per-packet lookup dereferences the chunk directly, and the device bound check reads the line size.
- Multi-dimensional attributes keep the value table.

### DIFF
--- a/common/value.h
+++ b/common/value.h
@@ -165,3 +165,101 @@ value_table_compact(
 		}
 	}
 }
+
+/*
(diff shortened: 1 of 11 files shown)


GeorgyKirichenko:
absorbed by #2521

PR #2494, opened 2026-09-03
perf(dataplane)!: absolutize pipeline stage hops at publish time

GeorgyKirichenko:
Stacked on #2441 — merge that first.

- Every stage counter, the chains, pipelines, functions, modules and device entries each stage walk, the module device index tables, the module generation back-pointer and storages, and every object link's object context are absolute addresses for the packet hot path, resolved by the publishing process from the counter registry ids and the controlplane-owned relative arrays when a generation is assigned to workers.
- The relative arrays stay authoritative for the free and link paths and as derivation sources, so repeated passes stay correct across dataplane restarts and re-publications.
- Zero-length stage arrays skip their clearing to keep the undefined NULL memset out of the controlplane install path.
- The schedule worklists are provably built before any packet is scheduled, so the per-packet readiness check is gone and the absolute-address contract of the schedule rlist is documented on the list itself.
- Bumps YANET_MODULE_ABI_VERSION to 25.

### DIFF
--- a/common/rlist.h
+++ b/common/rlist.h
@@ -5,6 +5,11 @@
 // A member embeds a node and is reached back through the node; a head is
 // a plain node that points at itself when empty. Insertion appends at
 // the tail, so first-out order matches insertion order.
+//
+// A node embedded in shared memory keeps absolute addresses in its
+// links: only the process that builds the list ever writes or walks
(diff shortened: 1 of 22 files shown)


GeorgyKirichenko:
@codex review

PR #2498, opened 2026-09-03
fix(fwstate,cli): render config timeouts in milliseconds

OrIX219:
- Timeouts in the human view of a config are shown in milliseconds instead of the stored nanoseconds, which were unreadable at the scale connection lifetimes are set at. Requests and the JSON output keep the nanoseconds the wire and the stored configuration use.
- A timeout shorter than a whole millisecond keeps its remainder rather than truncating: zero is how these fields spell unset, so a live timeout must never be rendered as one.
- The human view of a config and of the config list no longer prints a JSON document regardless of the requested format: a config renders as a settings table and the list as one name per line.
- Fields the module documents as ignored are left out of the human view; the JSON output still carries the whole message.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3352,7 +3352,7 @@ dependencies = [
  "log",
  "prost",
  "serde",
- "serde_json",
+ "tabled",
  "tonic",
  "tonic-prost",
  "tonic-prost-build",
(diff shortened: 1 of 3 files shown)


OrIX219:
@codex review

OrIX219:
@codex review

GeorgyKirichenko (dismissed):

PR #2503, opened 2026-09-03
build(counters): build rure from a submodule instead of a wrap

OrIX219:
subprojects/regex is a git submodule pinned to the commit the wrap named, so it arrives with the rest of the tree the way dpdk and libpcap already do. A repository consuming yanet2 as a submodule could not fetch it otherwise: meson cloned it anonymously during configure, and that clone failed in CI unless a credential was threaded into the build.

lib/counters builds rure from that checkout with a cargo target of its own, since meson drives the crate neither as a meson nor as a cmake subproject. The wrap and its packagefiles overlay are gone with it, and so is the dependency lookup that resolved one from the other.

The archive now lands beside the other counters libraries rather than under subprojects, so the cgo link lines that name it follow it there.


OrIX219:
@codex review

OrIX219:
@codex review

OrIX219:
@codex review

GeorgyKirichenko (approved):

PR #2509, opened 2026-09-04
fix(fwstate): require sync settings only when configuring sync

### LINKED ISSUE #1758: fix(fwstate): UpdateConfig rejects a bare create by requiring sync fields unconditionally
## Motivation

`FWStateService.UpdateConfig` (`modules/fwstate/controlplane/service.go`) validates sync fields unconditionally: verified against current HEAD (`bb3dc1cd`), the function rejects the request with `InvalidArgument` if `req.SyncConfig == nil` or `req.MapConfig == nil` before it even looks at whether this is a first-time create or an update of an existing config, and later calls `validateSyncConfig(syncConfig)` (`service.go:267`) which requires four sync fields to be populated. So a bare first `fwstate create` is rejected unless four sync fields are supplied.

This was the root cause of the `fwstate list` empty-result hint being unusually wide (39 characters wider than any sibling module's create hint, none of which need more than a bare `--name`) — because the hint has to name all four required flags (`update --name <name> --sync-...`). A generic wrap fix already shipped for the hint's rendering (hanging-indent wrap for all hints), so the fwstate hint no longer breaks visually even at its current width — but the underlying defect this issue tracks is the validation itself, not the hint's rendering, and it is still live.

If `validateSyncConfig` became conditional on whether this is a create vs. an update of sync configuration, the hint would collapse to the same shape as every other module's (`update --name <name>`), and a first-time user creating an fwstate config without sync options configured would stop hitting a confusing validation error on their very first command.

This is server-side correctness, not CLI polish.

## Acceptance

- [ ] `fwstate`'s `UpdateConfig` only requires the four sync fields when actually setting sync configuration, not unconditionally on every create/update.
- [ ] A bare `fwstate create --name <name>` (no sync flags) succeeds.
- [ ] The `fwstate list` empty-result hint shortens to match the sibling-module shape (`update --name <name>`).

Source: migrated from the local task tracker's TASK-0168, filed 2026-07-29, migrated 2026-08-06.

OrIX219:
- Synchronization is optional: a config naming no part of the sync destination is accepted and installs a module that matches no sync packet, so a first create needs nothing but a name. Naming only part of that destination is still refused, since such a module would look configured while matching nothing.
- The linked maps are optional the same way. An unnamed map keeps the link the config already has, and one never named declares no link at all, which the module answers by counting and dropping that family's synced state until a later update names its map.
- An update that leaves out the sync settings or the map names no longer resets them: the settings were reset to the defaults and the links dropped, where the request had asked for no change to either.
- An address stated at a width other than an IPv6 one is refused instead of being read as far as it goes.
- The dataplane claims no packet while no sync destination is configured, so an unconfigured module cannot be made to insert state by a packet addressed to the unset destination.
- The list hint of the module CLI names the flags a create actually needs, and the client-side requirement for both map flags is gone.

### DIFF
--- a/modules/fwstate/controlplane/service.go
+++ b/modules/fwstate/controlplane/service.go
@@ -92,10 +92,6 @@ func NewMetricsFactory(extra ...grpcmetrics.Option) grpcmetrics.Factory {
 const (
 	// moduleType is the registered shared-memory type for fwstate configs.
 	moduleType = "fwstate"
-
-	// maxSyncPort is the highest value accepted for port_multicast,
-	// matching the width of the C-side uint16 port field.
-	maxSyncPort uint32 = 65535
 )
 
 // FWStateServiceName and MetricsServiceName are the fully-qualified gRPC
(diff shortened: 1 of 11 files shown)


modules/fwstate/controlplane/service.go
```diff
@@ -447,39 +452,69 @@ func (m *FWStateService) unpublishConfig(name string) {
...
-		return status.Errorf(codes.InvalidArgument, "missing required sync config fields: %v", missing)
+// validateSyncConfig checks the sync settings a construction would
+// install: the request already merged over the config it replaces.
+//
+// Synchronization is optional. A config naming no part of the sync
+// destination leaves the module a passthrough that matches no sync
+// packet, which is what a first create carrying no sync settings
+// installs. Naming only part of it is refused: such a module would look
+// configured while matching nothing.
+func validateSyncConfig(cfg *fwstatepb.SyncConfig) error {
```
3Hren: func (m *SyncConfig) Validate() error

modules/fwstate/controlplane/update_test.go
3Hren, line 1: No separate tests. Place test cases in a file `<something>_test.go` when you use entities from `<something>`.

modules/fwstate/tests/dataplane/fwstate.go
```diff
@@ -132,6 +132,16 @@ func SetSyncSuppressTimeout(cpModule *C.struct_cp_module, ns uint64) {
 	m.sync_config.sync_suppress_timeout = C.uint64_t(ns)
 }
 
+// ClearSyncDestination leaves the module config without a sync
+// destination, the state a config created with no sync settings carries.
+func ClearSyncDestination(cpModule *C.struct_cp_module) {
+	m := (*C.struct_fwstate_module_config)(unsafe.Pointer(cpModule))
+	m.sync_config.port_multicast = 0
+	for idx := range 16 {
```
3Hren: memset?

3Hren (changes requested):

GeorgyKirichenko (approved):

PR #2519, opened 2026-09-06
fix(cli): preserve errors from broken connections

3Hren:
- Return normal connection or RPC errors when a server closes a TCP or Unix socket instead of silently exiting with status 141.
- Preserve standard SIGPIPE behavior for a closed stdout pipe and retain vectored writes under backpressure.
- Cover closed sockets and partial writes with regression tests on Rust 1.88.


3Hren:
@codex review

PR #2520, opened 2026-09-06
feat(cli): compact the inspect report

3Hren:
- Replace the tree dump with a compact report sharing aligned names and properties across memory, devices, pipelines, functions and modules.
- Expand memory contexts with `--memory`, aligning `total` and `own` under a single root that includes pool usage and process metadata.
- Use muted tree connectors and memory-pressure colors, wrap to terminal width, and preserve the full JSON response and ASCII tree connectors when colors are disabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3384,8 +3384,9 @@ version = "0.1.0"
 dependencies = [
  "bytesize",
  "clap",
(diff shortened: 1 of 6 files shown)


3Hren:
@codex review

PR #2544, opened 2026-09-07
perf(dataplane): run scheduled device entries first and poll the rest after

### LINKED ISSUE #2541: perf(dataplane): do not poll device entry on each worker loop
Device entry polling traverse the whole pipeline and consumes CPU cycles. Also an empty pipeline entry will be polled in cases when there will be a packet to schedule on later.

The idea is to process non empty device entries only then return to them that was not touched in the worker loop round and process them only there will not be touched for at least 1 millisecond.

GeorgyKirichenko:
- The generation carries one device-entry home list and a ready list; a round drains the home list onto a stack list of untouched entries at constant cost through a new list concatenation routine.
- The scheduling path moves an entry onto the ready list from wherever it sits, and the round works each ready entry back to the home list before running it, so packets run to exhaustion first and rerouted entries rejoin the ready tail within the same round.
- When ready is exhausted the whole untouched set joins ready at constant cost and runs through the same loop, so periodic device work such as trafgen generation keeps pacing itself on being called every round and anything a poll schedules returns to ready iteration.
- Bumps YANET_MODULE_ABI_VERSION to 26 for the generation worklist layout change.

Closes #2541.

### DIFF
--- a/common/rlist.h
+++ b/common/rlist.h
@@ -44,3 +44,24 @@ static inline int
 rlist_empty(struct rlist *head) {
 	return head->next == head;
 }
+
+// Move every item of one list onto the tail of another.
+//
+// The source head is left empty and neither list is walked, so the
+// move costs the same regardless of lengths.
+static inline void
+rlist_concat(struct rlist *dst, struct rlist *src) {
+	if (rlist_empty(src)) {
+		return;
+	}
+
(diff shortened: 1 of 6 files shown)


GeorgyKirichenko:
@codex review

PR #2547, opened 2026-09-07
feat(l3b): add the l3b load-balancer module

GeorgyKirichenko:
- Adds the l3b IP-in-IP load-balancer module: module-level destination filters classify initial-fragment TCP/UDP traffic into virtual services, per-service source filters gate it, and a weighted round-robin ring selects a real server to encapsulate the packet towards, deriving the outer source address from the real server's source network; one-packet scheduling drives the ring slot from the module link's per-worker packet counter.
- Virtual services are standalone named shared objects under objects/l3b, linked into module configurations by the service names their destination rules carry through cp_module_link_object and resolved per worker at execution-context build time; real-server weight (clamped to 1000) and state updates mutate the published objects in place, disabled servers leave the rebuilt ring so their share shifts to healthy backends, and a service update adopts the live session table so pinned flows keep their backend across list changes.
- Every service owns a session table — a layered statemap table in a second cp_object — pinning each client flow to its backend by destination address with a TCP-flag-aware lifetime policy that refreshes on activity and re-pins UDP flows and connection setups only; pure L3 balancing keeps the client's ports out of the session identity, the balancer answers ICMP echo requests for service addresses in place, SYN packets carry a clamped MSS, and the outer header's DSCP is marked per a never/only-default/always policy.
- Service objects carry per-worker counters — incoming, filter-rejected, ring-empty, disabled-real drop reasons, generated echo replies, per-real throughput and the link packets feeding one-packet scheduling — scraped under the object's tags; ListSessions pages through the pinned flows and GetService inspects the live scheduler masks and real servers.
- Extracts the generic layered hash map core into lib/statemap with a startup callback registry, and builds lib/l3state for the flow-state primitives, leaving fwstate with only its firewall-specific pieces.
- Ships the control plane — L3bService gRPC, Go bindings, a Rust CLI with YAML-driven service and module-config documents — and end-to-end functional tests covering the filter, session, ring and encapsulation paths, one-packet scheduling, echo replies, MSS and DSCP policies, plus ports of the first-generation balancer autotests: wrr, tcp/udp, real IPv4, outer-source-network, acl-balancer-route and metabalancer-decap.


GeorgyKirichenko:
All 9 review findings addressed in ee5b764a:

- The session table is now published in the same agent_update_objects call as its service (the hard-coded count of 1 dropped the table from the registry, which also wedged DeleteVirtualService).
- Real-server state updates rebuild the scheduler ring over the enabled servers only — a disabled backend's share shifts to the healthy ones instead of blackholing new flows — and weight updates respect the enabled set too.
- Real-server indexes are validated on the Go side against the configured count before crossing into C, surfacing InvalidArgument instead of Internal.
- The keep-old promotion callback had the typedef's argument order swapped (new first, stale second), so a layer rotation installed the newly proposed backend instead of the pinned one.
- Malformed service payloads (bad address lengths, family mismatches, bad filter rules) return InvalidArgument instead of Internal.
- ListSessions caps a requested page at 100000 records so a huge limit cannot translate into a huge allocation.
- The encapsulation path reads the real-server state with an atomic acquire load, pairing with the controlplane's release store.
- Session listings copy each key/value pair twice and keep it only when both reads agree, so a concurrent re-pin yields either the old or the new record, never a torn mix.
- The default session table index dropped to 64K slots and the default agent arena rose to 64 MiB so a default service with its table fits.

GeorgyKirichenko:
ICMP echo replies for service addresses added (6671a61c), mirroring the first-generation balancer:

- A destination rule matching ICMP echo (proto 1 / 58 with subtype = echo-request type, expressible as `icmp`/`icmp6` with an exact subtype in CLI documents) routes the request to its service, which answers in place: reply type, swapped source/destination, TTL restarted at 64, IPv4 header checksum recomputed and ICMP/ICMPv6 checksums updated incrementally — no session, scheduler, tunnel or source filter involved.
- An echo request no rule claims is forwarded untouched (a ping to some other address), preserving the old pass-through semantics.
- The reply is observed through the service object's `icmp_replied` [packets, bytes] counter alongside `incoming`; functional test verifies the reply's addresses, lack of tunneling and both counters.

GeorgyKirichenko:
PR refreshed: body rewritten to cover the full branch, and the rebase onto the external-round generation wait had silently dropped two earlier review fixes during conflict resolution — both restored in 54c0af55: the ListSessions page cap (max 100000 records, so a huge limit cannot translate into a huge allocation) and the tear-resistant session reads (each key/value pair copied twice, kept only when both reads agree, so a concurrent re-pin yields either the old or the new record). All other findings remain verified fixed in the tree.

PR #2549, opened 2026-09-08
fix(fwstate,cli): prevent concurrent update reverts

### LINKED ISSUE #2442: fix(fwstate,cli): update can silently revert a concurrent change
## Motivation

`yanet-cli fwstate update` offers partial-update semantics (set only the flags you pass, keep the rest) and implements them client-side as a read-modify-write: it reads the stored config with `ShowConfig`, merges the passed flags into the reply and publishes the merged document with `UpdateConfig`. Nothing ties the write to the read — no version, no precondition — and the merged request carries every stored field as an explicit value. Between the two RPCs any other writer (a second `update` invocation, an operator apply) can publish its own change, which the first writer's echoed stale values then overwrite. Both writers report success; the observed result is a config neither of them asked for, and nothing signals that a change was reverted.

The window is human-scale on a hand-operated stand, but the failure mode is silent config corruption, and the pattern breeds secondary defects in the read leg (#2338 was one).

## Evidence

Code-proven.

- The CLI read-modify-write, with no concurrency token: read and merge — https://github.com/yanet-platform/yanet2/blob/ca873215b0bba1d0eb5f4880574a89a3b962607d/modules/fwstate/cli/src/main.rs#L157-L176, publish — https://github.com/yanet-platform/yanet2/blob/ca873215b0bba1d0eb5f4880574a89a3b962607d/modules/fwstate/cli/src/main.rs#L220-L231.
- The server already merges field-by-field under its mutation lock: a zero or empty sync field in the request keeps the stored value — `ToCWithDefaults`, https://github.com/yanet-platform/yanet2/blob/ca873215b0bba1d0eb5f4880574a89a3b962607d/modules/fwstate/controlplane/fwstatepb/v1/convert.go#L28-L67, applied from `UpdateConfig` https://github.com/yanet-platform/yanet2/blob/ca873215b0bba1d0eb5f4880574a89a3b962607d/modules/fwstate/controlplane/service.go#L193-L253. Concurrent updates that sent only their flagged fields would therefore commute; it is the CLI echoing stored values as explicit fields that reintroduces the lost update.
- The pre-flight read is genuinely needed only for the map names, which the wire requires non-empty on every update — https://github.com/yanet-platform/yanet2/blob/ca873215b0bba1d0eb5f4880574a89a3b962607d/modules/fwstate/controlplane/fwstatepb/v1/fwstate.proto#L29-L37.
- `ShowConfigResponse` carries nothing an update could use as a precondition — https://github.com/yanet-platform/yanet2/blob/ca873215b0bba1d0eb5f4880574a89a3b962607d/modules/fwstate/controlplane/fwstatepb/v1/fwstate.proto#L46-L51.
- A side effect of the same wire contract: an explicit zero is indistinguishable from "keep", so clearing a stored field back to zero is inexpressible through `update`.

Related: #2338 (the read leg swallowed errors, fixed independently), #1758 (server field requirements on a bare create shape what a first `update` must send).

## Decision points

The update contract itself is unresolved:

- Does `yanet-cli fwstate update` keep a client-side merge at all, lean on the server's zero-means-keep merge, or become a whole-document upsert?
- Whichever layer owns partial semantics, what guarantee makes a read-dependent write safe against a concurrent writer, and how does a user express "set this field to zero"?

## Scope

`yanet-cli-fwstate update` semantics and the `fwstatepb` show/update contract they rest on. The fwstatemap CLI and the sync dataplane are out of scope.

## Acceptance

- [ ] A recorded decision on the update contract: whether partial updates stay, which layer owns them and what concurrency guarantee an update carries.
- [ ] Under the chosen contract, an update that did not observe a concurrent change can no longer silently revert it.
- [ ] The affected contracts (CLI arguments, wire messages) match the decision.

sakateka:
- Makes fwstate CLI updates send only explicitly supplied fields.
- Merges masked updates under the service mutation lock, preserving independent map, endpoint, and timeout changes.
- Keeps legacy unmasked requests and explicit zero, empty, and endpoint-clear semantics compatible with current fwstate validation.
- Closes #2442.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -3,6 +3,27 @@
 Manual steps required when upgrading an existing deployment. The packaging
 does not perform these automatically.
 
+## FWState CLI partial updates
+
+Upgrade the fwstate control plane before using the new CLI. The CLI now sends
+only explicitly supplied settings with an update mask, without reading the
+stored configuration first. The service merges these settings under its
+mutation lock, so updates to different fields preserve each other; writes to
+the same field use the last successfully published value.
+
+An omitted flag preserves its field. An explicit zero timeout writes zero;
+an empty map-name flag unlinks that family. The resulting configuration must
+still pass validation. A present empty mask preserves every field and creates
+a missing configuration with defaults.
+
+Requests without a mask retain their legacy merging, destination-set replacement
+and endpoint-clear flags. Masked updates preserve the unselected endpoint;
+the CLI encodes `--no-multicast` and `--no-unicast` as masked clears.
+Do not combine a mask with the legacy endpoint-clear flags.
+Older servers ignore the new mask, so clears and endpoint updates can behave
+incorrectly. Older clients and the web configuration editor still send stored
+values and do not gain protection against stale-document overwrites.
+
 ## Controlplane systemd unit became a template
 
 `yanet2-controlplane.service` is replaced by the template
--- a/modules/fwstate/cli/src/args.rs
+++ b/modules/fwstate/cli/src/args.rs
@@ -30,7 +30,7 @@ pub enum ModeCmd {
     List,
     /// Delete a fwstate configuration.
     Delete(DeleteCmd),
-    /// Update fwstate configuration (map and sync settings).
+    /// Create or update only the supplied map and sync settings.
     Update(UpdateCmd),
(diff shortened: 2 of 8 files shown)


sakateka:
@codex review

sakateka:
@codex review

PR #2560, opened 2026-09-09
perf(dataplane): interleave packet rx and tx

### LINKED ISSUE #2527: perf(dataplane): interleave packet rx/tx
Just after receiving packet contents is far out of CPU caches what leads to huge amount of stale cycles on packet parsing. The idea is to read new packet batch and issue prefetch just before transmitting of previous one.

GeorgyKirichenko:
- Receive the next batch and prefetch its cold lines just before transmitting the previous one; parsing waits for the next round, so the prefetches overlap a whole transmit instead of an in-burst lookahead of eight.
- The received batch is held raw in a per-worker staging array across the round boundary, so no configuration pointers outlive the generation acknowledgement; the batch is parsed and scheduled at the start of the next round under that round's snapshot.
- The no-config startup path drops the staged batch and still drains the ring every round; every consumer resets the staging count once the mbufs leave its hands.

Closes #2527.

### DIFF
--- a/dataplane/worker.c
+++ b/dataplane/worker.c
@@ -67,52 +67,49 @@
 
 #include <string.h>
 
-// Lookahead that hides the two cache misses every received packet costs.
+// Receive the next batch and warm its coldest lines ahead of the parse.
 //
 // Every received packet costs two cache lines nobody has touched lately:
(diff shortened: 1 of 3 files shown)


GeorgyKirichenko:
@codex review

PR #2566, opened 2026-09-10
refactor(cli): route the small config clis through the unary helper

3Hren:
- Replaces the hand-written client, map_err and into_inner chain in blackhole, decap, dscp, unrdup, nat64, route-mpls and pdump with Service::unary, and the NotFound closures with Service::not_found.
- Drops the per-crate request and response log lines, which the helper now emits uniformly, and the log dependency of the crates that had no other use for it.
- Behaviour is unchanged apart from the log level of the request line in nat64, which moves from debug to trace like everywhere else.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3178,7 +3178,6 @@ version = "0.1.0"
 dependencies = [
  "clap",
  "clap_complete",
- "log",
  "prost",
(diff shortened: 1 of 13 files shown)


3Hren:
@codex review

PR #2567, opened 2026-09-10
refactor(cli): route the module clis through the unary helper

3Hren:
- Replaces the hand-written client, map_err and into_inner chain in forward, mirror, acl, fwstate, route, balancer2 and fwstate-map with Service::unary, and route's NotFound closure with Service::not_found.
- Drops the per-crate request and response log lines, which the helper now emits uniformly, and the log dependency of acl and fwstate that had no other use for it.
- Streaming calls keep the direct client path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3135,7 +3135,6 @@ version = "0.1.0"
 dependencies = [
  "clap",
  "clap_complete",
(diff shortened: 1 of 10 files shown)


3Hren:
@codex review

PR #2571, opened 2026-09-10
refactor(cli): print list commands through the shared name renderer

3Hren:
- Every list command renders its names through display::print_names, so all of them sort the names and escape control characters instead of four different renderings (plain lines, ptree, a table, a YAML or JSON dump).
- fwstate drops its own escaping and its one-column table, balancer2 drops its ptree dependency.
- The JSON payloads are unchanged.


3Hren:
@codex review

PR #2582, opened 2026-09-10
fix(controlplane): serialize per-call agent attach in pipeline and function services

### LINKED ISSUE #2525: fix(controlplane): serialize per-call agent attach in pipeline and function services
## Motivation

`DeviceService` now serializes its per-call agent attach with a mutex (PR #2522, `controlplane/builtin/device.go`), because `agent_attach` supersedes any live agent of the same fixed name and, once the superseded agent's loaded counts read zero, frees its arena in the same call — a shared-memory use-after-free if the superseded agent is still in use by an overlapping call. `PipelineService` and `FunctionService` attach the same fixed agent name (`"pipeline"`, `"function"`) per RPC with no such serialization, so two overlapping calls on either service can free the arena the other call is still using.

Each of the three services also has a second, independent gap: every attach reserves a 1 MB arena that the delete/update paths never allocate from (the touched generation clone comes from the global config memory context, not the per-call agent arena), and `agent_detach` is a no-op, so the arena is reclaimed only when a later attach under the same name supersedes it. `pipeline.go` and `function.go` carry a comment claiming the memory "will be free after [the] corresponding call is done," which is not what happens.

## Evidence

- Pipeline and function attach the same fixed name per call with the stale comment, no mutex — https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/controlplane/builtin/pipeline.go#L18-L26, https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/controlplane/builtin/function.go#L18-L26.
- Device service's fix for the identical race, showing the pattern not yet applied to pipeline/function — https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/controlplane/builtin/device.go#L17-L35.
- `agent_attach` supersedes the same-name entry unconditionally under the config lock — https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/lib/controlplane/agent/agent.c#L320-L379.
- The reclaim guard frees a superseded agent once its loaded counts read zero, with the unenforced single-live-holder precondition already tracked separately by #2001 — https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/lib/controlplane/agent/agent.c#L1700-L1727.
- `agent_detach` is a no-op, so an arena is only reclaimed by a later supersede — https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/lib/controlplane/agent/agent.c#L544-L548.
- The delete/update path clones the generation from the global config memory context, not from the per-call agent's arena — https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/lib/controlplane/config/zone.c#L107, https://github.com/yanet-platform/yanet2/blob/c5628874e1315526de3a4729163aedc6b2279346/lib/controlplane/config/zone.c#L816-L845.

Surfaced by review of PR #2522, which fixed only the device service.

## Scope

`controlplane/builtin/pipeline.go` and `controlplane/builtin/function.go`. Whether the fix is a mutex per service (mirroring device.go) or one long-lived shared agent is the implementer's call. Out of scope: the general single-live-holder precondition tracked by #2001, and any change to `agent_attach`/`agent_detach` themselves.

## Acceptance

- [ ] Two overlapping RPCs on `PipelineService` (or `FunctionService`) can no longer free an arena the other call is still using.
- [ ] The per-call arena is either sized to what the call path actually allocates from it, or the attach stops reserving one it never uses.
- [ ] The `pipeline.go`/`function.go` comments describing memory reclamation match the actual lifetime.

egnees:
- Serialize the attach-through-call lifetime of the pipeline and function services, so two overlapping RPCs can no longer free the shared-memory arena the other call is still using, mirroring the device service.
- Drop the per-call reservation from a megabyte to a single page, since neither the update nor the delete path allocates from the agent's own arena: the touched generation is cloned from the global configuration pool.
- Replace the comments claiming the agent's memory is released once the call completes with the lifetime the code actually has.
- Cover both services with a concurrency test over overlapping updates and deletes, and with a test pinning that the reservation still leaves the arena a spare block, so the services do not surface as permanently full in the arena occupancy readout.

Closes #2525.

### DIFF
--- a/controlplane/builtin/function.go
+++ b/controlplane/builtin/function.go
@@ -3,6 +3,7 @@ package builtin
 import (
 	"context"
 	"errors"
+	"sync"
 
 	"github.com/c2h5oh/datasize"
 	"go.uber.org/zap"
@@ -17,17 +18,22 @@ import (
 
 const functionAgentName = "function"
 
-// Function agent is not persistent: it is created
-// on every call of update/assign/delete.
-// Memory, allocated for function agent, will be free after
-// corresponding call is done. So, on every call we need to allocate
-// memory for temporary operations only. For now, 1MB is
-// sufficient.
-const functionAgentMemory = datasize.MB
+// A token reservation: attaching refuses a zero size, and an arena that can
+// hand back no block at all is reported as fully occupied.
+//
+// An update or a delete clones the touched generation from the global
+// configuration pool, so no call ever allocates from the agent's own arena.
+// The arena outlives its call — detaching releases nothing — and is
+// reclaimed only when the next attach under the same name supersedes it.
+const functionAgentMemory = 4 * datasize.KB
 
 // Function is an in-process gRPC service for managing functions.
 type Function struct {
 	ynpb.UnimplementedFunctionServiceServer
+
+	// One attach-through-mutation lifetime at a time under the fixed agent
+	// name, because a concurrent attach reclaims the agent still in use.
+	mu sync.Mutex
 
 	instanceID uint32
 	shm        *ffi.SharedMemory
@@ -190,6 +196,9 @@ func (m *Function) Update(
 		function.Chains = append(function.Chains, functionChain)
 	}
 
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	agent, err := m.shm.AgentAttach(functionAgentName, m.instanceID, functionAgentMemory)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
@@ -219,6 +228,9 @@ func (m *Function) Delete(
 		return nil, status.Error(codes.InvalidArgument, "function name is required")
 	}
 	functionName := reqId.Name
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
 
 	agent, err := m.shm.AgentAttach(functionAgentName, m.instanceID, functionAgentMemory)
 	if err != nil {
--- a/controlplane/builtin/function_test.go
+++ b/controlplane/builtin/function_test.go
@@ -1,12 +1,16 @@
 package builtin_test
 
 import (
+	"fmt"
 	"testing"
 
+	"github.com/c2h5oh/datasize"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	dataplaneut "github.com/yanet-platform/yanet2/bindings/go/dataplane_ut"
 	commonpb "github.com/yanet-platform/yanet2/common/commonpb/v1"
 	"github.com/yanet-platform/yanet2/controlplane/builtin"
 	ynpb "github.com/yanet-platform/yanet2/controlplane/ynpb/v1"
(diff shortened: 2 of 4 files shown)


egnees:
@codex review

egnees:
Addressed all three findings in the amended commit.

- **P2, force the overlap.** The callers now open their windows together on one gate and repeat the round, at 32 callers x 64 rounds. Measured on this branch: with the serialization stripped out, both tests fail 10/10 runs (`cannot build execution context for zero workers`, i.e. the superseded agent read after its struct was reclaimed); with it in place, 20/20 green plus a clean `-race` pass. A seam that pins the window deterministically would need a test-only hook inside the attach path of the service, which seemed a worse trade than a barrier plus repetition.
- **P3, the one-byte arena.** Correct, and the previous wording was misleading: an eight-byte-aligned round of a one-byte span leaves the agent with no usable block at all. The comment now says so outright, so nothing later mistakes it for a usable minimum.
- **P3, the mutex comment.** Rewritten in both files to state the lifetime invariant without naming the field or restating the lock.

egnees:
@codex review

egnees:
Deep review round on top of the Codex pass found one defect in my own previous revision, now fixed in `047a6cc0`.

Shrinking the reservation to one byte was wrong in a way the compile and the tests did not show. `allocate_arenas` records a one-byte span and `block_allocator_put_arena` rounds both bounds to eight-byte alignment, so the agent ended up with no block at all: measured `limit=1, free=0`. Occupancy is reported as reserved minus free, so both services would have sat permanently at 100% in `yanet-cli inspect` (rendered in error style with a `!` past 95%), in the web inspect view and in `memory_arena_free_bytes`. A benign condition dressed as critical memory pressure.

The reservation is now a single page: still three orders of magnitude below the old megabyte, still allocated from by nothing, but a well-formed arena that reads 0%. Measured across sizes: `1B -> 100%`, `16B -> 0%`, `4KB -> 0%`, `1MB -> 0%`.

Added `Test_Pipeline_Update_LeavesArenaSpare` and the function counterpart to pin it; both fail deterministically at one byte and pass at a page. The concurrency tests are unchanged and still fail 10/10 with the serialization removed, 20/20 green with it.

Out of candidate, so left alone: `device.go` carries the same claim this PR removed from the other two, that its megabyte "covers a delete's temporary allocations". The delete path allocates from the global configuration pool there too, so that arena is equally untouched. Issue #2525 scoped device out; worth a follow-up.

egnees:
@codex review

egnees:
Round two of the deep review closes with no new findings, and the sanitizers now name the defect precisely rather than by its symptom.

Rebuilt this branch with `-Db_sanitize=address,undefined` and ran the package the way the `build-asan-test` job does. With the serialization removed:

```
ERROR: AddressSanitizer: use-after-poison on address 0x7f6233ab23d0
SUMMARY: AddressSanitizer: use-after-poison in agent_update_functions
```

That is the superseded agent being read after a concurrent attach handed its block back to the allocator, which poisons what it reclaims: exactly the use-after-free #2525 describes, and it reproduces deterministically. With the serialization in place the whole package is clean under ASan and UBSan, `LeavesArenaSpare` included, so the page-sized reservation still yields a block once red zones shift the arena base.

GeorgyKirichenko (approved):

PR #2583, opened 2026-09-10
feat(cli): filter and order the gateway listing, select inspect sections

### LINKED ISSUE #2367: feat(cli): gateway list ordering, filters and staleness, and inspect section filters
## Motivation

`yanet-cli gateway list` prints the registry in registry order, which the discovery module itself documents as unordered (and sorts for its own use), so two runs differ; it has no filter by kind or name; and an external backend that stopped heart-beating shows only an age — the operator has to know the heartbeat interval to judge it. `yanet-cli inspect` prints one tree of everything, with no way to select a section or an agent and no depth cap on the memory tree.

## Outcome

`gateway list` sorted, with `--kind`/`--suffix` filters and a `stale` column/field; `inspect --only <section>` / `--agent <name>` / `--depth N`.

## Evidence

- Registry order — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/cli/modules/gateway/src/main.rs#L90-L94; discovery sorts — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/cli/core/src/discovery.rs#L459-L471.
- Age only — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/cli/modules/gateway/src/main.rs#L149-L164.
- Single-tree inspect — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/cli/modules/inspect/src/main.rs#L24-L33.

## Scope

`cli/modules/gateway`, `cli/modules/inspect`.

## Acceptance

- [ ] `gateway list` output is stable across runs; filters and `stale` work in both renderings.
- [ ] `inspect` accepts section, agent and depth selectors.

egnees:
- `gateway list` settles the order of its own table instead of inheriting whatever the registry contract leaves open — nothing on the wire promises an order — so the same registry renders the same way on every run.
- `gateway list` takes `--kind` and `--suffix`, narrowing the table and the JSON payload alike; `--suffix` matches the trailing dot-separated segment, the same rule discovery already uses to recognise a service family, and completes from the segments the gateway currently serves.
- `gateway list` reports staleness beside the age, so an external backend that stopped registering is named as such instead of leaving the operator to know the registration lifetime. `--stale-after` moves that line, defaulting to the gateway's own registration lifetime; a kind that is not known to re-register keeps its age but is never judged, so a gateway too old to classify its backends does not turn every built-in stale.
- The JSON payload keeps the wire record and gains one `stale` field beside it.
- `inspect` takes `--only` to pick report sections and `--agent` to pick one agent's memory, both narrowing the JSON payload too; `--agent` completes from the agents the instance reports.
- `inspect` takes `--depth` to cap the printed memory context tree, which it expands without `--memory`. Subtree totals stay those of the whole tree, so a cut branch still accounts for everything below it.
- `ync` gains the raw age behind its age formatter, now kept to subsecond precision so a fractional threshold is judged on the real age, and exposes the service-name suffix helpers, so neither CLI copies them.

Closes #2367.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3382,7 +3382,10 @@ name = "yanet-cli-gateway"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "clap_complete",
+ "humantime",
  "prost-types",
+ "serde",
  "tabled",
  "tonic",
  "yanet-cli",
@@ -3395,6 +3398,7 @@ version = "0.1.0"
 dependencies = [
  "bytesize",
  "clap",
+ "clap_complete",
  "tonic",
  "unicode-segmentation",
  "unicode-width",
--- a/cli/core/src/discovery.rs
+++ b/cli/core/src/discovery.rs
@@ -245,11 +245,17 @@ pub async fn list_services(connection: &Connection, suffix: &str) -> Result<Vec<
     Ok(services)
 }
 
-/// Reports whether `name`'s last dot-separated segment is exactly `suffix`,
-/// so that a service merely mentioning it elsewhere in its name is not
-/// mistaken for a match.
-fn has_suffix(name: &str, suffix: &str) -> bool {
-    name.rsplit('.').next() == Some(suffix)
+/// Returns the last dot-separated segment of a service name, the segment that
+/// names the family it belongs to.
+pub fn suffix_of(name: &str) -> &str {
+    name.rsplit('.').next().unwrap_or(name)
+}
+
+/// Reports whether a service name's last dot-separated segment is exactly the
+/// one given, so that a service merely mentioning it elsewhere in its name is
+/// not mistaken for a match.
+pub fn has_suffix(name: &str, suffix: &str) -> bool {
+    suffix_of(name) == suffix
 }
 
 /// Resolves a short alias (e.g. `route`) against the discovered `services`.
@@ -428,6 +434,13 @@ mod test {
             format!("operators.pipeline.operatorpb.v1.{suffix}"),
             format!("operators.route.operatorpb.v1.{suffix}"),
         ]
+    }
+
+    #[test]
+    fn test_suffix_of_returns_the_trailing_segment() {
+        assert_eq!("ReadinessService", suffix_of("controlplane.ynpb.v1.ReadinessService"));
+        assert_eq!("Gateway", suffix_of("Gateway"));
+        assert_eq!("", suffix_of(""));
     }
 
     #[test]
--- a/cli/core/src/humanfmt.rs
+++ b/cli/core/src/humanfmt.rs
@@ -5,9 +5,31 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use prost_types::Timestamp;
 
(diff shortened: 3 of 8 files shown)


egnees:
@codex review

egnees:
Addressed all four findings:

- **Keep runtime RPCs on the logging wrapper** — both `gateway list` and `inspect` now issue their RPC through `Service::unary`, so `-vv` traces the request and the unwrapped response again. Completion lookups stay on the raw client.
- **Avoid declaring unspecified legacy backends stale** — the kind now answers two separate questions: whether an age is worth showing (external and unspecified, unchanged) and whether an age can settle staleness (external only). An unclassified backend keeps its age and gets no verdict, so a version-skewed gateway no longer reports every built-in as stale.
- **Make the future-timestamp contract match the code** — the age helper's contract now says a timestamp ahead of the local clock reads as a zero age, which is what it returns and what the test pins.
- **Correct the default eviction-window explanation** — the default is described as the gateway's registration lifetime, with the periodic sweep called out separately as the reason an entry can still be listed past it.

@codex review

egnees:
Second round addressed:

- **Preserve subsecond precision in stale checks** — the age helper no longer truncates either instant to whole seconds; it subtracts the full timestamp from the clock and saturates at zero, so a fractional `--stale-after` (`500ms`) is judged on the real age. The age formatter still renders whole seconds, unchanged. Tests now drive it from fixed instants, including the 10.1s-vs-9.9s boundary case.
- **Test staleness behavior instead of rendered cells** — the cell tests no longer pin `yes`/`no`/`5s`. They assert the invariants: a verdict renders differently from its absence, a kind that is never judged renders the not-applicable sentinel even when handed a verdict, and a kind that shows an age renders neither sentinel. The inspect depth test compares totals through the byte formatter rather than a literal string.
- **Correct the fixture's claimed agent count** — the fixture is described as carrying the agents it is given.

@codex review

egnees:
Third round addressed:

- **Correct the ordering rationale** — you are right, `BackendRegistry.ListBackends` has sorted by service name since #740, so the old rationale was false. The sort stays, because nothing in the wire contract promises an order and the CLI should not inherit one gateway build's behaviour, but the comment now says exactly that instead of claiming the registry comes back shuffled. The commit body and PR description are corrected the same way.
- **Remove internal identifiers from the age contract** — the age helper's doc now describes an observation and a local clock reading in domain terms, keeping the absent-value, subsecond-precision and clock-skew guarantees.

@codex review

egnees:
Fourth round addressed:

- **Remove fixture parameter names from its contract** — the registry-entry fixture is now described as returning an entry last registered a given number of seconds before the reference instant, or with no registration time at all, without naming its arguments.
- Swept the rest of the diff for the same defect while I was there: the suffix predicate whose visibility this change widens also named its arguments in prose, and now describes a service name's trailing segment instead.

@codex review

PR #2584, opened 2026-09-10
feat(operators,cli): neighbour cli file input, filters and readable ages

### LINKED ISSUE #2362: feat(operators): neighbour cli file input, filters and readable ages
## Motivation

`yanet-cli operator neighbour add` builds exactly one entry although `UpdateNeighbours.entries` is repeated, so seeding a table is one process per neighbour. `show` has no filter by state or device. The `AGE` column prints `Duration`'s debug form — a day-old entry is `86400.00s` — while `ync` already has `humanfmt::format_age` producing `1d 3h`. `add --help` does not say the entry is permanent.

## Outcome

Bulk add from a file, `show --state`/`--device` filters, human ages, and help that states the permanent semantics.

## Evidence

- Single entry per call — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/cli/neighbour/src/main.rs#L264-L274; proto — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/operatorpb/v1/neighbour.proto#L265-L268.
- Age rendering — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/cli/neighbour/src/main.rs#L429-L433; shared formatter — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/cli/core/src/humanfmt.rs#L18-L40.
- Server sets `PERMANENT` — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/internal/operator/service_neighbour.go#L174.

## Scope

`operators/route/cli/neighbour`. The verb rename (`add` → `insert`) is part of the dictionary migration.

## Acceptance

- [ ] A file of neighbours is applied in one request.
- [ ] `show --state` and `--device` filter the table; ages render like the rest of the suite.
- [ ] Tests cover file parsing and the filters.

egnees:
- Add a bulk form that reads static neighbours from a YAML file and applies the whole file in one request, so seeding a table no longer costs one process per neighbour; the file is the command's positional, the table stays a flag.
- Name in the file the same keys the operator's own seed configuration names, and spell out in the help which address is the neighbour's and which is the local interface's, so a neighbour list can be moved between the two without being rewritten.
- Expand a merge key, so a repetitive file states a shared interface or hardware address once instead of per neighbour.
- Read the file before the connection is made, so a missing, malformed or empty one is reported with its path and line rather than hidden behind an unreachable endpoint.
- Refuse a file key the operator owns, so a state, a timestamp or a source written into a file is reported instead of being discarded in silence; an unrecognised key fails the load naming the file and listing what is accepted.
- Land a batch of neighbours as one replacement map instead of one write per entry. The cache hands a reader the table itself and that reader walks it without a lock, so the previous write-per-entry could abort the operator with a concurrent map iteration and write; a listing overlapping an update now sees the batch whole or not at all.
- Raise the request ceiling an operator accepts from the 4 MiB gRPC defaults to, to the 256 MiB the gateway standing in front of every operator already admits. This is a shared constructor, so it moves every operator; the gateway already passes a message that size through, so the operator behind it was the only hop refusing one. A reply keeps the far larger ceiling gRPC gives it by default, since a listing repeats a table back with everything the operator stamps on each entry.
- Narrow a listing by state and by device, in the human table and in the JSON payload alike, with the state spelled exactly as the state column renders it and an undefined discriminant resolved the same way on both sides.
- Read an empty table name as no name, so `--table=` from an unset shell variable no longer changes the default table while reporting one called `''`.
- Read an entry the far side never stamped as a dash. Its zero instant arrives as a second count well before the epoch, not as zero.
- State on the cache primitive that writing a single entry lands it in the table outstanding readers are walking; the comment there previously advertised a safety guarantee the type does not have, which is how the abort came to be written.
- State in the help that an added entry is permanent, never ages out, and is overwritten by a later add for the same next hop.
- Cover the file schema, the wire mapping, both filters, the batch atomicity and the raised ceiling with tests, including a race-detector test for a listing overlapping an update, and a guard that every state a listing can report has a filter spelling that selects it.

Closes #2362.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2963,7 +2963,7 @@ dependencies = [
  "netip",
  "prost",
  "serde",
- "serde_json",
+ "serde_yaml",
  "tabled",
  "tonic",
  "tonic-prost",
--- a/common/go/operator/server.go
+++ b/common/go/operator/server.go
@@ -11,6 +11,17 @@ import (
 	"github.com/yanet-platform/yanet2/common/go/xgrpc"
 )
 
+// Largest gRPC request an operator accepts.
+//
+// A bulk configuration push carries a whole table in one request, so the
+// four mebibytes gRPC defaults to would bound a legitimate update; the
+// gateway standing in front of the operators allows the same ceiling. Only
+// the request direction is capped: a reply repeats a table back with
+// everything the operator stamps on each entry, so it outgrows the request
+// that produced it and keeps the far larger ceiling gRPC gives it. A reply
+// travelling through the gateway is still bounded by the gateway.
+const maxRequestSize = 1024 * 1024 * 256
+
 type grpcServerOptions struct {
 	Log *zap.Logger
 }
@@ -49,7 +60,7 @@ func NewGRPCServer(
 		o(opts)
 	}
 
-	server := grpc.NewServer()
+	server := grpc.NewServer(grpc.MaxRecvMsgSize(maxRequestSize))
 	serviceNames := make([]string, len(services))
 	for idx, register := range services {
 		serviceNames[idx] = register(server)
--- a/common/go/operator/server_test.go
+++ b/common/go/operator/server_test.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"context"
 	"net"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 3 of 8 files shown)


egnees:
@codex review

egnees:
@codex review

egnees:
@codex review

egnees:
@codex review

egnees:
Addressed in `32140cfb`. The ceilings are no longer symmetric: the operator caps only the request direction (`MaxRecvMsgSize`), so a reply keeps the far larger ceiling grpc-go gives it by default instead of the 256 MiB I had pinned it to, and the client's decoding ceiling is now four times its encoding ceiling. A table accepted at the request ceiling therefore stays listable.

One limit is deliberately left alone: the gateway fronting the operators caps both directions at 256 MiB for every module, so the proxied path is still bounded there. Moving that is a repo-wide policy change rather than something this issue should decide, and paginating the listing is a proto and service change well outside `operators/route/cli/neighbour`.

@codex review

egnees:
@codex review

PR #2585, opened 2026-09-10
feat(route,cli)!: apply route batches from a file and show the matched lookup prefix

### LINKED ISSUE #2361: feat(operators): route cli batch insert, no-flush and lookup prefix
## Motivation

`yanet-cli operator route insert` handles one route per process and always sets `do_flush = true`, so seeding N static routes from a script costs N FIB rebuilds and N pushes to every gateway; the proto's `do_flush` cannot be turned off from the CLI. `lookup` answers with the routes but drops the matched prefix the server returns, so the operator cannot tell which entry won.

## Outcome

`insert`/`remove` accept `--no-flush` and a file of routes (one flush at the end); `lookup` prints the matched prefix in both renderings.

## Evidence

- `do_flush` hard-coded — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/cli/route/src/main.rs#L329, https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/cli/route/src/main.rs#L367; proto field — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/operatorpb/v1/route.proto#L77-L79.
- `LookupRouteResponse.prefix` discarded — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/cli/route/src/main.rs#L305-L317; proto — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/route/operatorpb/v1/route.proto#L61-L63.

## Scope

`operators/route/cli/route`.

## Acceptance

- [ ] `insert --file routes.yaml` (schema documented in `--help`) inserts every route and flushes once; `--no-flush` on single inserts defers the flush to an explicit `flush`.
- [ ] `lookup` shows the matched prefix in the table and in JSON.
- [ ] Tests cover file parsing and the `do_flush` flag.

egnees:
- Adds a `batch` mode: `batch insert <PATH>` and `batch remove <PATH>` read a YAML document of routes and apply the whole set in one run; the schema is printed under the long help of both.
- A batch asks for exactly one flush, riding the last route, so seeding N routes no longer costs N FIB rebuilds and N pushes to every gateway.
- The path is a positional rather than the `--file` flag the issue sketches, because the CLI contract in `AGENTS.md` reserves flags for everything except the file a command consumes, and `insert`/`remove` already spend their positional on the prefix.
- `--no-flush` skips the flush request on both the single-route and the batch form. It does not stage the change: the operator's reconcile loop snapshots the RIB every `DefaultReconcileInterval` (30s) regardless, so this suppresses the immediate rebuild rather than holding the routes back until an explicit `flush`, and the help and success lines say only that.
- The batch document is refused before the first request when it names no routes, leaves an entry without a nexthop, or carries a key outside the schema.
- A run that fails part-way puts the position in the error message, not only the hint, so `--format json` carries it too; the hint says how much of the batch is already in the RIB.
- Adds `Error::with_context` to `ync`, which prefixes an error message while keeping kind, service and raw status intact — the hint is invisible to the serializing backend, so per-item progress belongs in the message.
- `lookup` reports the prefix the server matched: above the table in human output, and as a top-level field in JSON, which now carries the whole response rather than only its route list.
- Drops the merged JSON-shape pin on the route message, per the CLI test policy that retires those when their crate is next changed.
- Closes #2361.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2990,7 +2990,7 @@ dependencies = [
  "netip",
  "prost",
  "serde",
- "serde_json",
+ "serde_yaml",
  "tabled",
  "tonic",
  "tonic-prost",
--- a/cli/core/src/errors.rs
+++ b/cli/core/src/errors.rs
@@ -240,6 +240,19 @@ impl Error {
         self
     }
 
+    /// Prefix the failure text with context the caller has and the server
+    /// does not, such as which of a sequence of calls failed.
+    ///
+    /// The context reaches the machine-readable rendering, which drops
+    /// anything meant for a reader alone. The classification and the
+    /// server's own wording are untouched, so the verbose diagnostic still
+    /// shows exactly what came back.
+    pub fn with_context(mut self, context: impl Display) -> Self {
+        self.message = format!("{context}: {}", self.message);
+
+        self
+    }
+
     /// The category this error was mapped to.
     pub fn kind(&self) -> ErrorKind {
         self.kind
@@ -529,6 +542,17 @@ mod test {
 
         assert_eq!("received fatal alert: CertificateRequired", err.message);
         assert_eq!(Some("transport error".to_owned()), err.raw_message);
+    }
+
+    #[test]
+    fn test_with_context_prefixes_the_message_and_keeps_the_classification() {
+        let status = Status::unavailable("backend down");
+        let err =
+            Error::from_status(status, "insert", "grpc://[::1]:8080", "test.Service").with_context("route 2 of 3");
+
+        assert_eq!("route 2 of 3: backend down", err.message());
+        assert_eq!(ErrorKind::Unavailable, err.kind());
+        assert_eq!(Some("backend down".to_owned()), err.raw_message);
     }
 
     #[test]
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -180,3 +180,10 @@ release.
 (`target`, `mode`, `counter`) and its `devices` as named objects. The
 previous flat rule form is refused, and `--format json` prints the wire
 message.
+
+## Route lookup prints the matched prefix in JSON
+
+`yanet-cli-operator-route lookup --format json` prints an object with a
+`prefix` field and a `routes` array, where it printed the route array
+alone. A script reading the old form reaches its routes through `.routes`
+now.
--- a/operators/route/cli/route/Cargo.toml
+++ b/operators/route/cli/route/Cargo.toml
@@ -17,13 +17,13 @@ clap_complete = { version = "4.5", features = ["unstable-dynamic"] }
 prost = "0.14"
 netip = "0.3"
 colored = "3"
-serde = "1"
+serde = { version = "1", features = ["derive"] }
 tabled = { version = "0.21", default-features = false, features = ["ansi", "derive"] }
 tonic = { version = "0.14", features = ["gzip"] }
 tonic-prost = "0.14"
 
 [dev-dependencies]
-serde_json = "1"
+serde_yaml = "0.9"
 
 [build-dependencies]
 ync-build = { path = "../../../../cli/ync-build", version = "0.1", package = "yanet-cli-build" }
--- a/operators/route/cli/route/src/main.rs
+++ b/operators/route/cli/route/src/main.rs
@@ -5,24 +5,32 @@
 //! has propagated) and drives the operator-owned RIB.
 
 use core::{
+    error::Error as StdError,
     fmt::{self, Display, Formatter},
     net::IpAddr,
+    slice,
 };
-use std::collections::HashMap;
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
 
 use clap::{CommandFactory, Parser};
 use clap_complete::engine::{ArgValueCandidates, CompletionCandidate};
 use colored::Colorize;
-use commonpb::{pb::IpPrefix, serde_with};
+use commonpb::{
+    pb::{IpAddress, IpPrefix},
+    serde_with,
+};
 use netip::{Contiguous, IpNetwork};
 use tabled::Tabled;
 use tonic::codec::CompressionEncoding;
 use ync::{
     GlobalArgs,
-    client::{LayeredChannel, Service},
+    client::{self, LayeredChannel, Service},
     completion, display,
     errors::Error,
-    output,
+    output, yaml,
 };
 
 use crate::operatorpb::{
@@ -64,9 +72,11 @@ pub enum ModeCmd {
     /// Perform RIB route lookup.
     Lookup(RouteLookupCmd),
     /// Insert a unicast static route.
-    Insert(RouteInsertCmd),
+    Insert(RouteEditCmd),
     /// Remove a unicast static route.
-    Remove(RouteRemoveCmd),
+    Remove(RouteEditCmd),
+    /// Apply a whole file of routes in one run.
+    Batch(RouteBatchCmd),
     /// Flush RIB to FIB for a configuration.
     Flush(RouteFlushCmd),
 }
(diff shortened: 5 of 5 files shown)


egnees:
@codex review

egnees:
All three findings were valid and are addressed in `55a94ac4`.

**Keep deferred edits out of periodic snapshots (P1).** Confirmed: `Reconciler.Run` calls `Snapshot()` then `Apply()` unconditionally each pass, `RouteSource.Snapshot` always reports `ok=true` off the live RIB, and `DefaultReconcileInterval` is 30s, so `do_flush=false` suppresses the immediate wake and nothing more. Staging the mutation would be a server-side change outside this issue's scope, so the flag now claims only what it does: `--no-flush` is documented as "Skip the flush, leaving the change for the next reconcile pass or an explicit flush", the success line ends `; flush skipped` rather than "not flushed", and the partial-batch hint says routes are "already in the RIB" instead of claiming they are unpublished.

**Make batch files positional (P1).** Right — no yanet CLI passes a file through a flag, and `insert`/`remove` had already spent their positional on the prefix, so there was no room for a second one. The batch form is now its own mode with the path as its positional: `batch insert <PATH> -n <name>` and `batch remove <PATH> -n <name>`. `insert`/`remove` go back to a required prefix and `--via`, and only gain `--no-flush`. This departs from the `insert --file routes.yaml` spelling in the issue's acceptance line; the issue's Outcome section does not fix the spelling and the CLI contract does, so the contract won.

**Include batch failure context in JSON errors (P2).** The position now goes into the message, which every backend reports: `Error::with_context` (new, in `ync`) prefixes it while leaving kind, service and the raw status untouched, so `--format json` yields `"message": "route 2 of 3: nexthop is not reachable"`. The hint keeps the human-only consequence ("routes 1 to 1 are already in the RIB", or "no route of the batch reached the RIB" when the first request fails), appended above whatever hint the status mapping produced. Three unit tests pin all three shapes.

@codex review

egnees:
All four addressed in `40c37430`.

**Keep deferred edits out of periodic snapshots (P1).** The behaviour cannot change from the CLI — staging a RIB mutation is a server-side capability the operator does not have, and adding it is outside this issue's scope (`operators/route/cli/route`). What was still wrong is that the wording left room to read "an explicit flush" as *required*. It no longer does: `--no-flush` is documented as "Skip the immediate flush; the operator still publishes the change on its next reconcile pass", the success line ends `; no immediate flush`, and the helper that picks the flushing request records that skipping it only drops the immediate rebuild and that a run is never atomic either way. If actual staging is wanted, that is a separate issue against the operator, not this CLI.

**Distinguish removed routes in partial-batch hints (P2).** Correct — "already in the RIB" was false for `batch remove`. The hint is now verb-neutral: "routes 1 to N were applied", true whether the batch inserted or removed them, so no verb parameter is needed to make it accurate.

**Avoid asserting the first failed RPC made no mutation (P2).** Correct. The hint no longer claims anything about the refused request: at position 1 it reads "route 1 is unconfirmed and routes 2 to 3 were never sent", and further in "routes 1 to N were applied, route N+1 is unconfirmed, and the rest were never sent". The helper's own comment states why a refused request is reported as unconfirmed rather than as undone.

**Remove the stale atomic-publication claim (P2).** Deleted. The test name already says what it pins, so per the repo's test conventions it now carries no doc comment at all.

@codex review

egnees:
Rebased onto current `main` (it had moved through #2573 and #2587) and addressed four of five in `052f8138`.

**Correct the documented path guarantee (P1).** Right — the alias carried a promise only one of its two producers keeps. The alias now says nothing about paths, and the guarantee sits on `load_batch`, which is what actually prepends it.

**Rewrite the error documentation without internal identifiers (P1).** Reworded to the caller-visible contract only: context the caller has and the server does not, why prefixing rather than hinting is what survives into the machine-readable rendering, and that the classification and the server's own wording are left alone. No field, parameter or backend names remain.

**Remove comments that only paraphrase helper code (P1).** Dropped on all six you named, plus the batch-file command struct for the same reason. What is left carries a reason that is not in the code: why the file schema rejects unknown keys, why the file is checked before the first request, why the flush rides the last route, why a refused request is called unconfirmed, and why insert and remove share one argument set.

**Omit the unsent-route warning when the last route fails (P2).** Fixed — a failure on the final entry now reads "routes 1 to 2 were applied and route 3 is unconfirmed", with no unsent clause. A test covers that case alongside the first-route and middle-route ones.

**Keep deferred edits out of periodic snapshots (P1) — not fixable here, and nothing now promises otherwise.** The finding is about the operator, not this CLI: the reconcile loop republishes the live RIB on its own timer, so no value of `do_flush` can hold an edit back. This PR touches `operators/route/cli/route` and `cli/core`; adding a staging capability means changing `RouteService` and the reconcile source, which is a different change against the operator and deserves its own issue. What the CLI can control is the claim it makes, and it no longer makes the wrong one: `--no-flush` reads "Skip the immediate flush; the operator still publishes the change on its next reconcile pass", the success line ends `; no immediate flush`, and the flush-selection helper records that a run is never atomic either way. The issue's own motivation — N routes costing N FIB rebuilds and N pushes to every gateway — is fully served by collapsing a batch to one requested rebuild.

While rebasing I also moved the matched-prefix line onto `display::KeyValue` from #2587, which is the shared renderer for a key/value block.

@codex review

egnees:
Both new findings fixed in `e00e124b`, along with a round of my own deep review.

**Parse batch files before connecting (P1).** Correct, and the consequence was real: with a mistyped path and a dead endpoint the command reported the connection, not the file. The batch file is now read in `run` before `RouteService::new`, with the endpoint label from `client::resolve_label`, exactly as `modules/forward/cli` does for its update file. Verified against a dead endpoint: a missing file and an empty document each exit 1 naming the file, while a connection failure with a good file still exits 4 with its own hint.

**Keep deferred edits out of periodic snapshots (P1) — still not fixable in this CLI.** No value of `do_flush` can hold an edit back, because the operator's reconcile loop republishes the live RIB on its own timer. Changing that means teaching `RouteService` and the reconcile source to stage a pending RIB, which is a change to the operator, not to `operators/route/cli/route`. The CLI's own claims are accurate: `--no-flush` reads "Skip the immediate flush; the operator still publishes the change on its next reconcile pass", the success line ends `; no immediate flush`, and the flush-selection helper records that a run is never atomic either way. The issue's motivation — N routes costing N FIB rebuilds and N pushes to every gateway — is met by collapsing a batch to one requested rebuild. I am leaving this as a documented limitation rather than silently widening the PR into the operator; if real staging is wanted it deserves its own issue.

My own review pass on top of that found five more, all fixed here:

- `slice::from_ref` was imported from `std` where `core` provides it, against the deny-level convention.
- A one-route batch reported "Inserted 1 routes"; the count is now singular when it should be.
- Nothing pinned the path guarantee documented on the loader, so a test now writes a document the parser accepts and the checks reject, and asserts the path leads the message.
- Two test comments and the loader's error alias named fields or used an I/O word for what is mostly a content failure; reworded.
- `run`'s comment said the command returns success when "the RPC" succeeded, which stopped being one RPC.

@codex review

egnees:
Rebased onto `main` (through #2588–#2593). The conflict was the import block; `serialize_route_source` now sits on the shared `commonpb::serde_with` helper from #2590, and the matched-prefix line already used `display::KeyValue`. Batch loading stays on `yaml::load` rather than `yaml::load_document`: the batch document is not a wire message, so it follows the same loader as the acl, balancer2, route and unrdup files.

@codex review

egnees:
@codex review

egnees:
The two P1 threads GitHub still anchors inline are from earlier commits; both are addressed on `ed1ddb73`.

**Reject repeated BIRD prefixes** — the batch check now refuses a document that names one prefix twice with `source: bird`, and one that gives a bird route several nexthops, which the server refuses mid-batch anyway. Prefixes are compared after host bits are masked off, so two spellings of one prefix still collide.

**Keep deferred edits out of periodic snapshots** — no explicit flush is promised any more. `--no-flush` reads "the operator still publishes the change on its next reconcile pass", the success line says only "no immediate flush", and a batch is documented as never atomic, since the reconciler snapshots the RIB on its own schedule either way.

The re-review on `ed1ddb739b` found no issues and CI is green.

PR #2588, opened 2026-09-10
refactor(cli): render show commands as a key/value block

3Hren:
- decap, dscp, nat64, pdump, blackhole, trafgen, fwstate and the logging show print their object through display::KeyValue instead of a ptree tree, a two-column table or bare println lines.
- decap, dscp, nat64 and pdump drop their ptree dependency, fwstate drops tabled.
- The JSON payloads are unchanged.


3Hren:
@codex review

PR #2589, opened 2026-09-10
build(cli): share one proto build script helper across the cli crates

3Hren:
- Adds the yanet-cli-build crate (ync-build), a client-only tonic build with the shared packages mapped to their crates, the blanket Serialize derive, a per-crate hook for further attributes and one rerun-if-changed per proto.
- Every module, device, object and operator CLI build script uses it instead of repeating the tonic configuration, and drops the serde import inside its proto module because the derive is now qualified.
- Generated code is unchanged apart from the qualified derive path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2643,8 +2643,8 @@ dependencies = [
  "tabled",
  "tonic",
  "tonic-prost",
- "tonic-prost-build",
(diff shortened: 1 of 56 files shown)


3Hren:
@codex review

PR #2590, opened 2026-09-10
refactor(commonpb): share the timestamp, duration and enum name serializers

3Hren:
- Adds commonpb::serde_with with the timestamp and duration serializers readinesspb and ynpb each carried, the short and lowercase enum name helpers behind serialize_state, serialize_backend_kind, the route source and neighbour state names, and declared_name/from_declared_name for an enum field written by its proto name.
- forward reuses filterpb::null_as_default instead of its own copy.
- readinesspb's rerun-if-changed path now points at the proto it names.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3110,6 +3110,7 @@ version = "0.1.0"
 dependencies = [
  "netip",
  "prost",
+ "prost-types",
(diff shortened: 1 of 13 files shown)


3Hren:
@codex review

PR #2592, opened 2026-09-10
fix(cli): keep the place a rejected module config field came from

### LINKED ISSUE #2369: fix(cli): rules files accepted by the operators are rejected by the module clis
## Motivation

The forward operator loads a rules file with omitted keys (`counter`, `devices`, `srcs`, `dsts`) — its Go loader tolerates them. The same file fed to `yanet-cli forward update --rules` is rejected, because the CLI's schema has no defaults: every list must be spelled out as `[]`. Mirror's schema has the same requirement; acl's does not (its `Rule` gets `serde(default)` in the build script). A file that works in production under the operator fails in the operator's own maintenance CLI.

## Evidence

- Forward CLI schema without defaults — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/forward/cli/src/main.rs#L183-L192, documented at https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/forward/cli/src/main.rs#L229-L232; operator loader — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/operators/forward/internal/operator/rules.go#L44-L56.
- Mirror schema — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/mirror/cli/src/main.rs#L138-L149.
- acl as the counter-example — https://github.com/yanet-platform/yanet2/blob/620c131607d1feaf198e6135f8ca7845b8f4ca59/modules/acl/cli/build.rs#L43.

## Scope

YAML schemas of `modules/forward/cli` and `modules/mirror/cli`.

## Acceptance

- [ ] A rules file the forward operator accepts is accepted by `yanet-cli forward update`.
- [ ] Mirror rules may omit empty lists.
- [ ] Tests load a file with omitted keys in both crates.

egnees:
While this branch was in review, #2591 landed the mirror migration it opened with, so what is left is the loader itself and the last acceptance test #2369 asks for.

- `load_document` reads a document that merges nothing straight from the file text, so a rejected field names its path and line again: `mirror0.yaml: rules[1]: unknown field `sourcs4`, expected one of `action`, `devices`, … at line 7 column 5` instead of the bare `unknown field `sourcs4``. Mirror lost that position when it moved off `yaml::load`, and forward never had it.
- A document is the first one in the file, not the last surviving one, and a second document is refused however it is spelled — an explicit `null` or `{}` after the separator was silently dropped, while the operators refuse both, as `common/go/xproto/yaml_test.go` pins. A bare separator closing the stream stays tolerated, a directive ahead of the document included.
- The parser's document count and the text both have to call a tail bare before it is tolerated. The text alone reads only the line breaks Rust knows, so a file broken on a lone carriage return, NEL or U+2028 would have been truncated to its first document in silence; the parser alone cannot tell a bare separator from a spelled-out null.
- Whether a merge key expanded is decided by merging and comparing rather than by a predicate over the tree, which had no arm for a tagged value and would have sent an unexpanded merge key to typed deserialization.
- Mirror gains the test #2369 asks for, loading a file whose empty lists are omitted, and forward's equivalent loads a file rather than a string, so both cover the path `update` actually takes.

Closes #2369.

### DIFF
--- a/cli/core/src/yaml.rs
+++ b/cli/core/src/yaml.rs
@@ -28,7 +28,9 @@ pub fn load<T: DeserializeOwned>(path: impl AsRef<Path>) -> Result<T, Error> {
 /// Loads the one document at `path` as `T` the way the operators read a
 /// file: merge keys expand and an empty document is the default value.
 ///
-/// A file holding more than one document is refused.
+/// A file holding more than one document is refused, a bare separator
+/// closing the stream excepted. A rejected field names the place it came
+/// from, unless a merge key rebuilt the document first.
 pub fn load_document<T: DeserializeOwned + Default>(path: impl AsRef<Path>) -> Result<T, Error> {
     let path = path.as_ref();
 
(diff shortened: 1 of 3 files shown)


egnees:
@codex review

The mirror wire-form migration this PR opened with landed separately in #2591 while it was in review, so the branch was retargeted onto it: what is left is the loader hardening plus the acceptance test #2369 still wanted. Notes on the four findings from the previous round:

- *Bare separators vs explicit nulls* — fixed. A document is now the first one in the file rather than the last surviving one, and the tail is judged from the text, so `name: c0` followed by `---` then `null`, `{}` or a tagged null is refused as a second document, while a bare separator closing the stream stays tolerated. Tests cover the cases `common/go/xproto/yaml_test.go` pins.
- *Quoted `"<<"`* — the merge-key predicate is exactly as accurate as `apply_merge` itself, which has always expanded a quoted `"<<"` because `serde_yaml::Value` drops the tag. The predicate changes nothing here; making the two readers agree on this needs a style-preserving parser, not a different predicate.
- *Timestamp normalization* — real, but the divergence is the operator normalizing `2024-01-01` to `2024-01-01T00:00:00Z` while the CLI keeps what the file says. Mimicking that in the CLI would corrupt string fields, so instead the loader now only claims to tolerate what the operators tolerate, which is the direction the issue is about.
- *Round-trip promise* — the unqualified sentence was in this branch's own upgrade note, which is gone; the note on main makes no such promise.

egnees:
@codex review

Both findings from the last round are fixed, and a deep review pass found a third that they set up.

- *Directives* — a `%YAML`/`%TAG` line before the first `---` no longer counts as document content, so a directive-prefixed stream with a bare separator tail loads.
- *Merge keys inside tagged values* — the predicate is gone. Whether a merge expanded is now decided by merging a clone and comparing, so `Value::Tagged` and anything else `apply_merge` reaches is covered by construction rather than by an arm someone has to remember to add.
- *New, and the reason both of the above are worth having* — the text scan alone reads only the line breaks `str::lines` knows, while libyaml also breaks on a lone `\r`, NEL, U+2028 and U+2029. A file broken that way parses into several documents but scans as one line, so the tail read as bare and the load returned only the first document, silently. The parser's own document count and the text now both have to call a tail bare before it is tolerated: the parser catches the line breaks the scan misses, the scan tells a bare separator from a spelled-out null. Regression tests cover all four break characters plus CRLF.

On the leading-separator case: `---\n---\nname: c0\n` is now refused, which matches `common/go/xproto/yaml_test.go`'s "document after empty ones".

egnees:
@codex review

Fixed, and as a class rather than an instance. The tail scan no longer borrows the standard library's idea of a line break or of whitespace: it splits on the characters libyaml breaks on (LF, CR, CRLF, NEL, U+2028, U+2029) and treats only space and tab as the blank after a marker, the way YAML does. `name: c0<NEL>---<NEL>null` is refused now, and so are the CR, U+2028 and U+2029 spellings; the same four breaks with a bare separator tail still load.

The parser-count conjunct stays as the backstop, so a future gap in the scan can only cause a loud false refusal, never a silent truncation. Tests cover all four break characters against both a value tail and an explicit-null tail, plus the bare-separator tail and CRLF.

egnees:
@codex review

Fixed. A byte-order mark is stripped from the front of each line before the scan classifies it, so a BOM preamble no longer occupies the first document slot and `﻿\n---\nname: c0\n---\n` loads with its bare tail tolerated. A BOM on a content line was already counted correctly and stays that way; a BOM immediately before a marker is refused by libyaml itself, so it never reaches the scan.

Both cases are tested. This is the third presentation detail the scan had to learn — directives, then the parser's line breaks and blanks, now the BOM — so worth restating why the shape is safe: the scan can only ever cause a loud false refusal, because a tail is tolerated only when the parser also reports every document past the first as null. Silent truncation is not reachable from a gap in it.

GeorgyKirichenko (approved):

PR #2594, opened 2026-09-10
refactor(cli): turn the service wrappers into handler functions

3Hren:
- Every CLI's `<X>Service` becomes a type alias of `ync::client::Service` and its methods free functions taking the connected service, so a crate no longer carries a struct and constructor that only forwarded to ync.
- `run` connects through `Service::connect_for` in place, the two-step connection of fwstate, fwstate-map and the route operator included.
- The code-cli skill skeleton follows the same shape, a crate talking to several services keeps its local struct.
- Drops the route CLI's connection-refused test, which only exercised ync's own error mapping.


3Hren:
@codex review

PR #2595, opened 2026-09-10
refactor(cli): share the address octet and enum name conversions

3Hren:
- `commonpb` exposes `ip_octets` and `ip_from_octets`, the conversions its `IpAddress` already did, so balancer2 drops its private copies and error type and unrdup its `IpAddress` builder.
- Balancer2 keeps one `Proto` enum for the flag and the config file, and names its protocol and scheduler through the generated enums.
- The fwstate-map endpoint formatter unmaps IPv4-mapped addresses through `to_canonical`.


3Hren:
@codex review

PR #2600, opened 2026-09-10
test(cli): drop the tests that only exercised clap

3Hren:
- The `debug_assert` and `try_parse_from` tests of ready, metrics, device-vlan, fwstate-map, route and the route operator go, per the CLI test policy: defaults, ranges, conflicts and flag placement are clap's contract, not the crate's logic.
- Forward's two null-field tests parse a string instead of writing a scratch file.


3Hren:
@codex review

PR #2605, opened 2026-09-11
fix(route): refuse a fib with more than 1024 distinct nexthops

3Hren:
- The FIB update keyed ECMP route lists with a 1024-bit set and panicked inside the gRPC handler on the 1025th distinct hardware nexthop, taking the whole director down under the shared-memory lock.
- Such a FIB is now refused with `InvalidArgument` and the abandoned config is freed, so the previously published FIB stays live and the operator gets a request error instead of a crash loop.
- A functional test against shared memory pins both the refusal and the reclamation (thirty refusals on a 2 MB arena), and a service test pins the error code.


3Hren:
@codex review

PR #2607, opened 2026-09-11
fix(dataplane,vlan,nat64): keep header prepends clear of the packet descriptor

3Hren:
- The packet descriptor shares the mbuf headroom with the frame, yet every header prepend was guarded against the full headroom only, so a lineage encapsulated a few times over wrote its new header across the descriptor's list and mbuf pointers.
- Every prepend in the tree now goes through one helper that refuses once the room above the descriptor is spent, callers drop the packet as before on a refused prepend, and the vlan output stops writing through an unchecked prepend.
- A route-mpls functional test loops a packet through re-encapsulation and pins the refusal at the seventh 32-byte pass, which on unchanged code silently ran nine passes.

### DIFF
--- a/devices/vlan/dataplane/dataplane.c
+++ b/devices/vlan/dataplane/dataplane.c
@@ -116,8 +116,12 @@ vlan_output_handle(
 		}
 
 		// Inject new vlan header
-		// FIXME: check error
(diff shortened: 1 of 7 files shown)


3Hren:
@codex review

GeorgyKirichenko (approved):

PR #2610, opened 2026-09-11
refactor(controlplane,modules,devices): share the shared-memory attach config and helper

3Hren:
- Add an attach config that every module and device config embeds inline, together with one attach helper that maps the shared memory, attaches the agent and releases both.
- Rewrite the module and device constructors and configs onto them, keeping each one's arena size, extra fields and YAML keys.
- Report validation errors of inline config fields at the parent's path.


3Hren:
@codex review

PR #2611, opened 2026-09-11
refactor(ffi,modules,devices): classify a typed free once in the shared-memory wrappers

3Hren:
- Let the ffi module and device wrappers run the typed free of their owner through a callback and classify the outcome, so the refusal contract lives in one place.
- Reduce every module and device binding's free to its own C call, which also drops the device type from a failed device free's message.
- Leave the fwstate map object handle as is, since it wraps a shared object rather than a module or device config.


3Hren:
@codex review

3Hren:
@codex review

PR #2613, opened 2026-09-11
feat(cli): discover services that qualify their family's name

egnees:
- A gateway serves one metrics service per dataplane instance and, since #2562, one for the hardware ports, whose name qualifies the trailing segment the family shares; discovery matched that segment exactly, so the port endpoint was reachable only by its full name and appeared in no suggestion, no shell completion and under no short alias.
- Match a family by the trailing part of a registered service's name rather than by its whole last segment, so a member that scopes itself with a qualifier is discovered, suggested, completed and hinted like every other one.
- Give a qualified member the qualifier as its short alias and let it answer to that word alone, so it never joins the candidates for a word that named another service before it was discovered: a word that is nobody's qualifier is matched over exactly the services that were discovered before, which on the registry as it stands leaves every alias in use resolving as it did and adds only `port`.
- Check that against the services a configured gateway really registers — its own, one per built-in module and one per running operator — rather than against an invented registry, since `controlplane` is already ambiguous there and a fixture without the modules would have certified a guarantee nobody has.
- Move alias derivation, the alias map and alias resolution onto the family, since an alias only means anything within one; resolution and its outcome are no longer part of the library's surface, and the two that remain are methods now, which a CLI outside this repository reaching for them would notice.

Follows up on #2562, where the gap was raised in review.

### DIFF
--- a/cli/core/src/discovery.rs
+++ b/cli/core/src/discovery.rs
@@ -4,9 +4,12 @@
 //! a family of services — readiness, metrics, and so on — is simply the
 //! entries whose name ends in that family's own trailing segment (e.g.
 //! `ReadinessService`, `MetricsService`): the built-in one plus one per
-//! running operator. That list is what a CLI probes when no service is
-//! named, what a short alias resolves against, and what an error hint
-//! suggests.
+//! running operator, plus any that prefix that segment with a qualifier
+//! naming the narrower scope they serve. That list is what a CLI probes
+//! when no service is named, what a short alias resolves against, and what
+//! an error hint suggests. Ending a service name in a family's segment is
+//! therefore an undertaking to answer that family's own request, since
+//! every member is probed with it unasked.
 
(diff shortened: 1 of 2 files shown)


egnees:
@codex review

egnees:
Closing in favour of #2644, which repairs the same gap in one line: a family takes a service whose own name *ends with* the family name, so `PortMetricsService` is discovered, suggested, completed and alias-resolvable without any new machinery.

The alias rework carried here turns out to guard against something largely pre-existing. Module metrics services register as `modules.<name>.controlplane.<name>pb.v1.MetricsService`, and alias resolution matches a substring of the whole name, so on any node with a module configured `controlplane` already selects several metrics services — before this change and before #2562. What is left is a gateway-only deployment, where `port` resolves uniquely. That is not enough to justify moving alias derivation, the alias map and alias resolution onto the family type.

PR #2614, opened 2026-09-12
feat(route,dataplane): add the fib as a shared object

3Hren:
- Introduce the route_fib object type: a forwarding table published on its own, whose lifetime is counted by configuration generations and whose per-nexthop counters live in the object's link counter registry.
- Extract the table core (trees, nexthops, route lists, iterator) into a library shared by the module config and the object, with the shared-memory layout and the FFI surface of the module config unchanged.
- Register the object type in the dataplane load list and export its factory next to the module one.
- Cover the object with a walk, publish, pin, refused-free and arena-baseline test and move the route fuzzer onto the shared core.
- Additive step of the route FIB object design, the module link, handler switch and Go service follow separately.

### DIFF
--- a/dataplane/dataplane.c
+++ b/dataplane/dataplane.c
@@ -595,7 +595,7 @@ dataplane_init(
 		}
 
 		static const char *objects[] = {
-			"fwstate_map_v4", "fwstate_map_v6"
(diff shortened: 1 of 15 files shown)


3Hren:
@codex review

PR #2618, opened 2026-09-12
feat(common): merge two lpm tries into one with per source remaps

GeorgyKirichenko:
- A lockstep walk of both source tries subdivides exactly where either source descends, so every merged leaf lies inside a region where both source lookups are constant; two value lines carry the source values per merged leaf, giving one walk that answers both lookups through the remaps.
- The walk grows host side scratch and cuts the final tables as value lines under the merged trie's context, so they are position independent and may live in shared memory next to it; values are carried opaquely, including the invalid sentinel.
- The owning lpm suite verifies the remap contract exhaustively over a small key space, over every source range endpoint with both neighbours for wide keys, over the empty and identical source edges, and over a starved merge returned to zero, with allocation accounting on teardown.

### DIFF
--- a/common/lpm.h
+++ b/common/lpm.h
@@ -160,6 +160,247 @@ lpm_free(struct lpm *lpm) {
 	memory_context_fini(memory_context);
 }
 
+/*
+ * Remap tables of a merged trie: one entry per merged leaf, holding the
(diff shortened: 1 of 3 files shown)


GeorgyKirichenko:
@codex review

PR #2619, opened 2026-09-13
perf(acl,dataplane)!: hoist per-burst absolutes into the ectx commit handler

### LINKED ISSUE #2579: perf(acl): hoist per-burst preamble into commit handler
## Motivation

`acl_handle_packets` recomputes its per-burst preamble on every burst
([modules/acl/dataplane/dataplane.c:131-L233](https://github.com/yanet-platform/yanet2/blob/a4d9f5de28a5420c4a26bd08957147bef782bd4c/modules/acl/dataplane/dataplane.c#L131-L233)):
nine `counter_get_address` lookups — allow/deny/check_pass/check_miss/create/sync/invalid/non_term/no_match
([L198-L233](https://github.com/yanet-platform/yanet2/blob/a4d9f5de28a5420c4a26bd08957147bef782bd4c/modules/acl/dataplane/dataplane.c#L198-L233)) —
the rules-registry counter handles array
([L195-L196](https://github.com/yanet-platform/yanet2/blob/a4d9f5de28a5420c4a26bd08957147bef782bd4c/modules/acl/dataplane/dataplane.c#L195-L196)),
and the v4/v6 object-link fwtable pointers
([L163-L164](https://github.com/yanet-platform/yanet2/blob/a4d9f5de28a5420c4a26bd08957147bef782bd4c/modules/acl/dataplane/dataplane.c#L163-L164));
per packet, the action resolution additionally walks `ADDR_OF(&acl_config->targets)`
([L532](https://github.com/yanet-platform/yanet2/blob/a4d9f5de28a5420c4a26bd08957147bef782bd4c/modules/acl/dataplane/dataplane.c#L532)).

Measured context (owner's perf analysis of this module): the preamble costs ~50%/pkt on
fragmented vlan-demux fronts — 82.8 vs 55.3 ns/pkt at batch=32. The existing benches track it:

```bash
taskset -c 8 go test -run '^$' -bench 'BenchmarkACLDataset' -benchtime 1s -count 6 ./modules/acl/tests/functional/
```

(and `BenchmarkACLDatasetTopo` for the fragmented case).

## Scope — two layers (constraint learned in #2603's review)

The commit handler contract (#2578, ABI 29) permits only **generation-invariant** writes into the
shared module config — the item is shared across generations and read concurrently by older
generations' workers. Anything **per-worker** must be derived in the per-worker ectx layer instead.

- **Generation-invariant (acl commit handler)**: the `targets` array pointer (absolutizing the
  config's own internal pointer); anything else in `struct acl_module_config` that is worker-free.
- **Per-worker (ectx-layer derivation)**: the 9 counter addresses (module counter storage is
  per-worker), the rules-registry counter handles array (per-worker runtime storages), and the
  v4/v6 object-link fwtable pointers (per-worker object ectx links). Design options: extend the
  per-worker resolve pass (`config_gen_ectx_resolve_counters` walks module ectxs already), or
  introduce a per-worker module prepare hook (runs once per worker per generation, single-writer
  per ectx by the same assigner argument). Decide in this issue before implementing.

## Outcome

The per-burst preamble disappears from `acl_handle_packets`: the hot path loads precomputed
absolute values for everything it currently resolves per burst.

## Constraints

- Depends on #2578 (merged in #2603): commit handlers are void, run once per published
  generation, generation-invariant writes only.
- No changes to other modules.

## Acceptance

- [ ] acl's commit handler absolutizes the generation-invariant values (targets pointer)
- [ ] per-worker values (9 counters, rules handles, fwtables) are derived in the ectx layer and precomputed for the hot path
- [ ] no per-burst `counter_get_address` / fwtable / handles resolution left in `acl_handle_packets`
- [ ] before/after numbers for `BenchmarkACLDataset` and `BenchmarkACLDatasetTopo` recorded in this issue, fragmented front included

Source: decompose of #2548 (owner-dictated step 2: "process acl module"); re-scoped after #2603 review established the generation-invariant vs per-worker split.

GeorgyKirichenko:
- Adds a generic per-ectx commit handler (module_commit_ectx_handler plus a declared buffer size, ABI 30): the control plane allocates a zeroed per-context buffer, and the per-worker absolutization pass fills it once per context per published generation, single-writer, idempotent, under the same release/acquire handoff as every other derived field.
- The acl hot path loads one abs_module_prepared pointer instead of resolving nine counters, the rules-registry handle array and both fwtables every burst; the generation-invariant targets pointer moves to the module commit handler.
- All twelve module constructors zero their descriptors: the loader copies every declared field, so an unset one would export heap garbage.
- Interleaved A/B against a rebuilt baseline: −1.1% dataset and −2.0% (median) topo at batch 32, ~0 at batch 512; the synthetic ruleset's classification body dominates, the win concentrates at the fragmented small-batch points the issue targets.

Closes #2579.


GeorgyKirichenko:
@codex review

PR #2637, opened 2026-09-15
refactor(controlplane): derive grpc service names from descriptors

3Hren:
- Derive static Go gRPC service names from generated descriptors for built-in services, devices, and modules, keeping registration and logging aligned.
- Correct the gateway registry from `Auth` to the registered `AuthService`, fixing authentication service discovery and lookup through the HTTP proxy.
- Extend the gateway backend-kind test to require `AuthService` as built-in and reject the stale `Auth` entry.


3Hren:
@codex review

PR #2639, opened 2026-09-15
refactor(cli): use generated grpc service names

3Hren:
- Replace 32 handwritten Rust gRPC service names with tonic-generated `*_server::SERVICE_NAME`, preserving wire identities and the public balancer export; enable server modules at the three existing generation entry points and update CLI guidance and the starter example.
- Measure the generation toggle alone against `93d82875` on a shared Linux VM with a Xeon E5-2683 v4, CPUs 8–15, eight Cargo jobs, Rust/Cargo 1.96.0 and protoc 3.21.12; use offline builds with warm dependencies and filesystem caches, cleaning only the 37 workspace packages before each rebuild and excluding cleaning time (six check and three release samples per variant).
- Observe workspace-rebuild medians of 8.206 → 9.114 s for `cargo check` (+11.07%) and 87.620 → 87.384 s for release (−0.27%); after adding an ACL RPC, medians are 2.525 → 2.628 s (+4.09%) and 85.566 → 87.341 s (+2.07%), respectively.
- Observe effectively unchanged aggregate size across 32 release executables (−576 bytes); these local measurements do not establish a release-build speedup or a CI timing guarantee.
- Account for existing `commonpb` rebuild invalidation: incorrect `rerun-if-changed` paths trigger reruns and dependent CLI rebuilds in both variants, including without source edits, so the schema-edit samples include broader rebuilds; fixing those paths is outside this PR.


3Hren:
@codex review

PR #2640, opened 2026-09-15
refactor(acl): normalize go initialisms

3Hren:
- Use Go's ACL initialism consistently in binding types, action conversion methods, and tests.
- Update every caller and the affected comments while preserving packet handling and wire formats.


3Hren:
@codex review

PR #2650, opened 2026-09-15
refactor(forward): validate grpc requests in pb messages

3Hren:
-Moves request-only name and action checks into the forward protobuf messages.
-Preserves filter conversion and counter materialization behavior.


3Hren:
@codex review

PR #2654, opened 2026-09-16
feat(pdump): validate requests in protobuf

3Hren:
• Moves pdump request-only checks into CGO-free protobuf Validate methods.
• Keeps filter compilation, default mode handling, and stateful handler work in place.
• Adds strict C-limit parity and sanitizer/fuzz build propagation.


3Hren:
@codex review

PR #2656, opened 2026-09-16
feat(unrdup): validate requests in protobuf

3Hren:
• Moves request-only unrdup checks into protobuf Validate methods.
• Preserves conversion, duplicate, and dataplane-dependent handler behavior.
• Adds boundary, enum, nested-message, and C-limit parity coverage.


3Hren:
@codex review

PR #2657, opened 2026-09-16
feat(l3b): validate requests in protobuf

3Hren:
• Moves l3b request-only names, nested messages, and list-element checks into protobuf Validate methods.
• Preserves backend conversion and state-dependent service errors.
• Adds boundary and nil-element validation coverage.


3Hren:
@codex review

PR #2658, opened 2026-09-16
feat(fwstate-map): validate requests in protobuf

3Hren:
• Moves map name, kind, direction, and cursor request checks into the protobuf package.
• Reuses field-aware map-name validation across fwstate and ACL callers.
• Adds C buffer parity and backward-cursor regression coverage.


3Hren:
@codex review

PR #2663, opened 2026-09-16
refactor(route-mpls): move request validation into protobuf messages

3Hren:
- Moved route-mpls name checks into protobuf request validators.
- Validated MPLS labels through nested next-hop messages before config mutations.
- Kept prefix and address conversion plus update and withdraw behavior in the service.
- Moved request-validation coverage into protobuf package tests.


3Hren:
@codex review

PR #2664, opened 2026-09-16
refactor(acl): validate requests in protobuf

3Hren:
- Moves ACL request-only validation into hand-written protobuf methods with field-aware limits.
- Keeps conversion, state, backend, and metrics-flight logic in handlers.
- Moves request-validation coverage beside the ACL protobuf definitions while retaining behavior tests.


3Hren:
@codex review

PR #2666, opened 2026-09-16
fix(modules): report deleting a referenced config as a failed precondition

3Hren:
- Map a module config delete refused because a chain still references the config to FailedPrecondition instead of Internal in acl, blackhole, decap, dscp, fwstate, nat64, pdump, route, route-mpls and unrdup.
- Keep Internal for every other backend failure.
- Cover the mapping with service tests and one dataplane harness test that deletes a config a device still reaches.


3Hren:
@codex review

PR #2668, opened 2026-09-16
fix(builtin,controlplane): report builtin counter and device delete failures with explicit codes

3Hren:
- Report deleting a predefined topology device as FailedPrecondition instead of InvalidArgument, tagged at the source in the control-plane config.
- Report failed counter reads (by tags, workers, ports) as Internal instead of Unknown.
- The `yanet-cli device delete` JSON error kind for a topology device changes from `invalid_argument` to `rpc`, the exit code stays 1.


3Hren:
@codex review

PR #2669, opened 2026-09-16
fix(gateway,bird-adapter): report registration and import setup failures with explicit codes

3Hren:
- Report a gateway Register that cannot create the backend client as Internal instead of Unknown.
- Report unusable bird-adapter import parameters as InvalidArgument and other setup failures as Internal instead of Unknown, keeping Unavailable for an unreachable route operator.


3Hren:
@codex review

PR #2670, opened 2026-09-16
fix(route-operator): report neighbour table refusals with their own codes

3Hren:
- Report creating an existing neighbour table as AlreadyExists, a missing table as NotFound and removing a built-in table as FailedPrecondition instead of Internal.
- Export the neighbour table refusals as sentinel errors.


3Hren:
@codex review

PR #2671, opened 2026-09-16
fix(route-operator): report route deletes and flushes of an unknown config as not found

3Hren:
- Report DeleteRoute and FlushRoutes for a name that is neither configured nor backed by a RIB as NotFound, as ShowRoutes and LookupRoute already do.
- Keep an empty success for a configured module that has no RIB yet.


3Hren:
@codex review

PR #2673, opened 2026-09-16
fix(route-mpls): reject an update event without a branch

3Hren:
- Reject a route-mpls update event that sets neither update nor withdraw as InvalidArgument instead of silently skipping it.


3Hren:
@codex review

PR #2674, opened 2026-09-16
fix(builtin): reject agent names the shared memory cannot hold

3Hren:
- Reject an ExtendAgent agent name with a NUL byte or longer than the C agent name buffer as InvalidArgument, so a truncated name can no longer grow another agent.


3Hren:
@codex review

PR #2677, opened 2026-09-16
fix(route,route-operator): reject wide mac addresses and contradictory family filters

3Hren:
- Reject a FIB nexthop or neighbour MAC address wider than EUI-48 as InvalidArgument instead of silently dropping its upper bits.
- Reject a FIB or route listing that asks for both IPv4-only and IPv6-only entries instead of returning an empty listing.


3Hren:
@codex review

PR #2678, opened 2026-09-16
test(dscp): pin the marking limits to the dataplane constants

3Hren:
- Name the DSCP flag and mark limits of the request rules and pin them to the dataplane marking constants with a parity test.


3Hren:
@codex review

PR #2679, opened 2026-09-16
fix(pdump,web): reject config updates without an update mask

3Hren:
- Reject a pdump SetConfig request without an update mask as InvalidArgument instead of silently ignoring its config, keeping an empty mask as create-with-defaults.
- Require the update mask in the web pdump API and remove the unused config hook that sent none.


3Hren:
@codex review

PR #2680, opened 2026-09-16
fix(fwstate): return an empty response for a tolerated missing config

3Hren:
- Return an empty ShowConfig response instead of a nil one when the request tolerates a missing config.


3Hren:
@codex review

PR #2681, opened 2026-09-16
fix(unrdup): stop nesting grpc statuses in config conversion errors

3Hren:
- Build unrdup config conversion errors as plain errors and report them once as InvalidArgument, so the message no longer carries a nested rpc error.


3Hren:
@codex review

PR #2682, opened 2026-09-16
refactor(fwstate): keep the merged sync config check out of request validation

3Hren:
- Rename the merged sync config check so the request validation interceptor can never run it on a received message.


3Hren:
@codex review

PR #2685, opened 2026-09-16
fix(common,decap,dscp): name the field of a malformed prefix

3Hren:
- Report a malformed decap or dscp prefix under its own field, such as prefixes6[1], instead of a generic prefixes[1].


3Hren:
@codex review

PR #2687, opened 2026-09-16
fix(modules,devices,builtin): name the missing config in not found errors

3Hren:
- Report a missing module or device config as config "<name>" not found in blackhole, decap, dscp, fwstate, nat64, route-mpls, trafgen and unrdup instead of four different wordings.
- Report a missing pipeline or function by its name instead of a bare not found.


3Hren:
@codex review

PR #2689, opened 2026-09-16
fix(l3b): stop nesting grpc statuses in filter conversion errors

3Hren:
- Report an invalid l3b filter network or range once as InvalidArgument, so the message no longer carries a nested rpc error.


3Hren:
@codex review

PR #2691, opened 2026-09-16
fix(forward,mirror): reject actions the dataplane cannot carry

3Hren:
- Reject a forward or mirror action with an undeclared mode as InvalidArgument instead of silently treating it as none.
- Reject an action target or counter with a NUL byte or longer than its C buffer instead of silently truncating it.


3Hren:
@codex review

Issue #2626, opened 2026-09-14
test(l3b): cover a staged cli-to-gateway packet smoke


moonug:
## Motivation

The l3b candidate contains a Rust CLI, Gateway registration, and in-process packet tests, but those pieces do not prove a staged CLI-to-packet workflow.

The framework CLIBinaryNames list [omits yanet-cli-l3b](https://github.com/yanet-platform/yanet2/blob/5d6a6eccc44c818b2150c42990b9a58788d16611/tests/functional/framework/framework.go#L54-L67), and [guest-local staging copies that list](https://github.com/yanet-platform/yanet2/blob/5d6a6eccc44c818b2150c42990b9a58788d16611/tests/functional/framework/framework.go#L1516-L1520).

The existing [l3b packet harness constructs shared-memory objects directly](https://github.com/yanet-platform/yanet2/blob/5d6a6eccc44c818b2150c42990b9a58788d16611/modules/l3b/tests/functional/l3b_test.go#L49-L70), bypassing the CLI and Gateway.

## Closing outcome

One reproducible functional smoke proves the staged l3b CLI, real Gateway service, configuration activation, and resulting packet behavior together.

## Scope

A minimal existing l3b forwarding scenario in the functional framework. Converter restoration, exhaustive protocol coverage, performance qualification, and deployment are excluded.

## Acceptance

- [ ] The guest-local staged environment contains the executable l3b CLI from the tested build.
- [ ] The smoke creates and inspects a service and installs its module configuration through the CLI and real Gateway, without direct binding calls for those operations.
- [ ] The activated packet path produces the expected encapsulated packet toward the configured real.
- [ ] Missing binaries, failed commands, unavailable service registration, and packet mismatches fail the smoke rather than skip it.
- [ ] The runnable scenario records its tested revision and required environment.

## Coordination

Implementation must be coordinated with @GeorgyKirichenko on #2547 before work starts. This issue records the scope for that discussion; it does not authorize an independent change or defer correctness requirements to optional post-merge work.

Issue #2624, opened 2026-09-14
fix(l3b): reject candidates missing required counters


moonug:
## Motivation

A service candidate must not be accepted when one of its required counters cannot be registered. Otherwise traffic can be processed without the promised accounting.

## Evidence

At the historically inspected #2547 revision pinned below, [link-counter registration is checked, but service and per-real counter registration results are not](https://github.com/yanet-platform/yanet2/blob/5d6a6eccc44c818b2150c42990b9a58788d16611/objects/l3b/api/l3b_virtual_service_object.c#L187-L288).

The dataplane [silently skips COUNTER_INVALID identifiers](https://github.com/yanet-platform/yanet2/blob/5d6a6eccc44c818b2150c42990b9a58788d16611/modules/l3b/dataplane/process.h#L230-L243). These unchecked failure paths are code-proven; allocation failure was not injected during inspection.

## Scope

Required link, service, and per-real counter registration during candidate construction. Counter naming redesign and broader observability work are excluded.

## Acceptance

- [ ] Failure to register any required counter rejects the complete candidate with an error.
- [ ] Failed creation exposes no partial service; failed replacement preserves the prior accepted and active service.
- [ ] Unpublished resources are safely released.
- [ ] Failed replacement preserves the active service's existing counter mappings, session-table ownership and pinned sessions; coordinate shared failure coverage with #2622.
- [ ] Focused failure coverage verifies rejection for required service and per-real counters as well as the link counter.

## Coordination

On 2026-09-14, @moonug confirmed team agreement to start this bounded work on the merged l3b baseline. PR #2547 was merged on 2026-09-14; merge status is not passing-test evidence. This records @moonug's confirmation, not a separate confirmation from @GeorgyKirichenko.

@egness is the planned owner of C3 (required-counter candidate rejection) and C2 (#2622). GitHub assignment remains pending; this issue is currently unassigned. Coordinate shared candidate construction, cleanup and failure-test touches with #2622, and integration with @moonug on #2625 and #2621. No hard dependency or execution order between C2 and C3 is introduced.

Sprint 5 (2026-09-13 through 2026-09-26) commits two people to C2/C3 and C6 then C1, including scoped Echo mapping. Acceptance checks and integration evidence remain required; this coordination confirmation closes no implementation, test, qualification or deployment gate.

moonug:
## Sprint 6 Handoff

[...]

Issue #2622, opened 2026-09-14
fix(l3b): preserve session ownership across failed replacement and retry


moonug:
## Motivation

A failed replacement must leave the active service owning its existing session table. A later successful retry must preserve its pinned sessions.

Historical inspection of the #2547 candidate found ownership moving before installation succeeds; the pinned evidence below is retained, not a fresh verification of the merged baseline.

## Evidence

Code inspection at 5d6a6eccc44c818b2150c42990b9a58788d16611 shows that [CreateVirtualService clears previous.sessionTable during construction](https://github.com/yanet-platform/yanet2/blob/5d6a6eccc44c818b2150c42990b9a58788d16611/objects/l3b/bindings/go/cl3bobject/bindings.go#L251-L289).

The backend subsequently [builds the ring and publishes the replacement, discarding the candidate through Free on either error](https://github.com/yanet-platform/yanet2/blob/5d6a6eccc44c818b2150c42990b9a58788d16611/modules/l3b/controlplane/backend.go#L286-L315). The retained service handle then has no table to offer to a retry.

This failure sequence has not been executed during this inspection.

## Scope

Replacement failure before installation, candidate cleanup, and successful retry on the existing service/session-table boundary. Independent deletion, resizing, and reclamation redesign are excluded.

## Acceptance

- [ ] Failure after candidate construction but before installation leaves the active service, table ownership, and existing sessions unchanged.
- [ ] Cleanup cannot destroy the active table or lose the handle needed for safe later cleanup.
- [ ] A subsequent successful replacement preserves the same table and pinned sessions.
- [ ] A focused failure-then-retry regression exercises the real ownership boundary.

## Coordination

On 2026-09-14, @moonug confirmed team agreement to start this bounded work on the merged l3b baseline. PR #2547 was merged on 2026-09-14; merge status is not passing-test evidence. This records @moonug's confirmation, not a separate confirmation from @GeorgyKirichenko.

@egness is the planned owner of C2 (session ownership across failed replacement and retry) and C3 (#2624). GitHub assignment remains pending; this issue is currently unassigned. Coordinate shared candidate construction, cleanup and failure-test touches with #2624, and integration with @moonug on #2625 and #2621. No hard dependency or execution order between C2 and C3 is introduced.

[...]

PR #2692, opened 2026-09-16
fix(l3b): use a relative pointer for per-real counter ids

### LINKED ISSUE #2676: fix(l3b): prevent dataplane crashes when reading per-real counters
## Motivation

A staged CLI-to-Gateway smoke developed for #2626 reproducibly crashes the dataplane on the first TCP packet after successful L3B service creation and configuration activation.

The expected result is an IP-in-IP packet toward the configured real. Instead, the dataplane receives SIGSEGV and packet capture remains empty. This blocks the packet-behavior acceptance criterion of #2626.

## Evidence

Reproduced three times on revision `b12f76f25806a60bf85266313ebd2449a6d5b2a3` with an uncommitted functional-framework smoke and staging changes. No L3B product code was changed.

Environment: Linux x86_64, QEMU 8.2.2 with KVM, separate control-plane and dataplane processes, and 512 MiB of guest hugepages. SHA256 checks confirmed that guest executables matched the candidate build.

Reproduction scenario:

1. Enable L3B in the functional harness and stage the candidate `yanet-cli-l3b`.
2. Through the CLI and real Gateway, create a fresh service with one real: destination `172.16.0.10`, source network `192.0.2.0/24`, default enabled state and weight 1. Allow TCP source port 80 from `0.0.0.0/0`.
3. Install a module destination filter selecting that service for TCP traffic to `192.168.1.1/32`, then activate an L3B-to-forward pipeline.
4. Send a TCP packet from `10.0.0.1:80` to `192.168.1.1`.
5. Expect IP-in-IP encapsulation from `192.0.2.1` to `172.16.0.10`, preserving the inner packet. Observe the dataplane crash instead.

The unpublished smoke is named `Test_L3B_StagedCLIEncapsulatesTCP`; it is not yet available on main. Service creation, inspection, module installation and pipeline activation all succeeded before packet injection. The existing VM `TestForward` passed on the same candidate.

The cause is confirmed by source and runtime evidence:

- [`real_counter_ids` is a pointer stored inside the shared-memory service structure](https://github.com/yanet-platform/yanet2/blob/b12f76f25806a60bf85266313ebd2449a6d5b2a3/objects/l3b/api/l3b_virtual_service_object.h#L187).
- [The constructor stores the absolute allocation address directly](https://github.com/yanet-platform/yanet2/blob/b12f76f25806a60bf85266313ebd2449a6d5b2a3/objects/l3b/api/l3b_virtual_service_object.c#L269-L287).
- [The dataplane directly indexes that pointer](https://github.com/yanet-platform/yanet2/blob/b12f76f25806a60bf85266313ebd2449a6d5b2a3/modules/l3b/dataplane/process.h#L883-L884). The existing-session path also indexes it at line 828.

In the diagnostic run, fault address `0x7d466e202820` lay within the control-plane mapping `0x7d4668000000–0x7d467a000000`. Symbolization identified `process.h:883`; disassembly confirmed the faulting array read. L3B counted one incoming packet and 84 bytes but did not reach the per-real counter or emit a packet.

## Scope

Cross-process addressability and lifecycle correctness of L3B per-real counter IDs.

This issue blocks #2626. It is distinct from replacement ownership in #2622 and required-counter registration failure handling in #2624. The constructor code overlaps with #2624, so coordinate implementation with @egnees and @GeorgyKirichenko.

Full protocol coverage, performance qualification and deployment are out of scope.

## Acceptance

- [ ] A fresh service configured through the CLI and real Gateway processes the reproduction packet without crashing and emits the expected encapsulated packet.
- [ ] Per-real counters remain readable and correctly updated with separate control-plane and dataplane mappings, including new-session and existing-session paths.
- [ ] Service cleanup remains correct for the affected counter-ID storage.
- [ ] Regression coverage exercises separate process mappings rather than relying solely on the in-process harness.
- [ ] Validation records the exact candidate revision, commands and environment.

moonug:
- Fix a dataplane crash when per-real counter IDs contain addresses valid only in the control-plane shared-memory mapping.
- Store the counter-ID array as a relative pointer and resolve it when processing new or pinned sessions, handling construction failures, and destroying the service.
- Add a regression that maps the same shared-memory segment at two addresses, removes the original mapping, and verifies scheduler-path and pinned-session counter updates through the remaining mapping.
- Extend the existing functional counter test with a repeated flow to cover pinned-session accounting.
- Address the product blocker discovered by #2626; the staged CLI-to-Gateway packet smoke remains a separate follow-up and is not added by this PR.
- Keep replacement ownership (#2622), required-counter registration failure handling (#2624), qualification, and deployment outside this change.

Closes #2676.


moonug:
Removed the L3B test-level UBSan recovery and LSan disable overrides in d7ad514d. The fresh native address/undefined-instrumented target passed four executions with fatal UBSan and leak detection enabled; verbose logs confirmed LSan ran.

The full local build was limited by disk exhaustion, so these results cover the targeted regression only. They do not prove logical cleanup within mmap-backed allocator arenas. Fresh CI will verify the updated candidate.

GeorgyKirichenko (approved):

PR #2701, opened 2026-09-17
feat(cli): show response download progress

3Hren:
- A command waiting on a large response draws a progress line on stderr: a spinner, then a bar of the downloaded bytes once the gRPC message header gives the size.
- The line appears only after a short delay, only for a human on a terminal and not under verbose logging, and it is cleared before the data is printed.
- On a narrow terminal the parts of the line give way so it never wraps.
- Byte counting is a transparent layer of the shared CLI channel, summed over calls running at once, and reports only while a command draws a progress line.
- The first caller, the ipfw-style ACL show, comes in a follow-up PR.


3Hren:
@codex review

